PR #13, opened 2018-05-08
Git plugin readme

MichaelAnckaert:


nntoan:
In the future, I want to list all aliases of git plugin into the Wiki. Because of using git and git aliases is what developers have to type daily. And for me the term of aliases is as same as constants, they SHOULD NOT changed.

PR #17, opened 2018-05-18
Major rearrangement and cleanup

al3xxx:
What was done
- aliases split into categories files with some cleanup
- added selective completion load (into init file and bashrc template)
- functions, history config and shell options are split into categories, some things moved up to oh-my-bash.sh top (non interactive bailout)
- some macos related stuff removed


nntoan:
Closed as it's a part of #39.

PR #21, opened 2018-07-24
Added function to CD up faster using "bu [N]".

tmidi:
@nntoan @MichaelAnckaert 
"bu 3" will take the user 3 folders up => cd ../../../
I believe this will add a fast way for users to change directory up and unlike the method used in aliases, this method is not limited to a specific number of predefined aliases.

### DIFF
--- a/lib/misc.sh
+++ b/lib/misc.sh
@@ -24,3 +24,29 @@ fi
 
 # recognize comments
 shopt -s interactive_comments
+
+# Faster Change Directory up
+function bu () {
+  function bu_usage () {
+     cat <<-EOF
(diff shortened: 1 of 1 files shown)


nntoan:
@tmidi Thanks for your contribution, it looks great! But can you please turn it into a plugin? You can take this as a reference: https://github.com/ohmybash/oh-my-bash/blob/master/plugins/aws/aws.plugin.sh

Cheers

tmidi:
@nntoan thank you for reviewing my change, please see my new PR #22, meanwhile I am closing the

PR #23, opened 2018-07-26
plugins/fasd: Add fasd plugin.

pokowaka:
Fasd offers quick access to files and directories. It
is inspired by tools like autojump, z and v. Fasd keeps track of files
and directories you have accessed, so that you can quickly reference
them in the command line

This plugin sets up fasd to use safe_command_prompt handling
of oh-my-bash to avoid conflicts with fasd's normal bash prompt
handling.

More information about fasd can be found here:
https://github.com/clvv/fasd
.

### DIFF
--- a/oh-my-bash.sh
+++ b/oh-my-bash.sh
@@ -31,6 +31,11 @@ for config_file in $OSH/lib/*.sh; do
   source $config_file
 done
 
+# Load colors first so they can be use in base theme
+source "${OSH}/themes/colours.theme.sh"
+
+# The base theme defines a set of functions that our plugins can use.
+source "${OSH}/themes/base.theme.sh"
 
 is_plugin() {
   local base_dir=$1
@@ -101,9 +106,6 @@ for completion_file in $OSH_CUSTOM/*.sh; do
 done
 unset completion_file
 
-# Load colors first so they can be use in base theme
-source "${OSH}/themes/colours.theme.sh"
-source "${OSH}/themes/base.theme.sh"
 
 # Load the theme
 if [ "$OSH_THEME" = "random" ]; then
--- a/plugins/fasd/fasd.plugin.sh
+++ b/plugins/fasd/fasd.plugin.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env sh
+
+# Initialize fasd, without setting up the prompt hook, which
+# we want to be handled by oh-my-bash.
+
+_fasd_prompt_func() {
+  [ -z "$_FASD_SINK" ] && _FASD_SINK=/dev/null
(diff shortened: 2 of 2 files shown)


oh-my-bash.sh
```diff
@@ -31,6 +31,11 @@ for config_file in $OSH/lib/*.sh; do
   source $config_file
 done
 
+# Load colors first so they can be use in base theme
+source "${OSH}/themes/colours.theme.sh"
```
nntoan: @pokowaka Can you please explain why there is a change inside this bootstrap while you are adding new plugin?

nntoan: @pokowaka Themes should be loaded after plugins, completions, aliases...etc because only theme is something very specified but plugin is for global usage.

Kreyren:
@nntoan Requesting TLDR info on said milestone.

oh-my-bash.sh
Kreyren: @pokowaka any eta?

akinomyoga: Another solution is to move `safe_append_prompt_command` to `lib/` (e.g. `lib/base.sh` or `lib/utils.sh`). I don't think `safe_append_prompt_command` should be considered a part of the utilities for themes.

akinomyoga (commented):
(Something is wrong. I cannot make review comments in this PR although I have started the review. I try to finish this review and start a new review,)

akinomyoga (changes requested):
I don't know the reason, but it seems I cannot make comments to the code in this PR. I make comments here.

https://github.com/ohmybash/oh-my-bash/pull/23/files#diff-8e4df7007d5f147f88ea273d34067414e2e7eda99510e4f6241532afbe40a6f4R13-R14

These two lines can be simply `if type_exists fasd; then`.

https://github.com/ohmybash/oh-my-bash/pull/23/files#diff-8e4df7007d5f147f88ea273d34067414e2e7eda99510e4f6241532afbe40a6f4R9

` sed "s/^[ ]*[0-9]*[ ]*//"` should be  `sed \"s/^[ ]*[0-9]*[ ]*//\"`. The double quotations needs to be escaped by the backslashes.

oh-my-bash.sh
akinomyoga: I have moved the definition of `safe_append_prompt_command ` to `lib/utils.sh`, and then renamed it to `_omb_util_append_prompt_command` in commit a3b8e5f.

akinomyoga:
Because it seems like the original branch has been deleted, I have tailored it locally and merged it at f00c823. Closing.

PR #24, opened 2018-07-26
Allow overriding of prompt addition

pokowaka:
Some environments have stringent security features that do not allow the
execution of functions inside $PROMPT_COMMAND, breaking every individual
theme.

We now allow the overriding of the prompt setter command to workaround
this.

Example usage:

function append_prompt_command_override() {¶
    pre_prompt_functions+=($1)
}

### DIFF
--- a/themes/base.theme.sh
+++ b/themes/base.theme.sh
@@ -499,6 +499,12 @@ function aws_profile {
   fi
 }
 
+
+# Returns true if $1 is a shell function.
+fn_exists() {
+  type $1 | grep -q 'shell function'
+}
+
 function safe_append_prompt_command {
     local prompt_re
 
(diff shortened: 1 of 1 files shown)


DDifDEVzero:
On a fresh install of debian linux with these new lines above in your code i got this error: bash:

"type: append_prompt_command_override: not found."

How can i fix this?

DDifDEVzero:
So, I think I found the error:

The original code is

**# See if we need to use the overriden version
    if [[ $(fn_exists append_prompt_command_override) ]]; then
       append_prompt_command_override $1
       return
    fi**

and I think you have to insert "function" before append_prompt_command_override like the 1st post:

Example usage:

function append_prompt_command_override() {¶
pre_prompt_functions+=($1)
}

and so the code look like this:

**# See if we need to use the overriden version
    if [[ $(fn_exists function append_prompt_command_override) ]]; then
       append_prompt_command_override $1
       return
    fi**

Enjoy!

pokowaka:
I'm traveling with limited access to the internet for at least another week. Would you be able to create a pull request?

Issue #32, opened 2018-08-15
bash: type: append_prompt_command_override: not found.


DDifDEVzero:
Since the new update of the code, there could be a bug when you open your terminal:

bash: type: append_prompt_command_override: not found.

And I think I found the error:

The original code is

**# See if we need to use the overriden version
if [[ $(fn_exists append_prompt_command_override) ]]; then
append_prompt_command_override $1
return
fi**

and I think you have to insert "function" before append_prompt_command_override like the 1st post:

Example usage:

function append_prompt_command_override() {¶
pre_prompt_functions+=($1)
}

and so the code look like this:

**# See if we need to use the overriden version
if [[ $(fn_exists function append_prompt_command_override) ]]; then
append_prompt_command_override $1
return
fi**

Tested on kde neon with success!

inietov:
Hey, I have this problem in my WSL terminal. But I can't find this code to try the fix. Where is the place in wich I need to edit this lines?

DDifDEVzero:
Dear Inietov! As I can see the bug was fixed yesterday in github so please reinstall it  and all works fine now.

PR #33, opened 2018-08-15
function append_prompt_override

DDifDEVzero:
set "function" before append_prompt_override in line 521 so that bash could known the command


nntoan:
Fixes #32

PR #35, opened 2018-08-24
Oh My Bash - Code of Conduct

nntoan:
We're welcoming all reviews.


Kreyren:
What's the usecase for Code Of Conduct here? Afaik CONTRIBTING.md should be sufficient.

Issue #42, opened 2019-01-13
Should use less general envvar/alias/function names when possible to avoid collisions


sljinlet:
I already had an extensive bash setup when I recently found OhMyBash and decided to try it.  I had a number (4-5 so far) of my existing additions and tweaks break because of things that the OMB configuration defines or redefines that clash with my existing stuff.

The way I see it, there are two classes of problems here:

1) Arbitrary names, like all of the settings in the .bashrc file (ex: CASE_SENSITIVE), that could/should be named with a OMB-specific prefix (ex: OMB_CASE_INSENSITIVE) to avoid unnecessary collisions

2) User-facing names, like the "glb" alias in the git plugin.

I see no reason not to do refactor to avoid case #1 as much as possible. I assume that this project is not just targeted and brand new bash users, but also at seasoned veterans like myself.  Why not adopt some sort of "namespace" naming style that avoids collisions that won't affect the end user and will serve to clarify the purpose of the things so named even when they don't collide with something else.

In terms of #2, I think more care should be taken to avoid the unconditional setting of names that should be seen as likely candidates for collision with existing setups.  I mentioned 'gbl' specifically because that one hit me.  I have written my own git script and named it 'gbl'.  It has nothing to do with "git blame".  Even if it did, I'd want to keep its functionality rather than having it overridden by a trivial alias.  In this case, I would suggest either providing a setting variable that allows a prefix to be prepended to all of the defined git aliases (which could default to the empty string), or offering a setting variable that would exclude many if not all of the git aliases from being defined, so that the git plugin would only provide truly novel functionality.  As it is, I disabled the git plugin completely rather than have to think much about this specific problem, as I wasn't aware of anything useful that the OMB git plugin was giving me.

Kreyren:
Referencing https://github.com/ohmybash/oh-my-bash/pull/73 commits welcomed

Kreyren:
`CASE_SENSITIVE` is fixed in https://github.com/ohmybash/oh-my-bash/pull/73/commits/5a5488e7bc7e830805ab7b74339a20ae9c3a531e

Kreyren:
![image](https://user-images.githubusercontent.com/11302521/61761326-b3a03500-adce-11e9-99d6-fc94e1c52022.png)

Please avoid referencing other issues that are not relevant

Kreyren:
Assigned P2 since this is Code Quality which is beeing investigated atm.

PR #73, opened 2019-07-24
Code improvements

Kreyren:
Fixing spellcheck and code quality


Kreyren:
TODO: Error handling should be global

Kreyren:
Merged in https://github.com/RXT067/what-the-bash

custom/aliases/example.aliases.sh
```diff
@@ -1,2 +0,0 @@
-# Add your own custom alias in the custom/aliases directory. Aliases placed
```
nntoan: @Kreyren The custom folder has been put here as an example, reference for people who want to use their custom aliases/completion/plugins/themes.

If you are going to remove it, please also documented regards the "custom" in Wiki.

Thanks.

oh-my-bash.bash
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env bash
```
nntoan: @Kreyren The installation of `oh-my-bash` will create a template automatically and the filename inside it at this moment is `oh-my-bash.sh`, I suggest we create a merge script if someone upgrade from the previous versions their terminal won't mess up by missing sourced files.

custom/aliases/example.aliases.sh
Kreyren: @nntoan Said directory and files are not removed, but renamed on `example.aliases.bash`

oh-my-bash.bash
Kreyren: Sounds like bad practice to rely on script to make shabang. Forcing shabang everywhere should be a better way to handle this assuming situation where said file woudn't be executed correctly because it's missing said shabang. 

> I suggest we create a merge script 

i'm not confident in making this merge -> passing.

Recommend making it a part of the `install.bash`

Kreyren:
Do not merge untill issues are resolved.

themes/rjorgenson/rjorgenson.theme.bash
```diff
@@ -49,8 +50,8 @@ function is_integer() { # helper function for todo-txt-count
 }
 
 todo_txt_count() {
-    if `hash todo.sh 2>&-`; then # is todo.sh installed
-        count=`todo.sh ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }'`
+    if `hash todo.bash 2>&-`; then # is todo.bash installed
```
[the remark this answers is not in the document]
Kreyren: Resolved in https://github.com/ohmybash/oh-my-bash/pull/73/commits/0c36bbfe2725cf4b79353e9e0ad2c7a54590592d

Kreyren:
@nntoan 

> I suggest we create a merge script

I think that these changes doesn't affect the functionality they are just code improvements. -> Merge script shoudn't be required.

Kreyren:
Assigned P1 since this should be investigated with priority to make further merges possible.

oh-my-bash.bash
```diff
@@ -18,78 +18,77 @@ if ! command -v die > /dev/null; then	die()	{
...
+  if is_completion "$OSH_CUSTOM" "$completion"; then
+    fpath=(${OSH_CUSTOM}/completions/${completion} $fpath)
+  elif is_completion "$OSH" "$completion"; then
+    fpath=("${OSH}/completions/${completion}" "$fpath")
   fi
 done
 
 is_alias() {
-  local base_dir=$1
-  local name=$2
```
akinomyoga: The same as the previous comment. Quoting is unneeded.

oh-my-bash.bash
```diff
@@ -142,20 +141,20 @@ source "${OSH}/themes/base.theme.sh"
 
 # Load the theme
 if [ "$OSH_THEME" = "random" ]; then
-  themes=($OSH/themes/*/*theme.sh)
+  themes=("${OSH}/themes/*/*theme.sh")
   N=${#themes[@]}
   ((N=(RANDOM%N)))
-  RANDOM_THEME=${themes[$N]}
+  RANDOM_THEME="${themes[$N]}"
```
akinomyoga, line 147: Quoting is unneeded also for the variable assignments.

oh-my-bash.bash
```diff
@@ -95,9 +95,9 @@ done
 # Figure out the SHORT hostname
 if [[ $(uname) == Darwin* ]]; then
   # macOS's $HOST changes with dhcp, etc. Use ComputerName if possible.
-  SHORT_HOST=$(scutil --get ComputerName 2>/dev/null) || SHORT_HOST=${HOST/.*/}
+  SHORT_HOST=$(scutil --get ComputerName 2>/dev/null) || SHORT_HOST="${HOST/.*/}"
 else
-  SHORT_HOST=${HOST/.*/}
+  SHORT_HOST="${HOST/.*/}"
```
akinomyoga, line 100: These are also unneeded changes. They won't change anything in the behavior.

oh-my-bash.sh
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env bash
```
akinomyoga, line 1: I think this commit should just be dropped. It will wash out the GitHub history of different files and also duplicates `oh-my-bash.bash` to `oh-my-bash.sh`, but need to check later whether this commit actually contains something essential.

completions/brew.completion.bash
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 if which brew >/dev/null 2>&1; then
-  if [ -f `brew --prefix`/etc/bash_completion ]; then
-    . `brew --prefix`/etc/bash_completion
+  if [ -f $(brew --prefix`/etc/bash_completion) ]; then
+    brew --prefix '/etc/bash_completion'
```
akinomyoga, line 4: The meaning is changed here. It should be

```
  if [ -f "$(brew --prefix)/etc/bash_completion" ]; then
    . "$(brew --prefix)/etc/bash_completion"
```

completions/salt.completion.bash
```diff
@@ -112,7 +112,7 @@ _salt(){
         ;;
      -N|--nodegroup)
 	    MASTER_CONFIG='/etc/salt/master'
-        COMPREPLY=($(compgen -W "`awk -F ':'  'BEGIN {print_line = 0};  /^nodegroups/ {print_line = 1;getline } print_line && /^  */ {print $1} /^[^ ]/ {print_line = 0}' <${MASTER_CONFIG}`" -- ${cur}))
+        COMPREPLY="$(compgen -W "`awk -F ':'  'BEGIN {print_line = 0};  /^nodegroups/ {print_line = 1;getline } print_line && /^  */ {print $1} /^[^ ]/ {print_line = 0}' <${MASTER_CONFIG}`" -- ${cur})"
```
akinomyoga: Why is it assigned as a single word? I don't understand this change. I need to later look into the details.

themes/agnoster/agnoster.base.bash
```diff
@@ -220,7 +220,7 @@ prompt_virtualenv() {
 
 # Context: user@hostname (who am I and where am I)
 prompt_context() {
-    local user=`whoami`
+    local user="$(whoami)"
```
akinomyoga, line 223: Quoting unneeded.

themes/rjorgenson/rjorgenson.theme.bash
```diff
@@ -50,8 +50,8 @@ function is_integer() { # helper function for todo-txt-count
 }
 
 todo_txt_count() {
-    if `hash todo.bash 2>&-`; then # is todo.bash installed
-        count=`todo.bash ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }'`
+    if "$(hash todo.bash 2>&-)"; then # is todo.bash installed
+        count="$(todo.bash ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }')"
```
akinomyoga, line 54: Quoting unneeded.

themes/rjorgenson/rjorgenson.theme.bash
```diff
@@ -50,8 +50,8 @@ function is_integer() { # helper function for todo-txt-count
 }
 
 todo_txt_count() {
-    if `hash todo.bash 2>&-`; then # is todo.bash installed
-        count=`todo.bash ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }'`
+    if "$(hash todo.bash 2>&-)"; then # is todo.bash installed
```
akinomyoga, line 53: We may use `type_exists`. See also #239

themes/agnoster/agnoster.base.bash
```diff
@@ -129,7 +129,7 @@ bg_color() {
 # for codes (mycodes) as otherwise it'll clobber the original.
 # this changes from BASH v3 to BASH v4.
 ansi() {
-    local seq
+    local seq|
```
akinomyoga, line 132: This must be a typo. This will make this line and the next line (`declare -a mycodes`) run in subshells, which is equivalent to removing these two lines from the code.

themes/agnoster/agnoster.base.bash
```diff
@@ -292,7 +292,7 @@ __command_rprompt() {
         [ $n -gt 40 ] || break
         times="$times ${tz%%:*}\e[30;1m:\e[0;36;1m"
         times="$times$(TZ=${tz#*:} date +%H:%M)\e[0m"
-        n=$(( $n - 10 ))
+        n="$( $n - 10 )"
```
akinomyoga, line 295: No way. This is an arithmetic expansion but not trying to execute a command with the name $n. Also, quoting is unneeded.

completions/bundler.completion.bash
```diff
@@ -9,8 +9,8 @@
 #   . completion-bundle
 
 __bundle() {
-  local cur=$2
-  local prev=$3
+  local cur="$2"
+  local prev="$3"
```
akinomyoga, line 13: Unneeded.

completions/bundler.completion.bash
```diff
@@ -32,7 +32,7 @@ __bundle() {
 
 __bundle_get_command() {
     local i
-    for ((i=1; i < $COMP_CWORD; ++i)); do
+    for (i=1; i < $COMP_CWORD; ++i); do
```
akinomyoga, line 35: Wrong.

completions/docker-machine.completion.bash
```diff
@@ -231,7 +231,7 @@ _docker_machine() {
     local i
     local command=docker-machine
 
-    for (( i=1; i < ${cword}; ++i)); do
+    for ( i=1; i < ${cword}; ++i); do
```
akinomyoga, line 234: Wrong. It will never work.

akinomyoga (dismissed):
I think this PR should be considered before making large changes to the main script as this includes many useful changes (along with many bugs which should be addressed before merging).

This PR includes the filename changes from `.sh` to `.bash`. Some special care might be paid for the existing users, who might be directly sourcing oh-my-bash files. Anyway, the change of the name `oh-my-bash.sh` -> `oh-my-bash.bash` affects all the users. For example, we may make a symbolic link to `oh-my-bash.bash` from `oh-my-bash.sh`.

akinomyoga:
The changes are overlapped with #270. @Kreyren Can I take over this PR? I am going to rebase the changes on top of the current master while fixing the problems that I made above and excluding the changes that have already been addressed in #270.

akinomyoga:
@Kreyren I'll edit this branch from now.

akinomyoga:
@Kreyren It seems like you are blocking the edits by maintainers so that we can do nothing under this PR unless you respond and update the PR by yourself. Could you enable the edits by checking the checkbox on the bottom of the right column of this page (the following checkbox)?

image

----

The updated patch set is 99344e6...5bd8b39. Here is the summary of rebasing:

* drop 3f80ca6 - oh-my-bash.sh : Replaced $OSTYPE with $(uname)
  - `$OSTYPE` is more reliable
* pick f51e361 - oh-my-bash.sh : Added error handling
  - einfo: renamed to `_omb_log_note`
  - warn: merged with `e_warning` and renamed to `_omb_log_warning`
  - die: renamed to `_omb_log_die` and fixed
* pick f5c0018 - oh-my-bash: Improved sanity for non-interactive shell
* drop 9dee040 - oh-my-bash: Fixed this abomination
  - `true` is not treated as non-zero value in Bash arithmetic evaluations
* drop 016ff57 - oh-my-bash: code improvement
  - already applied
* drop acfdadf - oh-my-bash: removed additional space
  - already applied
* drop 2f6c777 - oh-my-bash: changed extension on .bash
  - will be addressed in #270
* pick 68f95be - oh-my-bash: Double quoting and variable sanitization
  -  partially applied. Most are already applied.
* drop a99d2ef - oh-my-bash: Sanitization for i$, not paying for dev licence
  - ineffective change
* drop 5a5488e - templates/bashrc.osh-template: reaction on #42
  - will be addressed in #180
* drop c324ece - Fixes for bash syntax
  - will be addressed in #270
* drop 161854d - Replace .sh with .bash in all files
  - will be addressed in #270
* drop de4d2f3 - sync
  - already applied (this was originally a merge commit of `master`)
* drop d3748c4 - sync
  - already applied (this seems like also the change from `master`)
* drop 688de57 - Attempts to fix brew.completion.bash
* drop 8ba2e51 - fixing syntax err
  - non-sense. simply wrong. Fixed in a new commit.
* drop 15b9c95 - Attempt to fix salt.completetion
* drop aaa88d2 - Attempt to fix QA for salt.completion
* drop 26cf9fd - Fixing https://app.codacy.com/app/OhMyBash/oh-my-bash/file/37180390922/issues/source?bid=8509045&fileBranchId=13883871#l115
  - non-sense. simply wrong. Fixed in a new commit.
* pick 2345678 - themes/agnoster/agnoster.base.bash : QA fix
* drop ab79502 - themes/agnoster.base : attempt to fix QA
  - non-sense. simply wrong.
* pick 0c36bbf - themes/rjorgenson/rjorgenson.theme.bash : QA fix
  - partially wrong, partially applied.
* drop 7e6e9b6 - QA fix
  - non-sense. simply wrong.
* drop 7970c31 - QA: https://app.codacy.com/app/OhMyBash/oh-my-bash/file/37180391040/issues/source?bid=8509045&fileBranchId=13883871#l234
  - non-sense. simply wrong.

akinomyoga:
@Kreyren It seems like the email address used in these old commits (`w*****x@gmail.com`) seems not bound to your account. Have you changed your email address? You can actually register multiple email addresses to the GitHub account. Or, maybe I can rewrite the email addresses of these commits by `git filter-branch`.

I'm not satisfied with the quality of the shell function `die` (renamed to `_omb_log_die`) in 621c8af...f23984b. First of all, `die` introduced in 4561ee9 clearly has a bug, which makes it difficult to correctly guess what is the original intent. I have fixed it by guess in 4561ee9...f23984b, but the design still seems a bit unnatural.

I actually have doubts on whether we should apply f23984b...58c5c63 although I have picked and edited them in rebasing. Maybe we can show warning messages, but I feel exiting the shell is too much.

oh-my-bash.sh
```diff
@@ -84,7 +84,7 @@ for alias in ${aliases[@]}; do
 done
 
 # Figure out the SHORT hostname
-if [[ "$OSTYPE" = darwin* ]]; then
+if [[ $(uname) == Darwin* ]]; then
```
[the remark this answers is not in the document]
akinomyoga: I reject it.

oh-my-bash.bash
```diff
@@ -18,78 +18,77 @@ if ! command -v die > /dev/null; then	die()	{
...
 
 
 is_plugin() {
-  local base_dir=$1
-  local name=$2
-  test -f $base_dir/plugins/$name/$name.plugin.sh \
-    || test -f $base_dir/plugins/$name/_$name
+  local base_dir="$1"
+  local name="$2"
+	[[ -f "$base_dir/plugins/${name}/{${name}.plugin.sh,_${name}}" ]]
```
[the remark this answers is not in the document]
akinomyoga: I'll address these indentation issues in another PR.

oh-my-bash.sh
akinomyoga: This file turned to be generated by just a mistake in synchronization with the master branch. I dropped the related merging commit.

completions/brew.completion.bash
akinomyoga: I reject this wrong commit.

completions/salt.completion.bash
akinomyoga: Rejected.

themes/rjorgenson/rjorgenson.theme.bash
akinomyoga: Partially applied.

Kreyren:
> @Kreyren It seems like the email address used in these old commits (w*****x@gmail.com) seems not bound to your account. Have you changed your email address? You can actually register multiple email addresses to the GitHub account. Or, maybe I can rewrite the email addresses of these commits by git filter-branch. -- @akinomyoga (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002061061)

Mostly changing emails on random

> The changes are overlapped with #270. @Kreyren Can I take over this PR?  -- @akinomyoga (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1000756839)

Most of those are just quick rewrites of the already present code i doubt that it's worth the effort to take over this MR if you want to do refactoring then consider abstracting to reduce the duplicates and inefficient declarations.

themes/rjorgenson/rjorgenson.theme.bash
akinomyoga: Updated to use `_omb_util_command_exists`.

themes/agnoster/agnoster.base.bash
akinomyoga: Rejected.

themes/agnoster/agnoster.base.bash
akinomyoga: Rejected.

completions/bundler.completion.bash
akinomyoga: Rejected.

completions/docker-machine.completion.bash
akinomyoga: Rejected.

completions/bundler.completion.bash
akinomyoga: I have dropped this and all the related redundant quotations.

Kreyren:
> @Kreyren It seems like you are blocking the edits by maintainers so that we can do nothing under this PR unless you respond and update the PR by yourself. Could you enable the edits by checking the checkbox on the bottom of the right column of this page (the following checkbox)? -- @akinomyoga (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002055209)

I don't have that option there, but when i was submitting the MR it was checked.

akinomyoga:
@Kreyren Thank you for your replies!

> Mostly changing emails on random

OK, I think I will probably do `filter-branch` for your contributions.

> Most of those are just quick rewrites of the already present code i doubt that it's worth the effort to take over this MR if you want to do refactoring then consider abstracting to reduce the duplicates and inefficient declarations.

Hmm, OK! But I have actually completed rebasing it on top of the current master.

> I don't have that option there, but when i was submitting the MR it was checked.

Hmm... I'm not sure what's happening here, but then that's fine! Please forget about it. Sorry for the noise! I'll merge the changes outside this PR.

Kreyren:
> I'm not satisfied with the quality of the shell function die (renamed to _omb_log_die) in 621c8af...f23984b. First of all, die introduced in 4561ee9 clearly has a bug, which makes it difficult to correctly guess what is the original intent. I have fixed it by guess in 4561ee9...f23984b, but the design still seems a bit unnatural. -- @akinomyoga (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002061061)

It's not a bug it's designed that way to standardize the exit codes and provide user-friendly explanation, but now i would write it as:

```shell
# Error handling
## if {einfo,warn,die}() are not set globally (so that end-user can customize the error) then sets script's definitions
command -v einfo >/dev/null || einfo() { printf "INFO: %s\\n" "$1" ;}
command -v ewarn >/dev/null || ewarn() { printf "WARN: %s\\n" "$2" ;}
command -v die >/dev/null || die() { pritf "FATAL: %s\\n" "$2"; exit 1 ;}
```

So that the exit codes can be defined per script as needed or enable users to set their own "enforced" functions.

akinomyoga:
> It's not a bug

I was talking about `exit` in the subshells:

```bash
(printf ... ;  exit "$1") || (printf ,,,; exit $1)
````

These `exit`s will not actually cause the termination of the parent shell but just terminate the subshell. As a result, `die` exits the shell *only when the first argument is `1`*. Is that what you have intended? In the original implementation in  4561ee9,

```bash
die 1 'Message 1' # -> This will terminate shell.
die 2 'Message 2' # -> This will not terminate shell.
```

> but now i would write it as:

OK! Thank you for the example! I'll update it in the suggested manner!

oh-my-bash.bash
akinomyoga: > The installation of `oh-my-bash` will create a template automatically and the filename inside it at this moment is `oh-my-bash.sh`, I suggest we create a merge script if someone upgrade from the previous versions their terminal won't mess up by missing sourced files.

Let's address it in #270. I feel we can just put a symbolic link from `oh-my-bash.sh` to `oh-my-bash.bash`. It is essentially impossible to correctly automatically update the generated file in the general case because users might edit the generated file.

> Sounds like bad practice to rely on script to make shabang. Forcing shabang everywhere should be a better way to handle this assuming situation where said file woudn't be executed correctly because it's missing said shabang.

Maybe I miss the point of the discussion, but I'm not sure whether we should add these shebangs to the script files which are sourced from other script files because they are not supposed to be executed by `exec(2)` syscalls.

Kreyren:
> I was talking about exit in the subshells. The original die in 4561ee9 contains the following structure: -- @akinomyoga (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002079264)

Agree those were supposed to be curly brackets to avoid unwanted escapes e.g.

```sh
{ printf "%s\\n" "whatever" ; exit 1 } || exit 1
```

In case `printff()` is not available to make the code more robust.

akinomyoga:
> Agree those were supposed to be curly brackets to avoid unwanted escapes e.g.
> 
> ```shell
> { printf "%s\\n" "whatever" ; exit 1 } || exit 1
> ```
> 
> In case `printff()` is not available to make the code more robust.

OK! Now everything is now crystal-clear! Thank you for all your help!

----

Now 99344e6...968723c is the latest patch set. I have applied `filter-branch` for your email address, so now these commits will be counted for your contributions. I have updated `_omb_log_die` accordingly. I have updated the check for the non-interactive shells.

Kreyren:
> `_omb_log_die`

I don't like the naming it's designed to be `die` as that is the well known function in other programming languages so it's not supposed to be specific to omb.

Kreyren:
and convention for the other functions is `e-name` meaning `enforced` for phonetic communication which is a common convention for wrappers used in computer science to allow extensibility.

akinomyoga:
> > `_omb_log_die`
> 
> I don't like the naming it's designed to be `die` as that is the well known function in other programming languages so it's not supposed to be specific to omb.

Yeah, thank you for the comment! Actually, the naming convention is one of the topics that I wanted to discuss with other people. The problem is that `oh-my-bash` might be used with other Bash plugins outside `oh-my-bash`. Just as we are thinking that *we want to define die*, there are likely to be authors of other external plugins thinking that *they want to define die*. In addition, the users might also think they want to define the function *die*. The definition and the design of these implementations from different frameworks can be incompatible with each other, so there can arise real problems. Also, simpler names are more likely to conflict with real tools/commands.

In that case (of multiple definitions of `die`, etc.), how do you solve the problem? In well-structured programming languages, we typically use *namespaces*. If the user wants to import the names in the global namespace, something like `alias die=_omb_log_die` (akin to `using omb::log::die;` in C++, or `from omb.log import die` in Python) should be defined at the user side. As you have encouraged us to use more organized programming languages such as Guile and Rust in #268, I think you can understand the situation. If `oh-my-bash` was the authority of the Bash/Shell scripting world and had the power to force all the other shell frameworks to be written on top of the `oh-my-bash` world, we might have argued the proper design of the shell language and utilities, but that's not the case, unfortunately.

My policy is that « *we should not pollute the global interactive namespace with the functions for non-interactive uses, and only the functions that are supposed to be directly called from the command line should be placed in the global namespace* ». This is the reason that the namespace `_omb` is started by the underscore `_` implying it is not in the *normal* namespace. The functions `die`, `ewarn`, and `einfo` are clearly the components for a composite code and will not be called from the command line as a single command. Thus they fall into *the non-global commands* in my classification.

Let me hear what is your policy for avoiding conflicts! Or, do you think we should explicitly prohibit the `oh-my-bash` users from combining it with other external plugins such as `fzf`, `bash-completion`, etc.?

akinomyoga:
> and convention for the other functions is `e-name` meaning `enforced` for phonetic communication which is a common convention for wrappers used in computer science to allow extensibility.

Thanks for the clarification! I once thought of sticking with the `e` prefix along with adding the namespaces, which resulted in something like `_omb_log_e_warning`, etc. I felt it too verbose so just dropped `_e` but actually there are no strong preferences for me in this aspect.

Kreyren:
I am aware of the presented issue thus why the provided definitions are prefixed with `command -v function-name >/dev/null || function-name() { ... }` so it's expected to work with multiple definitions that overwrite the provided on demand.

akinomyoga:
So that it can be replaced by the user-provided version of `_omb_log_die`, etc. if and only if the user intensionally replaces the logging function for OMB. Only in this way, we can ensure that the behavior of the replaced function is compatible with the original version of `_omb_log_die`.

Kreyren:
i guess both options are valid due to how shell script is limited functionally e.g. in rustlang i would just use `use Die::die::die() as omb_log_die` which i see as an optimal solution here, but we can't really do outsourced through libraries on shell efficiently.

akinomyoga:
Thanks for the reply.

> i guess both options are valid due to how shell scripting is functionally limited

Yes, I feel it really depends on the purpose. If we are developing an exclusive framework or a new shell language that can be used to develop a standalone shell application on top of the framework, it is good to define a set of basic utilities that shapes the style and designs of the scripting in the shell language.

However, `oh-my-bash` is not the framework to develop a standalone shell application, but primarily a shell configuration library for interactive sessions, where the main target of the shell namespace is *the interactive commands that users input*. The configuration such as `oh-my-bash` is inevitably just a subcomponent of the ecosystem but cannot be the main part that has control over everything. As I have already written, I am strongly opposed to forcibly defining `die`, etc. (which are not supposed to be used as the interactive commands) in the global namespace by `oh-my-bash`.

Of course, if you are the only user of `oh-my-bash`, you can control everything so can safely define `die`, `ewarn`, `einfo`, etc. Actually, it is completely valid that you have additional *personal* configurations of `alias die=_omb_log_die`, etc. in your local configuration. However, now we would like to discuss the design of the master version of `oh-my-bash` designed for arbitrary users/environments.



> Also as a lisp developer i find the `_omb_log_die` naming conversion super annoying as it's too long.

Being a C++ programmer, I don't feel twelve characters are that long. It might be annoying for those who get used to a language with less flexibility (i.e. with many reserved names in the global namespace), but "*super annoying*" just hears like an exaggeration because recent strict languages tend to have longer *fully-qualified* (i.e. unaliased) names for outputting texts/logs, such as `std::cout` / `System.out.println` / `System.Console.WriteLine` / `console.log`, etc. I rather think it's just a matter of whether you get used to it, or it's an issue that can be compensated by tools such as autocompletion/predictions of the text editors.

I think we might think of renaming them to a bit shorter names `_omb_die`, etc., but I actually don't see any use of the function in the current codebase so reluctant to make it directly under the `_omb` namespace. In particular, `die` would be typically used to terminate the program on an error, but we wouldn't expect the interactive shell to be suddenly terminated by any error. At least, before moving to `_omb_die`, I would like to wait and see how it becomes to be used in the codebase for a wihle.

Kreyren:
I also work with C++ and saying that lisp is less flexible is hilarious.. The point of flexibility in terms of naming identifiers is to make them quick to write while not creating conflicts with naming convention that clutter the code that infringes on readability, maintenance and file size where C/C++ are usually very guilty of doing that..

So i prefer identifiers to be at around 5 characters, but I am not too opposed to use C/C++-like naming here.

example of GNU Guile code for comparison:

```scheme
#!/usr/local/bin/guile \
-l fact -e main -s
!#
;; define is used to define variables and functions
(define (choose n m)
  (/ (fact m) (* (fact (- m n)) (fact n))))

(define (main args)
  (let ((n (string->number (cadr args)))
        (m (string->number (caddr args))))
    (display (choose n m))
    (newline)))
```

```console
$ ./choose 0 4
1
$ ./choose 1 4
4
$ ./choose 2 4
6
$ ./choose 3 4
4
$ ./choose 4 4
1
$ ./choose 50 100
100891344545564193334812497256
```

From https://www.gnu.org/software/guile/manual/html_node/Scripting-Examples.html

That said GNU Guile is extensible and functional language so it can't be compared to C/C++/Rust as they are different usecases e.g. I would use GNU Guile to build C++ if i was tasked to do something with C++, but i wouldn't use C++ to build GNU Guile.

akinomyoga:
> I also work with C++ and saying that lisp is less flexible is hilarious.. 

Yeah, I think it was too hilarious. I haven't intended to say lisp is less flexible compared to C++ as a language, but I was specifically talking about the namespace or the identifiers that users can define in the global scope. For example, if the language or its standard library provides a global name `die`, it is considered a bad practice for a library to overwrite the function or to define another function of the same name in the local scope. It might be possible, but it will confuse the programmers. So `die` would be effectively reserved words, and users shouldn't define the function of the same name. That was what I was thinking in writing the previous reply. 

> The point of flexibility in terms of naming identifiers is to make them quick to write while not creating conflicts with naming convention that clutter the code that infringes on readability,

Are you talking about the identifiers in a local/limited scope? Isn't that the same with most languages that we can define short local identitiers? The example you have provided can be naturally translated to C++/Bash/etc. (though we might discuss the levels of optimizations in different languages). Anyway, I have been discussing *the global names* that are exposed to the users from the libraries (that are designed for the general users).

Kreyren:
> For example, if the language or its standard library provides a global name die, it is considered a bad practice for a library to overwrite the function or to define another function of the same name in the local scope. -- @akinomyoga (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096)

I do agree that:

```shell
die() { printf "FATAL: %s\\n" "$2"; exit 1 ;}
``` 

Is a bad practice thus why i am using:

```shell
command -v die >/dev/null || die() { printf "FATAL: %s\\n" "$2"; exit 1 ;}
```

Meaning:
a) if function `die` is NOT defined then define
b) if function `die` IS defined then use the already used

Which allows the user to set up their own handling of fatal errors in their `~/.bashrc` for oh-my-bash to use e.g.

```shell
# ~/.bashrc
die() { printf "Something terrible happened: %s\\n" "$2"; exit 1 ;}
```

> Are you talking about the identifiers in a local/limited scope? -- @akinomyoga  (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096)

yes `oh-my-bash` should run in it's limited scope such as sub-shell to avoid exports to user shell that the user didn't requested.

>  The example you have provided can be naturally translated to C++/Bash/etc.  -- @akinomyoga  (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096)

yes it would be more appropriate to provide this in a different language to also enable translations as those are really painful to do in shell:

```shell
command -v die >/dev/null || die() {
  case "$LANG" in
    # Czech
    cs-*) printf "FATÁLNÍ" %s\\n" "$2" ;;
    # Default
    *) printf "FATAL: %s\\n" "$2"
  esac
  
  exit 1
}
```

In general shell should just be used as shell manager meaning to manage `PS1`, aliases and what gets sourced for `PATH` defining complicated functions and commands for production is malpractice.

> The example you have provided can be naturally translated to C++/Bash/etc.  -- -- @akinomyoga  (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096)

Also don't think of shell as a C++ it has a very similar syntax, but it's more like very limited functional programming e.g. nixlang so it should be treated as such so defining `very_long_identifier_that_does_things` will eventually clutter the code and make it painful to work with, because you don't have the ability to source libraries in a standardized way for this kind of writting style to make sense.

> Anyway, I have been discussing the global names that are exposed to the users from the libraries (that are designed for the general users).  -- @akinomyoga  (https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096)

To clarify:

```shell
command -v die >/dev/null || die() { ... ;}
```

**Is NOT being exported outside of omb**, but **it's affected by definition outside of omb**.

tools/git-prompt.bash
```diff
@@ -109,10 +110,10 @@ __git_ps1_show_upstream ()
 
 	svn_remote=()
 	# get some config options from git-config
-	local output="$(git config -z --get-regexp '^(svn-remote\..*\.url|bash\.showupstream)$' 2>/dev/null | tr '\0\n' '\n ')"
+	local output="$(git config -z --get-regexp '^(svn-remote\..*\.url|bash\.bashowupstream)$' 2>/dev/null | tr '\0\n' '\n ')"
 	while read -r key value; do
 		case "$key" in
-		bash.showupstream)
+		bash.bashowupstream)
```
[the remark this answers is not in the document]
Kreyren: https://github.com/ohmybash/oh-my-bash/pull/73/commits/384d4964151c6505cf2219baf55295c23bd41481

tools/git-prompt.bash
```diff
@@ -109,10 +110,10 @@ __git_ps1_show_upstream ()
 
 	svn_remote=()
 	# get some config options from git-config
-	local output="$(git config -z --get-regexp '^(svn-remote\..*\.url|bash\.showupstream)$' 2>/dev/null | tr '\0\n' '\n ')"
+	local output="$(git config -z --get-regexp '^(svn-remote\..*\.url|bash\.bashowupstream)$' 2>/dev/null | tr '\0\n' '\n ')"
```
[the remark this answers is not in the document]
Kreyren: https://github.com/ohmybash/oh-my-bash/pull/73/commits/6cab3b826d26fbe3c4cdcc9de713f61b19c777f7

tools/git-prompt.bash
```diff
@@ -477,7 +478,7 @@ __git_ps1 ()
 		fi
 	elif [ "true" = "$inside_worktree" ]; then
 		if [ -n "${GIT_PS1_SHOWDIRTYSTATE-}" ] &&
-		   [ "$(git config --bool bash.showDirtyState)" != "false" ]
+		   [ "$(git config --bool bash.bashowDirtyState)" != "false" ]
```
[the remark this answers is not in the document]
Kreyren: https://github.com/ohmybash/oh-my-bash/commit/6cab3b826d26fbe3c4cdcc9de713f61b19c777f7

tools/git-prompt.bash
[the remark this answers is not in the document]
Kreyren: https://github.com/ohmybash/oh-my-bash/pull/73/commits/879609e82809cc419ad57c97d43f686ed2bf4d72

tools/git-prompt.bash
```diff
@@ -492,7 +493,7 @@ __git_ps1 ()
 		fi
 
 		if [ -n "${GIT_PS1_SHOWUNTRACKEDFILES-}" ] &&
-		   [ "$(git config --bool bash.showUntrackedFiles)" != "false" ] &&
+		   [ "$(git config --bool bash.bashowUntrackedFiles)" != "false" ] &&
```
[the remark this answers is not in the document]
Kreyren: https://github.com/ohmybash/oh-my-bash/pull/73/commits/1b698445bbc3b7a18df0b37d0b399e3feb9bd486

akinomyoga:
> Is a bad practice thus why i am using:
> 
> ```shell
> command -v die >/dev/null || die() { printf "FATAL: %s\\n" "$2"; exit 1 ;}
> ```
> 
> Meaning: a) if function `die` is NOT defined then define b) if function `die` IS defined then use the already used

I don't understand. This doesn't solve anything. It seems like you have not yet understood what is the problem. As I have already written in [the first reply](https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002160222) to your question on `_omb_util_die`, the problem is

> The definition and the design of these implementations from different frameworks can be incompatible with each other, so there can arise real problems.

How do you ensure that the existing `die` has the same assumption as OMB that the first argument is the exit status and the second argument is the message?

akinomyoga:
> > Are you talking about the identifiers in a local/limited scope? -- @akinomyoga  ([#73 (comment)](https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096))
> 
> yes `oh-my-bash` should run in it's limited scope such as sub-shell (subshells bad!) to avoid exports to user shell that the user didn't requested.

But, the file `lib/utils.sh` is sourced in the main shell but not in sub-shells. If we want to define these functions only in the subshell as you say, then we need to write something like `$(source lib/utils.sh; some_info_prompt)`. Is that what you suggest? Maybe that is another possible solution, but that is not the choice of the current codebase, so we need large refactoring of the entire codebase. Also, importantly, I don't think that approach is useful.

akinomyoga:
> > The example you have provided can be naturally translated to C++/Bash/etc.  -- @akinomyoga  ([#73 (comment)](https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096))
> 
> yes it would be more appropriate to provide this in a different language to also enable translations as those are really painful to do in shell:

Hmm, maybe I misunderstood the reason why you have provided the example of `(choose n m)` written in Guile. I just thought that you have provided the example to show the flexibility of identifiers in Guile, so I just pointed that the example you have provided doesn't really show the flexibility of Guile compared to the other language because the example can be just rewritten in other languages. That was what I meant in the above sentence.

akinomyoga:
> > The example you have provided can be naturally translated to C++/Bash/etc.  -- -- @akinomyoga  ([#73 (comment)](https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096))
> 
> Also don't think of shell as a C++ it has a very similar syntax, but it's more like very limited functional programming e.g. nixlang so it should be treated as such so defining `very_long_identifier_that_does_things` will eventually clutter the code and make it painful to work with, because you don't have the ability to source libraries in a standardized way for this kind of writting style to make sense.

Could you describe it in more detail? It appears to be the opposite for me. If there is a standardized way of importing libraries for the language, we expect from the language the standardized namespacing /scoping / module system, so we can define the identifiers in a namespace, etc. with *short identifiers*. However, if there is no standardized way of importing libraries, everything will be placed in the same namespace. In such a case, if every library defines functions of the same name but with different behavior, things will clutter. The only way of solving the situation is to protect ourselves by introducing effective namespaces by ourselves using the namespace prefixes to the identifiers.

Actually, you haven't answered the question in [my first reply to your comment on `_omb_log_die`](https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002160222), which is closely related to this discussion.

> Let me hear what is your policy for avoiding conflicts! 

In case you didn't understand my question correctly, let me explain it again: If `oh-my-bash` would be allowed to define `die`, then the other libraries/frameworks should also be allowed to define `die` because `oh-my-bash` is not the standard of Bash. Also, the user might want to define `die` by themselves. However, the problem is that the design of these `die` can be different. For example, `oh-my-bash` expects `die exit_status 'This is an error message'`, but the other libraries might expect `die 'This is an error message' exit_status' or juts `die This is an error message'. The user might want to define `die` for a completely different purpose.

akinomyoga:
> > Anyway, I have been discussing the global names that are exposed to the users from the libraries (that are designed for the general users).  -- @akinomyoga  ([#73 (comment)](https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002781096))
> 
> To clarify:
> 
> ```shell
> command -v die >/dev/null || die() { ... ;}
> ```
> 
> **Is NOT being exported outside of omb**, but **it's affected by definition outside of omb**.

Yes, that's the problem, or isn't that a problem that OMB is unexpectedly affected by the codes outside of OMB? I mean it is fine if the user *intentionally* changes the behavior of OMB by defining `die`, but the problem arises when the user or other libraries define `die` for their own purpose unrelated to OMB and the design of `die` is different.

It is more explicit, unambiguous, and readable that the user defines `_omb_log_die` (or possibly `_omb_log`) when the user intentionally replaces "*`die` of OMB*". Replacing `die` is less descriptive than `_omb_log_die`.

akinomyoga:
I dropped the commits related to these logging functions and merged the changes of this PR. I'm closing the PR now. We discuss the dropped commits for the logging functions at #290.

Issue #38, opened 2018-11-12
powerline theme runs sudo by default


jsharpe:
On some shared systems running sudo would result in an automatic ban of an account. Running sudo should be an opt in process by default rather than opt out.

Kreyren:
Better way to handle root elevation was submitted in https://github.com/ohmybash/oh-my-bash/pull/75

Kreyren:
Assigned P1 since this is security issue.

akinomyoga:
> Assigned P1 since this is security issue.

This doesn't have any actual security risk though it may cause troubles by false-positive detection of `sudo` abuses.

I've just made the sudo check opt-in 1d1b160. This should be the sufficient and correct fix. @jsharpe Thank you for the suggestion!

I also updated the check to use `cat` instead of relying on non-POSIX `uptime`8c6b66c.

PR #75, opened 2019-07-24
 themes/powerline: Improved sudo handling

Kreyren:
Fixes: https://github.com/ohmybash/oh-my-bash/issues/38

### DIFF
--- a/themes/powerline/powerline.base.sh
+++ b/themes/powerline/powerline.base.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Error handling
-if ! command -v info > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
+if ! command -v einfo > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
 if ! command -v warn > /dev/null; then	warn()	{	printf "WARN: %s\n" "$1"	1>&2	;	} fi
 if ! command -v die > /dev/null; then	die() { printf "FATAL: %s\n" "$2"	1>&2	;	exit "$1"	;	} fi
 
--- a/themes/powerline/powerline.base.sh
+++ b/themes/powerline/powerline.base.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Error handling
-if ! command -v info > /dev/null; then	info()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
+if ! command -v info > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
 if ! command -v warn > /dev/null; then	warn()	{	printf "WARN: %s\n" "$1"	1>&2	;	} fi
 if ! command -v die > /dev/null; then	die() { printf "FATAL: %s\n" "$2"	1>&2	;	exit "$1"	;	} fi
 
@@ -10,11 +10,11 @@ checkroot () {
 	if ! ((EUID)); then
 		return
 	elif [[ -x "$(command -v sudo)" ]] && [ -n "$OSH_CHECKROOT" ]; then
-			info "Failed to aquire root permission, trying reinvoking with 'sudo' prefix"
+			einfo "Failed to aquire root permission, trying reinvoking with 'sudo' prefix"
 			exec sudo "$0" "$@"
 			exit 1
 	elif [[ ! -x "$(command -v sudo)" ]] && [ -n "$OSH_CHECKROOT" ]; then
-		info "Failed to aquire root permission, trying reinvoking as root user."
+		einfo "Failed to aquire root permission, trying reinvoking as root user."
 		exec su -c "$0 $*"
 		exit 1
 	else
--- a/themes/powerline/powerline.base.sh
+++ b/themes/powerline/powerline.base.sh
@@ -1,5 +1,27 @@
 #!/usr/bin/env bash
 
+# Error handling
+if ! command -v info > /dev/null; then	info()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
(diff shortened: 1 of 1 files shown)


Kreyren:
Assigned help-wanted since testing of this functionality is needed.

Kreyren:
@nntoan requesting review

themes/powerline/powerline.base.sh
```diff
@@ -1,5 +1,27 @@
 #!/usr/bin/env bash
 
+# Error handling
+if ! command -v info > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
```
akinomyoga, line 4: typo

```suggestion
if ! command -v einfo > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
```

themes/powerline/powerline.base.sh
```diff
@@ -1,5 +1,27 @@
...
+	elif [[ -x "$(command -v sudo)" ]] && [ -n "$OSH_CHECKROOT" ]; then
+			einfo "Failed to aquire root permission, trying reinvoking with 'sudo' prefix"
+			exec sudo "$0" "$@"
+			exit 1
+	elif [[ ! -x "$(command -v sudo)" ]] && [ -n "$OSH_CHECKROOT" ]; then
+		einfo "Failed to aquire root permission, trying reinvoking as root user."
+		exec su -c "$0 $*"
+		exit 1
+	else
+		die 0 "Unable to elevate root access."
```
akinomyoga, line 21: This function `checkroot` will terminate the interactive session unless it is run with the root privilege.

akinomyoga (commented):
Unfortunately, there are many problems with this PR.

- I believe the original purpose of the code is to check whether `sudo` is activated in the current session. After using `sudo` in a shell session, `sudo` will accept commands from the same shell session without the password for a certain time duration. The original code shows whether that duration of password-less `sudo` is timed out or not. However, this patch breaks the behavior. The behavior of the password-less `sudo` is unrelated to whether the current effective user ID is 0 or not and also the behavior of `su`.
- After this change, the user will be required to input the password before every prompt. In the original code, the option `-n` is specified to `sudo` to check whether `sudo` can be executed without the password.
- After this change, error messages will be output before every prompt as far as the current effective user ID is not 0 (root).
- When the current effective user is 0 (root), nothing is executed by `checkroot` so that the check `grep -q` will always fail.

This PR seems like complete nonsense.

Also, there is another room of improvements in the original code.

- `uptime` is not standardized by POSIX. If the purpose is to check the behavior of `sudo`, the check doesn't necessarily need to be performed with `uptime`.
-

akinomyoga:
I've just made the option opt-in in commit 1d1b160. Let us just close this broken PR.

akinomyoga:
I actually tried to fix the PR, but finally I decided to reject it. I leave the fix that has not been adopted here.

```diff
From 096814794ab6b34bdb70f019cb846f8c89ee1c92 Mon Sep 17 00:00:00 2001
From: Koichi Murase <myoga.murase@gmail.com>
Date: Thu, 30 Dec 2021 21:53:53 +0900
Subject: [PATCH] themes/powerline: Adjust the sudo check for the current
 oh-my-bash

---
 themes/powerline/powerline.base.sh | 39 ++++++++++++++----------------
 1 file changed, 18 insertions(+), 21 deletions(-)

diff --git a/themes/powerline/powerline.base.sh b/themes/powerline/powerline.base.sh
index 3231930..add5b60 100644
--- a/themes/powerline/powerline.base.sh
+++ b/themes/powerline/powerline.base.sh
@@ -1,25 +1,22 @@
 #! bash oh-my-bash.module
 
-# Error handling
-if ! command -v info > /dev/null; then	einfo()	{	printf "INFO: %s\n" "$1"	1>&2	;	} fi
-if ! command -v warn > /dev/null; then	warn()	{	printf "WARN: %s\n" "$1"	1>&2	;	} fi
-if ! command -v die > /dev/null; then	die() { printf "FATAL: %s\n" "$2"	1>&2	;	exit "$1"	;	} fi
-
-# Check if executed as root, if not tries to use sudo
-checkroot () {
-	if ! ((EUID)); then
-		return
-	elif [[ -x "$(command -v sudo)" ]] && [ -n "$OSH_CHECKROOT" ]; then
-			einfo "Failed to aquire root permission, trying reinvoking with 'sudo' prefix"
-			exec sudo "$0" "$@"
-			exit 1
-	elif [[ ! -x "$(command -v sudo)" ]] && [ -n "$OSH_CHECKROOT" ]; then
-		einfo "Failed to aquire root permission, trying reinvoking as root user."
-		exec su -c "$0 $*"
-		exit 1
-	else
-		die 0 "Unable to elevate root access."
-	fi
+# Check if executed as root. If not, try to use sudo.
+function __powerline_sudo {
+  if ((EUID == 0)); then
+    return 0
+  elif [[ $OSH_CHECKROOT ]]; then
+    if _omb_util_binary_exists sudo; then
+      _omb_log_info "Failed to acquire the root permission. Retrying it with 'sudo' prefix" >&2
+      sudo "$0" "$@"
+    else
+      _omb_log_info "Failed to acquire the root permission. Retrying it as the root user." >&2
+      su -c "$0 $*"
+    fi
+    return 1
+  else
+    _omb_log_error "Unable to elevate to the root access." >&2
+    return 0
+  fi
 }
 
 # Define this here so it can be used by all of the Powerline themes
@@ -41,7 +38,7 @@ function __powerline_user_info_prompt {
   local color=${USER_INFO_THEME_PROMPT_COLOR}
 
   if [[ "${THEME_CHECK_SUDO}" = true ]]; then
-    if checkroot uptime 2>&1 | grep -q "load"; then
+    if __powerline_sudo uptime 2>&1 | grep -q "load"; then
       color=${USER_INFO_THEME_PROMPT_COLOR_SUDO}
     fi
   fi
-- 
2.33.1


```

PR #76, opened 2019-07-24
aliases/pckm : Added aliases for package manager handling

Kreyren:
Attempt to standardize aliases to make workflow with package managers more effective.


aliases/pckm.aliases.bash
```diff
@@ -0,0 +1,59 @@
...
+
+if [[ -x "$(command -v zypper)" ]]; then
+	# Zypper = Zen Yast Package Program (ZYPP?)
+	# Yast = Yet Another Silly/Setup Thing/Thing
+	alias lcp="${OSH_ROOT:+sudo }zypper"
+	alias lcpi="${OSH_ROOT:+sudo }zypper install"
+	alias lcps="${OSH_ROOT:+sudo }zypper search"
+	alias lcpsi="${OSH_ROOT:+sudo }zypper source-install"
+	alias lcpr="${OSH_ROOT:+sudo }zypper remove"
+	if [[ $(cat /etc/os-release | grep -m 1 -o 'openSUSE Tumbleweed') == "openSUSE Tumbleweed"  ]]; then
```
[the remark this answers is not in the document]
Kreyren: Is noted

aliases/pckm.aliases.bash
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+# Created by Jacob Hrbek github.com in 2019
+#  ---------------------------------------------------------------------------
+
+# Logic for root usage
+if [ -n "$EUID" ] && [ -x $(command -v sudo) ]; then
```
akinomyoga, line 6: This will always succeed without quoting.

```suggestion
if [ -n "$EUID" ] && [ -x "$(command -v sudo)" ]; then
```

akinomyoga: I'm again looking at this PR. There are additional several problems in this PR currently.

- `[ -n "$EUID" ]` tests if `EUID` is not an empty string, which will usually be always true. Here we should test that `EUID` is not an integer 0.
- Another condition for `[ -z "$OMB_USE_ROOT" ]` tests if `OMB_USE_ROOT` is empty. However, the effect should take effect only when it has the value `false` according to the modifications in README. Actually, the current implementation is counter-intuitive because oh-my-bash would *not* use the root privilege when any value is set to `OMB_USE_ROOT`.
- `$OMB_USE_ROOT` will fail with `set -u`.

aliases/pckm.aliases.bash
```diff
@@ -0,0 +1,62 @@
...
+# # Zypper = Zen Yast Package Program (ZYPP?)
+# if [[ -x "$(command -v zypper)" ]]; then
+#
+# 	# Yast = Yet Another Silly/Setup Thing/Thing
+# 	alias lcp="${OMB_ROOT:+sudo }zypper"
+# 	alias lcpi="${OMB_ROOT:+sudo }zypper install"
+# 	alias lcps="${OMB_ROOT:+sudo }zypper search"
+# 	alias lcpsi="${OMB_ROOT:+sudo }zypper source-install"
+# 	alias lcpr="${OMB_ROOT:+sudo }zypper remove"
+# 	if [[ $(cat /etc/os-release | grep -m 1 -o 'openSUSE Tumbleweed') == "openSUSE Tumbleweed"  ]]; then
```
akinomyoga, line 56: - The option `-m` is not POSIX compatible
- This test is totally verbose and redundant. We can just check the exit status of grep.

```suggestion
# 	if grep -q 'openSUSE Tumbleweed' /etc/os-release; then
```

akinomyoga:
I have pushed a rebased version to `pr76` branch. The patch set is 8797ea3...9299fb6
- I have fixed all the issues I have raised.
- The plugin name `pckm` is too short. I have renamed it to `package-manager`
- I have renamed `OMB_USE_ROOT` to `OMB_USE_SUDO`
- I have added a configuration `OMB_ALIAS_PACKAGE_MANAGER_SUDO` specifically for this plugin.
- I have added descriptions on the aliases in `aliases/README.md`

akinomyoga:
I have applied the change in 130d561, so I close this PR.

PR #77, opened 2019-07-24
General aliases

Kreyren:
Added ix()
Added xcopy alias
Added llm alias to list all loadable modules from current kernel

### DIFF
--- a/aliases/general.aliases.sh
+++ b/aliases/general.aliases.sh
@@ -29,5 +29,7 @@ alias fix_stty='stty sane'                  # fix_stty:     Restore terminal set
 alias cic='set completion-ignore-case On'   # cic:          Make tab-completion case-insensitive
 alias src='source ~/.bashrc'                # src:          Reload .bashrc file
 
-# Used for piping to remote pastebin from cmdline
+# Used for piping to remote pastebin from cmdline to generate a url
 [[ -x "$(command -v curl)" ]] && ix () { curl -n -F 'f:1=<-' http://ix.io ; }
+# Used for piping to clipboard
+[[ -x "$(command -v xclip)" ]] && alias xcopy="xclip -se c"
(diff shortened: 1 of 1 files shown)


Kreyren:
@nntoan Ready for review

aliases/general.aliases.sh
```diff
@@ -28,3 +28,10 @@ alias show_options='shopt'                  # Show_options: display bash options
 alias fix_stty='stty sane'                  # fix_stty:     Restore terminal settings when screwed up
 alias cic='set completion-ignore-case On'   # cic:          Make tab-completion case-insensitive
 alias src='source ~/.bashrc'                # src:          Reload .bashrc file
+
+# Outputs List of Loadable Modules (llm) for current kernel
+alias llm="find /lib/modules/$(uname -r) -type f -name '*.ko*'"
```
nntoan, line 33: @Kreyren I think these aliases probably go to misc.aliases.sh which make more sense than put it under general? What do you think?

Kreyren: I don't remember :thinking: Use your best judgement?

nntoan: @Kreyren Probably go to misc 😄  General should be kept as minimal as possible or something can affect daily tasks and not very specified to any OSes

akinomyoga (changes requested):
@Kreyren Can you move these new aliases to misc as @nntoan has suggested? I think we can readily merge them after that.

akinomyoga:
I have rebased the change on top of the current master and moved the aliases to `misc.aliases.sh` as suggested by nntoan. The current patch set is 8797ea3...0ace85d.

akinomyoga:
I have also applied changes from this PR in commit 209ffda. Let's close the PR now.

PR #84, opened 2019-07-24
README: Sanitization for local repositories

Kreyren:
Currently `README.md` is using external URL for image which is conflicting with local repositories without internet.

This fixes it and makes repository independant on external files.


Kreyren:
Merged in https://github.com/RXT067/what-the-bash

PR #92, opened 2019-08-02
themes/font: Make it possible to disable clock in font theme

mborsz:
Without this, font theme always shows a clock.

After this change, the font theme will show clock by default, but it will be possible to disable it by setting THEME_SHOW_CLOCK=false in .bashrc


Kreyren:
> `THEME_SHOW_CLOCK=true` -> `THEME_SHOW_CLOCK=${THEME_SHOW_CLOCK:-"true"}`

Why not using `THEME_SHOW_CLOCK=false`

Kreyren:
@nntoan please check my review before merging.

akinomyoga:
This is fine. Merging this.

Issue #63, opened 2019-06-22
Alias conflict between bashmarks.sh and directories.sh


Bulliby:
This alias in `oh-my-bash/plugins/bashmarks/bashmarks.plugin.sh`:

`alias d='bm -d'       # Delete a bookmark [bookmark_name]`

Is in conflict with this line of `oh-my-bash/lib/directories.sh `:

`alias d='dirs -v | head -10'`

PR #97, opened 2019-08-15
plugins/bashmarks: Fixed conflict with d alias

Kreyren:
Fixes: https://github.com/ohmybash/oh-my-bash/issues/63

### DIFF
--- a/plugins/bashmarks/bashmarks.plugin.sh
+++ b/plugins/bashmarks/bashmarks.plugin.sh
@@ -216,4 +216,4 @@ fi
 alias s='bm -a'       # Save a bookmark [bookmark_name]
 alias g='bm -g'       # Go to bookmark [bookmark_name]
 alias p='bm -p'       # Print bookmark of a path [path]
-alias d='bm -d'       # Delete a bookmark [bookmark_name]
+alias db='bm -d'      # Delete Bookmark [bookmark_name]


akinomyoga (commented):
I'm not sure if this is the proper fix. If we change the alias, it becomes inconsistent with the upstream https://github.com/huyng/bashmarks. Unless the upstream also changes the name at the same time, this change will confuse users.

By the way, I again found that the copyright notice of the upstream has been rewritten in this script. I think we need to check the copyright of all the plugins and themes that have the corresponding upstream.

akinomyoga:
I also lowered the priority of this PR. Our template loads `bashmarks` by default, so the existing users with the default setup should be familiar with the `bashmarks''s `d` rather than `d` in `lib/directories.sh`. If either one would be changed, what should be changed is the one in `lib/directories.sh`. The users who would prefer `d` in `lib/directories.sh` should already removed `bashmarks` from the plugin array or adjust the setup as they like, so they should be fine as is.

A possible problem is that the new users don't have access to `d` in `lib/directories.sh`. We may give a new name to the alias in `lib/directories.sh` while keeping `d`. We need to keep `d` in `lib/directories.sh` because otherwise, the existing users using them by removing `bashmarks` from the `plugin` array would be affected.

> By the way, I again found that the copyright notice of the upstream has been rewritten in this script. I think we need to check the copyright of all the plugins and themes that have the corresponding upstream.

This was solved in 0ff46f6d32e2e9387061dc5f1a6e50a773e84b83.

PR #104, opened 2019-08-21
plugins/battery: Attempt to fix #68

### LINKED ISSUE #68: "_command_exists" used in battery plugin but is not defined anywhere!
TLDR: `command_exists()` is defined in https://github.com/ohmybash/oh-my-bash/blob/429ffc1cbcd1a14f7ec0241a2e898d339eb634b6/lib/base.sh#L267, but it's not used in https://github.com/ohmybash/oh-my-bash/blob/master/plugins/battery/battery.plugin.sh which expects `_command_exists()` where `_` prefix is important.

Summarized by @Kreyren 

---

Title basically says it all...

A function "`_command_exists`" is used in the battery plugin, but is not defined anywhere else in the repository. 

*... removed personal "`cmd_found`" equivalent function, designed for older shells ...*

Kreyren:
This should handle `command_exists()` if function is not pulled from `lib/base.sh` -> plugin will define it by itself.

@antofthy Please test

Fixes: https://github.com/ohmybash/oh-my-bash/issues/68

### DIFF
--- a/plugins/battery/battery.plugin.sh
+++ b/plugins/battery/battery.plugin.sh
@@ -1,23 +1,26 @@
 #!/usr/bin/env bash
 
+# Bug: https://github.com/ohmybash/oh-my-bash/issues/68
+if ! command -v "command_exists" >/dev/null; then	command_exists() { command -v "$1" >/dev/null ;	} fi
+
 ac_adapter_connected(){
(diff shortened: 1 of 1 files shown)


Kreyren:
TODO: fix Code quality

PR #105, opened 2019-08-26
lib/directories: Add `cd` function, a clone of Zsh cd builtin command

chaifeng:
The 1 .. 9 aliases in `lib/directories.sh` work now.

    $ cd aa
    $ cd bb
    $ cd cc
    $ d
    1    ~/cc
    2    ~/bb
    3    ~/aa

    $ 3
    $ pwd
    ~/aa

    $ d
    1    ~/aa
    2    ~/cc
    3    ~/bb

### DIFF
--- a/lib/directories.sh
+++ b/lib/directories.sh
@@ -1,5 +1,30 @@
 #! bash oh-my-bash.module
 # Common directories functions
+
+# A clone of Zsh `cd` builtin command
+function cd() {
+  declare oldpwd="$OLDPWD"
+  declare -i index
+  if [[ "$#" -eq 1 && "$1" == -[1-9]* ]]; then
+    index="${1#-}"
+    if [[ "$index" -ge "${#DIRSTACK[@]}" ]]; then
+      builtin echo "cd: no such entry in dir stack" >&2
+      return 1
+    fi
+    set -- "${DIRSTACK[$index]}"
+  fi
+  builtin pushd . >/dev/null &&
+    OLDPWD="$oldpwd" builtin cd "$@" &&
+    oldpwd="$OLDPWD" &&
+    builtin pushd . >/dev/null &&
+    for ((index="${#DIRSTACK[@]}"-1; index>=1; index--)); do
+      if [[ "${DIRSTACK[0]/#~/$HOME}" == "${DIRSTACK[$index]}" ]]; then
+        builtin popd "+$index" >/dev/null || return 1
+      fi
+    done
+  OLDPWD="$oldpwd"
+}
+
 _omb_util_alias cd..='cd ../'                         # Go back 1 directory level (for fast typers)
 _omb_util_alias ..='cd ../'                           # Go back 1 directory level
 _omb_util_alias ...='cd ../../'                       # Go back 2 directory levels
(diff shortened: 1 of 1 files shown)


Kreyren:
Seems insane to overwrite coreutils function can you reference the zsh merge request?

chaifeng:
`cd` command is a builtin command for Bash & Zsh.

The Zsh `cd` command supports `-1`, `-2` these number options, as its manual explains:

> cd [ -qsLP ] [ arg ]
> cd [ -qsLP ] old new
> cd [ -qsLP ] {+|-}n
> ......
> The  third  form of cd extracts an entry from the directory stack, and changes to that directory.  An argument of the form `+n' identifies a stack entry by counting from the left of the list shown  by  the  dirs  command, starting  with  zero.   An argument of the form `-n' counts from the right.  If the PUSHD_MINUS option is set, the meanings of `+' and `-' in this context are swapped.

The Bash `cd` command doesn't support this kind of number options, so these aliases actually don't work in Bash, `alias 2='cd -2'`, `alias 3='cd -3'` ...

I really like this Zsh feature, it's very convenient.

The `cd` function I wrote doesn't break any behavior of the default Bash `cd` command, just removes the duplicate paths in the directory stack after switching directories, because Bash won't clean up for us.

nntoan:
@Kreyren @chaifeng This required a lot of testing, I think the first version needs to make it compatible must be a POXIS-shell.

Kreyren:
> @Kreyren @chaifeng This required a lot of testing, I think the first version needs to make it compatible must be a POXIS-shell.

Why do we need POXIS-shell on bash frontent?

akinomyoga:
I have rebased it on top of the latest master.

lib/directories.sh
```diff
@@ -1,5 +1,30 @@
...
+      return 1
+    fi
+    set -- "${DIRSTACK[$index]}"
+  fi
+  builtin pushd . >/dev/null &&
+    OLDPWD="$oldpwd" builtin cd "$@" &&
+    oldpwd="$OLDPWD" &&
+    builtin pushd . >/dev/null &&
+    for ((index="${#DIRSTACK[@]}"-1; index>=1; index--)); do
+      if [[ "${DIRSTACK[0]/#~/$HOME}" == "${DIRSTACK[$index]}" ]]; then
```
akinomyoga, line 21:
```suggestion
      if [[ ${DIRSTACK[0]/#~/$HOME} == "${DIRSTACK[index]}" ]]; then
```

lib/directories.sh
```diff
@@ -1,5 +1,30 @@
...
+        builtin popd "+$index" >/dev/null || return 1
+      fi
+    done
+  OLDPWD="$oldpwd"
```
akinomyoga, line 25:
```suggestion
  OLDPWD=$oldpwd
```

lib/directories.sh
```diff
@@ -22,6 +47,7 @@ _omb_util_alias 9='cd -9'
 _omb_util_alias md='mkdir -p'
 _omb_util_alias rd='rmdir'
 _omb_util_alias d='dirs -v | head -10'
+_omb_util_alias po=popd
```
akinomyoga, line 50:
```suggestion
_omb_util_alias po='popd'
```

akinomyoga:
I added further adjustments c45db79, 998c4c0, and a583e3c and squashed. I also edited commit message styles.

akinomyoga:
Currently, the existing aliases `2`..`9` defined in `lib/directories.sh` do not work. This PR tries to support it. As far as I see the implementation and try a little, this `cd` function seems to be designed very well. I would like to merge this PR later.

akinomyoga (approved):

PR #110, opened 2019-10-25
[THEME][POWERLINE] Add possibility to add /n to the end of commands

DiegoMow:
You can move the cursor to a new line after each command using the following variable:

    POWERLINE_PROMPT_NEWLINE=true

The prompt will receive a "\n" at the end. This is usefull if you navigate a lot to folders with extensive names.

### DIFF
--- a/themes/powerline/powerline.base.sh
+++ b/themes/powerline/powerline.base.sh
@@ -2,6 +2,7 @@
 
 # Define this here so it can be used by all of the Powerline themes
 THEME_CHECK_SUDO=${THEME_CHECK_SUDO:=true}
+POWERLINE_PROMPT_NEWLINE=${POWERLINE_PROMPT_NEWLINE:=false}
 
 function set_color {
   if [[ "${1}" != "-" ]]; then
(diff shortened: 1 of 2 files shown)


akinomyoga:
rebased

themes/powerline/powerline.base.sh
```diff
@@ -161,8 +162,9 @@ function __powerline_prompt_command {
   done
   [[ "${last_status}" -ne 0 ]] && __powerline_left_segment $(__powerline_last_status_prompt ${last_status})
   [[ -n "${LEFT_PROMPT}" ]] && LEFT_PROMPT+="$(set_color ${LAST_SEGMENT_COLOR} -)${separator_char}${_omb_prompt_normal}"
+  [[ "${POWERLINE_PROMPT_NEWLINE=}" == true ]] && END_PROMPT=" \n" || END_PROMPT=""
```
akinomyoga, line 165: I'm going to apply the following changes.

```suggestion
  [[ ${POWERLINE_PROMPT_NEWLINE-} == true ]] && END_PROMPT="\n" || END_PROMPT=" "
```

- I think we do not need a space before the newline. The newline was needed to make a space between the prompt and the command string, but now we have a newline after the prompt so the command is anyway in the beginning of the line.
- When we do not add the newline, we still need a space as before.
- In the conditional commands, the arguments are not needed to be quoted (except for the right-hand sides of `==`, `!=`, or `=~`)
- I think we do not need to make the variable `POWERLINE_PROMPT_NEWLINE` empty when it is not set. We can leave the unset state of the variable.

akinomyoga:
I wouldn't merge this at least now but would like to keep it as a customization example.

PR #131, opened 2020-02-15
Fix colored ls

ngaro:
- Without this fix `ls` output is not colored  on all systems
- Adds extra theme `garo` that I use (heavily based on another one)


nntoan:
Hi @ngaro, thanks for the PR. Can you please separate the fixes for LS and the theme adding into 2 pull requests instead?

I can't merge the fixes for `ls` (which is looking good) due to the quality review on the extra theme.

Thanks

PR #132, opened 2020-02-17
lib/theme-and-appearance: Fix broken ls-colors

ngaro:
* Without this fix `ls` output is not colored on all systems

### DIFF
--- a/aliases/general.aliases.sh
+++ b/aliases/general.aliases.sh
@@ -21,45 +21,46 @@
 #   1.  MAKE TERMINAL BETTER
 #   -----------------------------
 
-# Determines the use of the option `-v' on the first call
-# Ref. https://github.com/ohmybash/oh-my-bash/issues/351
-function _omb_alias_general_cp_init {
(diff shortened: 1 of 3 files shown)


lib/theme-and-appearance.sh
```diff
@@ -2,3 +2,4 @@
 
 # colored ls
 export LSCOLORS='Gxfxcxdxdxegedabagacad'
+alias ls='ls -G'
```
akinomyoga, line 5: This depends on the operating systems.

akinomyoga:
I have rebased it on top of the latest master.

There are two problems with the suggested change. The first one is that as already commented before, `ls -G` depends on the operating system. It is only available in BSD-like systems including macOS and will cause problems in Linux-like systems. The second is that the suggested alias conflicts with the other alias definition of `ls` in `lib/bourne-shell.sh`.

We need to switch the implementation depending on the `ls` available on the user's system.

akinomyoga (approved):

PR #135, opened 2020-02-22
A better progress bar

grobertson:
Fixed a bug in the existing progress bar plugin where "phase" was misnamed.

Then I couldn't leave well enough alone so I wrote a much better version.

I left the original version with my fix, and named my rewrite progress2. I wasn't sure about just rm'ing it, but my version is backwards compatible with the existing plugin.

To try it out, first source the progress2.plugin.sh script. Then just ...

`for i in {0..101}; do progress $i; done`

### DIFF
--- a/plugins/progress/example.sh
+++ b/plugins/progress/example.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+source $OSH/plugins/progress/progress2.plugin.sh
+
+for i in {0..101}; do progress $i; done
--- a/plugins/progress/progress.plugin.sh
+++ b/plugins/progress/progress.plugin.sh
@@ -1,21 +1,16 @@
 #!/usr/bin/env bash
-############################---Description---###################################
-#                                                                              #
-# Summary       : Show a progress bar GUI on terminal platform                 #
-# Support       : destro.nnt@gmail.com                                         #
-# Created date  : Aug 12,2014                                                  #
-# Latest Modified date : Aug 13,2014                                           #
-#                                                                              #
-################################################################################
+#Description:
 
-############################---Usage---#########################################
+# Summary       : Show a progress bar GUI on terminal platform
+# Support       : destro.nnt@gmail.com
+# Created date  : Aug 12,2014
+# Last Modified : Feb, 21, 2020
+# Last Modifier : FG Robertson (https://github.com/grobertson)                 
 
-# Copy below functions (delay and progress fuctions) into your shell script directly
-# Then invoke progress function to show progress bar 
 
-# In other way, you could import source indirectly then using. Nothing different
-
-################################################################################
+#Usage:
+# source into your script.
+# progress int n(1...100) string<"Phase">
 
 
 #
@@ -33,7 +28,7 @@ CURRENT_PROGRESS=0
 function progress()
 {
     PARAM_PROGRESS=$1;
-    PARAM_STATUS=$2;
+    PARAM_PHASE=$2;
 
     if [ $CURRENT_PROGRESS -le 0 -a $PARAM_PROGRESS -ge 0 ]  ; then echo -ne "[..........................] (0%)  $PARAM_PHASE \r"  ; delay; fi;
     if [ $CURRENT_PROGRESS -le 5 -a $PARAM_PROGRESS -ge 5 ]  ; then echo -ne "[#.........................] (5%)  $PARAM_PHASE \r"  ; delay; fi;
--- a/plugins/progress/progress2.plugin.sh
+++ b/plugins/progress/progress2.plugin.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
 # Copyright (c) 2020, F. Grant Robertson, https://github.com/grobertson/
 # All rights reserved.
 #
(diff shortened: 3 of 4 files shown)


grobertson (commented):
Ping. -- This has been sitting a while, and was approved a long time ago. Any plans to merge it?

Thanks!

plugins/progress/progress.plugin.sh
```diff
@@ -1,21 +1,16 @@
...
 
-# Copy below functions (delay and progress fuctions) into your shell script directly
-# Then invoke progress function to show progress bar 
 
-# In other way, you could import source indirectly then using. Nothing different
-
-################################################################################
+#Usage:
+# source into your script.
+# progress int n(1...100) string<"Phase">
```
akinomyoga, line 13: Could you explain the reasoning for changing the header style? If we should follow this style for some reason, should we also update `lib/util.sh`, which has a similar header.

plugins/progress/example.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+source $OSH/plugins/progress/progress2.plugin.sh
+
+for i in {0..101}; do progress $i; done
```
akinomyoga, line 5: typo?
```suggestion
for i in {0..101}; do progress2 $i; done
```

Also, why does the above example finally call `progress2 101`?

akinomyoga (commented):
I believe this new file should be placed in a separate directory `plugins/progress2`.

plugins/progress/progress2.plugin.sh
```diff
@@ -0,0 +1,77 @@
...
+            echo -e "\rDone!"
+        else
+            #Recompute everything every time because the shopt will update $COLUMNS
+            #   when the term size changes, and the progress bar will adjust itself!
+            _FIELD_COLS=$((COLUMNS - (${#_GAUGE_LABEL} + ${#_OPEN_GAUGE_CHAR} + ${#_CLOSE_GAUGE_CHAR} + ${#_OPEN_PCT_CHAR} + ${#_CLOSE_PCT_CHAR} + 4)))
+            # Bash doesn't 'do' floating point math and using "bc" is cheating
+            _LEFT_COUNT=$((_FIELD_COLS * _PCT_NOW / 100))
+            _RIGHT_COUNT=$((_FIELD_COLS - _LEFT_COUNT))
+            _COMPLETE=$(printf %${_LEFT_COUNT}s | tr " " $_COMPLETE_CHAR)
+            _REMAIN=$(printf %${_RIGHT_COUNT}s | tr " " $_REMAIN_CHAR)
```
akinomyoga, line 71: We would like to reduce the number of forks and execs for efficiency reasons. We require Bash 3.2 for `oh-my-bash`, so we can use `printf -v _COMPLETE ...`, and `tr` can be replaced by `_COMPLETE=${_COMPLETE// /$_COMPLETE_CHAR}`.

plugins/progress/progress2.plugin.sh
```diff
@@ -0,0 +1,77 @@
...
+    _CLOSE_PCT_CHAR=")"
+    _PCT_THEN=0
+}
+
+#
+# Description : Call with int {1..100} to display progress
+function progress2() {
+    # Set the initial vars if they haven't been set already.
+    if ((${#_REMAIN_CHAR} == 0)); then init_progress; fi
+    _PCT_NOW=$1
```
akinomyoga, line 55:
```suggestion
    local _PCT_NOW=$1
```

Please also declare `local` for the other variables whose values do not need to be preserved among the calls of `progress2`.

plugins/progress/progress2.plugin.sh
```diff
@@ -0,0 +1,77 @@
...
+            #Recompute everything every time because the shopt will update $COLUMNS
+            #   when the term size changes, and the progress bar will adjust itself!
+            _FIELD_COLS=$((COLUMNS - (${#_GAUGE_LABEL} + ${#_OPEN_GAUGE_CHAR} + ${#_CLOSE_GAUGE_CHAR} + ${#_OPEN_PCT_CHAR} + ${#_CLOSE_PCT_CHAR} + 4)))
+            # Bash doesn't 'do' floating point math and using "bc" is cheating
+            _LEFT_COUNT=$((_FIELD_COLS * _PCT_NOW / 100))
+            _RIGHT_COUNT=$((_FIELD_COLS - _LEFT_COUNT))
+            _COMPLETE=$(printf %${_LEFT_COUNT}s | tr " " $_COMPLETE_CHAR)
+            _REMAIN=$(printf %${_RIGHT_COUNT}s | tr " " $_REMAIN_CHAR)
+            echo -ne "$_GAUGE_LABEL$_OPEN_GAUGE_CHAR$_COMPLETE$_REMAIN$_CLOSE_GAUGE_CHAR $_OPEN_PCT_CHAR$_PCT_NOW%$_CLOSE_PCT_CHAR\r"
+            sleep $_STEP_DELAY
```
akinomyoga, line 73: Why do we insert a delay here? The progress bars are used to indicate the progress of an actual task that takes a long time. If we insert a delay here, the processing time will further increase. Or, is `progress2` intended to show fake progress bars?

plugins/progress/progress2.plugin.sh
```diff
@@ -0,0 +1,77 @@
...
+    shopt -s checkwinsize
+    _STEP_DELAY=0.01
+    _REMAIN_CHAR="."
+    _COMPLETE_CHAR="#"
+    _OPEN_GAUGE_CHAR='['
+    _CLOSE_GAUGE_CHAR=']'
+    _GAUGE_LABEL="Progress:"
+    _OPEN_PCT_CHAR="("
+    _CLOSE_PCT_CHAR=")"
+    _PCT_THEN=0
```
akinomyoga, line 47: We are recently thinking of namespacing variables specific to each plugin or alias. Maybe could you rename them to have the form `_OMB_PLUGIN_PROGRESS2_*`?

plugins/progress/progress2.plugin.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# See README.md
+
+#
+# Description: Called automatically if _REMAIN_CHAR is unset
+#   Call init_progress manually and then set your own values
+#   to customize the display.
+function init_progress {
+    # Make $LINES $COLUMNS available using the builtin magic of shopt!
+    shopt -s checkwinsize
```
akinomyoga, line 38: `shopt -s checkwinsize` will not take effect until the first execution of an external command or a subshell command `(...)`. For example, we can insert a line `[[ ${COLUMNS-} ]] || (true) # force update COLUMNS` after `shopt -s checkwinsize`.

akinomyoga (changes requested):

PR #148, opened 2020-04-21
New plugin to switch themes randomly

fastovst:
Inspired by zsh's plugin, this is a simple script, that takes a random theme name from .oh-my.bash/themes and activates it for a bash prompt.

### DIFF
--- a/plugins/theme/README.md
+++ b/plugins/theme/README.md
@@ -1,0 +1,13 @@
+# Theme switcher plugin
+
+Just run `theme` to load a randomly selected theme.
+
+## Example
+
+```
+❯ theme
+Switching to >>> cooperkid <<< theme...
+
+laptop @ ~   
+>_ 
+```
--- a/plugins/theme/theme.plugin.sh
+++ b/plugins/theme/theme.plugin.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+theme () {
+	THEME_TO_SWITCH=$(shuf <(ls -1 ~/.oh-my-bash/themes/ | grep -v THEMES.md) | head -1)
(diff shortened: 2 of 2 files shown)


Israel-Laguan:
It would be interesting to have an alternative to change this line in `.bashrc`

```bash
OSH_THEME="random"
```

fastovst:
> It would be interesting to have an alternative to change this line in `.bashrc`
> 
> ```shell
> OSH_THEME="random"
> ```

Yeah, that's a good idea, I think it can be achieved by creating a theme file with the same or similar code as in this plugin.

akinomyoga:
> Yeah, that's a good idea, I think it can be achieved by creating a theme file with the same or similar code as in this plugin.

I guess @Israel-Laguan has just meant to set `OSH_THEME=random` and source `.bashrc`. The value `random` is already processed as a special value in `oh-my-bash.sh`:
https://github.com/ohmybash/oh-my-bash/blob/f4a24a11de1b9c3599ef0977868793bf04d709af/oh-my-bash.sh#L134-L141

akinomyoga (changes requested):
https://github.com/ohmybash/oh-my-bash/pull/148/files#diff-240d75433f56370ee122baef5cbcb4e5e1bd817f77d5a0d9659fa3c29acb4343R6

I don't think it is a good idea to directly rewrite `.bashrc`.

https://github.com/ohmybash/oh-my-bash/pull/148/files#diff-240d75433f56370ee122baef5cbcb4e5e1bd817f77d5a0d9659fa3c29acb4343R8

This will start a new interactive session of Bash, so existing shell variables and functions without the export attribute will be lost. I think we can just drop this line.

akinomyoga:
Since this is already implemented as `OSH_THEME="random"`, let me close the PR. Anyway, thank you for opening a PR!

PR #158, opened 2020-06-11
lib, themes: Preserve users' PROMPT_DIRTRIM

dadreggors:
This change will allow a user to export the PROMPT_DIRTRIM variable in
their .bashrc and change the prompt path length.

### DIFF
--- a/lib/shopt.sh
+++ b/lib/shopt.sh
@@ -12,7 +12,7 @@ set -o noclobber
 shopt -s checkwinsize
 
 # Automatically trim long paths in the prompt (requires Bash 4.x)
-PROMPT_DIRTRIM=2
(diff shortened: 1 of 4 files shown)


dadreggors:
It seems this change could be used in the roderik them and the shopt lib as well

akinomyoga:
> It seems this change could be used in the roderik them and the shopt lib as well

Thank you for the PR! Could you also include the changes to these themes and libraries?

akinomyoga:
I've rebased it.

akinomyoga:
> > It seems this change could be used in the roderik them and the shopt lib as well
> 
> Thank you for the PR! Could you also include the changes to these themes and libraries?

I have applied the change to also `lib/shopt`, `themes/roderik`, and `themes/vscode`.

akinomyoga (approved):

PR #161, opened 2020-06-28
fix terminal title can not update error

mengruxing:


akinomyoga:
The PR title says that it tries to fix an error, but I don't see how this PR fixes an error. Could you describe what has been the error, and how the error is fixed in this PR? Also, this PR seems to change the style as well, which is unrelated to the purpose of the PR.

akinomyoga:
@mengruxing Thank you for updating the PR. I would like to confirm that you intentionally updated the branch `master` for this PR. Or have you just updated the `master` branch unrelatedly to the PR for your own uses? If it is the former, could you resolve the conflicts? If it is the latter, please let us know.

akinomyoga:
I again looked at this PR. As far as I check the commit history, I think the original patch was 7f4c2f3753ef670d0f9be44bdab995f04b16451b.

- [add `\[\e]0;\u@\h: \w\a\]` to top of PS1 to fix terminal title can no… · ohmybash/oh-my-bash@7f4c2f3](https://github.com/ohmybash/oh-my-bash/commit/7f4c2f3753ef670d0f9be44bdab995f04b16451b)

```diff
diff --git a/themes/brainy/brainy.theme.sh b/themes/brainy/brainy.theme.sh
index 90c747f..e442ea7 100644
--- a/themes/brainy/brainy.theme.sh
+++ b/themes/brainy/brainy.theme.sh
@@ -289,7 +289,7 @@ function __brainy_ps2 {
 function _omb_theme_PROMPT_COMMAND {
     exitcode="$?"

-    PS1="$(__brainy_ps1)"
+    PS1="\[\e]0;\u@\h: \w\a\]$(__brainy_ps1)"
     PS2="$(__brainy_ps2)"
 }

```

Since this PR was made based on the `master` branch, I guess the PR author has forgotten that they opened this PR based on `master` and later pushed their personal changes to the `master` branch.

On the other hand, a similar feature is available with the `xterm` plugin. I think we could leave this PR as a customization example, but the problem is that the branch of this PR `master` is already used by the author's personal settings. We wouldn't like to rewrite the author's personal setting.

I'd add the `Customization Example` label to this PR, but **the other users should look at the diff in this reply** for the original customization.

akinomyoga:
rebased.

themes/brainy/brainy.theme.sh
```diff
@@ -91,7 +91,7 @@ function ___brainy_prompt_user_info {
 	color=$_omb_prompt_bold_navy
 	if [ "${THEME_SHOW_SUDO}" == "true" ]; then
 		if [ $(sudo -n id -u 2>&1 | grep 0) ]; then
-			color=$_omb_prompt_bold_brown
+			color=$_omb_prompt_bold_purple
```
RobLoach, line 94: Seems unrelated to setting the terminal title.

themes/brainy/brainy.theme.sh
```diff
@@ -256,7 +257,7 @@ export SCM_THEME_PROMPT_CLEAN=" ${_omb_prompt_bold_green}✓${_omb_prompt_normal
 THEME_SHOW_SUDO=${THEME_SHOW_SUDO:-"true"}
 THEME_SHOW_SCM=${THEME_SHOW_SCM:-"true"}
 THEME_SHOW_RUBY=${THEME_SHOW_RUBY:-"false"}
-THEME_SHOW_PYTHON=${THEME_SHOW_PYTHON:-"false"}
+THEME_SHOW_PYTHON=${THEME_SHOW_PYTHON:-"true"}
```
RobLoach, line 260: This also seems unrelated to the tiel.

themes/brainy/brainy.theme.sh
akinomyoga: As I commented on https://github.com/ohmybash/oh-my-bash/pull/161#issuecomment-1501130240, these changes are likely to be their personal settings, but not the changes intended for the PR.

PR #167, opened 2020-07-18
Update README.md

fcatus:
The issue is still present for mac users.
```
bash-5.0$ sh -c "$(curl -fsSL https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/tools/install.sh)"
Error: Bash 4 required for Oh My Bash.
Error: Upgrade Bash and try again.
```
Since the issue appears to be relatively common, it might be wise to point to it in the readme file.

PR #168, opened 2020-07-22
Update README.md

fcatus:
See #167.


tralston:
Just curious, what's blocking this PR from being accepted? I'm a Mac user who today just ran into the same problem this PR solves, and found the solution via Google and a GH issue page.

PR #170, opened 2020-07-28
lib/grep.sh: fix expansion of GREP_OPTIONS

rrthomas:
Previously, only the first element of the array was expanded, which is
always the empty string, so the setting of GREP_OPTIONS was useless. Expand
the entire array.

### DIFF
--- a/lib/grep.sh
+++ b/lib/grep.sh
@@ -4,24 +4,24 @@ grep_flag_available() {
     echo | grep $1 "" >/dev/null 2>&1
 }
 
-GREP_OPTIONS=""
+GREP_OPTIONS=()
 
 # color grep results
 if grep_flag_available --color=auto; then
-    GREP_OPTIONS+=( " --color=auto" )
+    GREP_OPTIONS+=( "--color=auto" )
 fi
 
 # ignore VCS folders (if the necessary grep flags are available)
 VCS_FOLDERS="{.bzr,CVS,.git,.hg,.svn}"
 
 if grep_flag_available --exclude-dir=.cvs; then
-    GREP_OPTIONS+=( " --exclude-dir=$VCS_FOLDERS" )
+    GREP_OPTIONS+=( "--exclude-dir=$VCS_FOLDERS" )
 elif grep_flag_available --exclude=.cvs; then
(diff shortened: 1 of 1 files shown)


solonovamax:
I'd also like to add on two things:
- you can remove the spaces in front of the items added to `GREP_OPTIONS`.
Meaning, you'd do the following changes:
  - _
    ```diff
    @@ -10,1 +10,1 @@
    -    GREP_OPTIONS+=( "--color=auto" )
    +    GREP_OPTIONS+=( " --color=auto" )
    ```
  - _
    ```diff
    @@ -17,5 +17,5 @@
     if grep_flag_available --exclude-dir=.cvs; then
    -    GREP_OPTIONS+=( "--exclude-dir=$VCS_FOLDERS" )
    +    GREP_OPTIONS+=( " --exclude-dir=$VCS_FOLDERS" )
     elif grep_flag_available --exclude=.cvs; then
    -    GREP_OPTIONS+=( "--exclude=$VCS_FOLDERS" )
    +    GREP_OPTIONS+=( " --exclude=$VCS_FOLDERS" )
     fi
    ```
- You could swap out the initialization from `GREP_OPTIONS=""` to `GREP_OPTIONS=()`
Here's the diff for that:
  ```diff
  @@ -7,1 +7,1 @@
  -GREP_OPTIONS=""
  +GREP_OPTIONS=()
  ```

lib/grep.sh
```diff
@@ -21,7 +21,7 @@ elif grep_flag_available --exclude=.cvs; then
 fi
 
 # export grep settings
-alias grep="grep $GREP_OPTIONS"
+alias grep="grep ${GREP_OPTIONS[@]}"
```
akinomyoga, line 24:
```suggestion
alias grep="grep ${GREP_OPTIONS[*]}"
```

rrthomas:
Thanks, @akinomyoga, I have added your suggestion and rebased on current master.

akinomyoga:
Thank you! What do you think of the suggestions by @solonovamax? I think we should also apply these changes.

rrthomas:
@akinomyoga sorry, I hadn't seen these suggestions. Accepted and added; thanks @solonovamax!

akinomyoga:
Thank you! I'll merge this.

PR #171, opened 2020-08-28
lib/utils: Include functions when searching for a command

gzzsound:
The original function `type_exists` uses the option `-P` that limits the scope of the search to commands defined by files.

When debugging why the battery plugin was not working, I figured out that, also when the plugin was loaded, the function was overwritten by:

```
if ! type_exists 'battery_charge' ; then
    # if user has installed battery plugin, skip this...
    function battery_charge (){
        # no op
        echo -n
    }
fi
```

The issue is that`battery_charge` is defined as a function in the `battery` plugin and `type_exists` return `false`.

This PR expands the search scope (via the `-t` option) to:
-   alias (command is shell alias)
-   keyword (command is shell reserved word)
-   function (command is shell function)
-   builtin (command is shell builtin)
-   file (command is disk file)

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -33,7 +33,7 @@
 
 ############ Testing if packages, apps, gems, etc. are installed ###############
 
-# if type_exists 'git'; then
+# if _omb_util_command_exists 'git'; then
 #   e_success "Git good to go"
 # else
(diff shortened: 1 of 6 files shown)


akinomyoga:
Thanks for the report and fix!

PR #173, opened 2020-08-28
Set a timeout on git commands

### LINKED ISSUE #172: git_status_summary extremely slow for large repos
Large repos can turn the command prompt unresponsive for long stretches of time.

This is due to the fact that the call to 

` git status --porcelain -b`

can take a large amount of time:

```
 time  git status --porcelain -b
--- some data --
real    4m1.149s
user    3m59.189s
sys     0m1.948s
```

This makes oh-my-bash unusable.  The prompt should ideally not be delayed for more than 0.1s. See https://www.nngroup.com/articles/response-times-3-important-limits/

pokowaka:
The calculate of git_summary_status can be extremely expensive for large
repos. This results in the shell becoming unresponsive.

We now set a timeout of 0.2 seconds to make sure we always have
a responsive shell.

Fixes #172

### DIFF
--- a/themes/base.theme.sh
+++ b/themes/base.theme.sh
@@ -32,6 +32,7 @@ SCM_GIT_SHOW_CURRENT_USER=${SCM_GIT_SHOW_CURRENT_USER:=false}
 SCM_GIT_SHOW_MINIMAL_INFO=${SCM_GIT_SHOW_MINIMAL_INFO:=false}
 
 SCM_GIT='git'
+
 SCM_GIT_CHAR='±'
 SCM_GIT_DETACHED_CHAR='⌿'
 SCM_GIT_AHEAD_CHAR="↑"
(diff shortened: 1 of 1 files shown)


akinomyoga:
rebased.

akinomyoga:
Added fixes and squashed.

akinomyoga:
@pokowaka @HanJoHibender Do you have any thoughts on this?

I now think the default timeout 0.2 sec might be too short. If the timeout is too short, the prompt information will not be shown in many repositories. In fact, a user who picked this commit has increased the timeout to 1 sec:

- https://github.com/ghostrider-reborn/oh-my-bash/commit/89c727960838ee23fb078ab16f6a362dd65a058e

@ghostrider-reborn Maybe you also have a comment?

But still, there might be cases where Git information is wanted even if it takes time to show the prompt. Another approach is to get the Git information asynchronously as I mentioned in https://github.com/ohmybash/oh-my-bash/issues/172#issuecomment-1006343956.

akinomyoga:
As suggested in #6, another approach is to allow the user to disable the git prompt information for each repository.

Issue #176, opened 2020-09-27
Agnoster theme prompt broken when using pyenv


jverce:
The current prompt generated by the [agnoster](https://github.com/ohmybash/oh-my-bash/tree/master/themes/agnoster) theme is able to provide further information when the user is using a `virtualenv` environment.

The content of the prompt however cannot be properly generated for users that manage their Python runtime with [pyenv](https://github.com/pyenv/pyenv). For every shell prompt, users see the following output:
```
Python 2.7.17
basename: missing operand
Try 'basename --help' for more information.
# --> Incomplete prompt goes here <--
```

PR #177, opened 2020-09-27
themes/agnoster: Prompt broken when using pyenv (#176)

jverce:
* Make sure that the output of `python --version` is always captured.
* Refactor logic to extract Python version using `awk` instead of a
  for-loop.


nntoan (approved):
💯

Issue #137, opened 2020-03-03
set CASE_SENSITIVE="true" in .bashrc does not give case-sensitive completion


scstsfxt:
The environment is standard /bin/bash of CentOS6, and the issue is as stated in title.

grisu48:
I noticed the same behaviour and would also like to have case-sensitive completion.

tuxmontero:
same issue here, you shouldn't activate that function by default

kamichal:
Same in Arch:
```Linux xxx 5.4.36-1-MANJARO #1 SMP PREEMPT Wed Apr 29 18:47:01 UTC 2020 x86_64 GNU/Linux
```

saul365:
I attempted to solve it already tested it in Debian and Manjaro  it just changes 2 files in PR #180

Cherkah:
hi there,
what about the sensitive_case fix. cause on manjaro, konsole and yakuake do not reacts when the setting =true

akinomyoga:
This should have been solved by PR #180, which was merged on Feb 16, 2023. Closing.

PR #180, opened 2020-10-08
lib/shopt: Add support for CASE_SENSITIVE

saul365:
Solving the issue #137 changed the template and where the case sensitive attribute was defined

### DIFF
--- a/lib/shopt.sh
+++ b/lib/shopt.sh
@@ -27,7 +27,14 @@ shopt -s nocaseglob;
 ## SMARTER TAB-COMPLETION (Readline bindings) ##
 
 # Perform file completion in a case insensitive fashion
-bind "set completion-ignore-case on"
+if [[ ${CASE_SENSITIVE-} == false ]]; then
+	bind "set completion-ignore-case on"
+elif [[ ${CASE_SENSITIVE-} == true ]]; then
+	bind "set completion-ignore-case off"
+else
+	# By default, case sensitivity is disabled.
+	bind "set completion-ignore-case on"
+fi
 
 # Treat hyphens and underscores as equivalent
 bind "set completion-map-case on"
--- a/lib/shopt.sh
+++ b/lib/shopt.sh
@@ -27,9 +27,10 @@ shopt -s nocaseglob;
 ## SMARTER TAB-COMPLETION (Readline bindings) ##
 
 # Perform file completion in a case insensitive fashion
-if [[ ${CASE_SENSITIVE-} == false ]]; then
+# Note: CASE_SENSITIVE is the compatibility name
+if [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == false ]]; then
 	bind "set completion-ignore-case on"
-elif [[ ${CASE_SENSITIVE-} == true ]]; then
+elif [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == true ]]; then
 	bind "set completion-ignore-case off"
 else
 	# By default, case sensitivity is disabled.
--- a/templates/bashrc.osh-template
+++ b/templates/bashrc.osh-template
@@ -12,7 +12,7 @@ export OSH=~/.oh-my-bash
 OSH_THEME="font"
 
 # Uncomment the following line to use case-sensitive completion.
-# CASE_SENSITIVE="true"
+# OMB_CASE_SENSITIVE="true"
 
 # Uncomment the following line to use hyphen-insensitive completion. Case
 # sensitive completion must be off. _ and - will be interchangeable.


saul365:
Can it be accepted to hacktoberfest please. XD hust add the hacktoberfest-accepted label

lib/shopt.sh
```diff
@@ -27,7 +27,15 @@ shopt -s nocaseglob;
 ## SMARTER TAB-COMPLETION (Readline bindings) ##
 
 # Perform file completion in a case insensitive fashion
```
sirikid, line 29: Please update the comment

lib/shopt.sh
```diff
@@ -27,7 +27,15 @@ shopt -s nocaseglob;
...
-bind "set completion-ignore-case on"
+if [ ! -z "$CASE_SENSITIVE" ]; then
+	if [[ $CASE_SENSITIVE == "true" ]];then
+		bind "set completion-ignore-case off"
+	else
+		bind "set completion-ignore-case on"
+	fi
+else
+	bind "set completion-ignore-case on"
+fi
```
sirikid, line 38: Two branches are the same, please rewrite this

sirikid (changes requested):

lib/shopt.sh
jverce: I think the logic is fine. There's 2 branches that are the same, but that's just circumstantial (i.e. one corresponds to the "default" case and the other one for when the case sensitivity is explicitly enabled).

Maybe it can be slightly rewritten to be more clear?
```bash
if [[ ${CASE_SENSITIVE} == false ]]; then
  bind "set completion-ignore-case on"
elif [[ ${CASE_SENSITIVE} == true ]]; then
  bind "set completion-ignore-case off"
else
  # By default, case sensitivity is disabled.
  bind "set completion-ignore-case on"
fi
```

templates/bashrc.osh-template
```diff
@@ -6,7 +6,7 @@ export OSH=$HOME/.oh-my-bash
 OSH_THEME="font"
 
 # Uncomment the following line to use case-sensitive completion.
-# CASE_SENSITIVE="true"
+# export CASE_SENSITIVE="true"
```
jverce, line 9: Under which conditions is this `export` needed?
I did some tests on my side at it works just fine without modifying the scope of this variable.

jverce (changes requested):

lib/shopt.sh
jverce: @saul365 can you address this too?
The comment in the code is now inaccurate and needs to be amended.
Thanks!

jverce (changes requested):
There is still one comment to be addressed, other than that LGTM.

akinomyoga:
People are submitting similar PRs, but I don't see any progress in this PR. The original author @saul365 @nextia365 seems to be inactive. I decided to directly edit this PR.

lib/shopt.sh
```diff
@@ -27,7 +27,15 @@ shopt -s nocaseglob;
...
-bind "set completion-ignore-case on"
+# Note: CASE_SENSITIVE is the compatibility name
+if [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == false ]]; then
+	bind "set completion-ignore-case on"
+elif [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == true ]]; then
+	bind "set completion-ignore-case off"
+else
+	# By default, case sensitivity is disabled.
+	bind "set completion-ignore-case on"
+fi
```
akinomyoga, line 38: These three branches seem still redundant, but this seems to be addressed in #399 by @jjmcdn.

Also, I'd like to keep the original state when `[OMB_]CASE_SENSITIVE` is unspecified. The users may have a separate setting in e.g. `~/.inputrc`, which we wouldn't like to override.

lib/shopt.sh
akinomyoga: This is supposed to be addressed in #399 as I see the phrase

> address the remaining open suggestion in it. 

in https://github.com/ohmybash/oh-my-bash/pull/399#issue-1586451593, though I don't see the update to this line.

akinomyoga (dismissed):
There are still remaining points to discuss, but let us discuss them in #399.

akinomyoga (approved):
I've squashed some trivial commits.

lib/shopt.sh
akinomyoga: > Also, I'd like to keep the original state when `[OMB_]CASE_SENSITIVE` is unspecified.

Hmm, but OMB seems to have set `completion-ignore-case` always on so far, so it is reasonable to keep that behavior. Let me mark this conservation *Resolved*.

PR #181, opened 2020-10-14
base.theme.sh: use “command” to run git and svn everywhere

rrthomas:
This was largely done correctly, but several invocations of “git” were not
done with “command”. This means that they could pick up aliases for git,
such as “hub” or “lab”, which would still work, but (especially with “lab”)
make things really slow.

### DIFF
--- a/completions/svn.completion.sh
+++ b/completions/svn.completion.sh
@@ -79,7 +79,7 @@ function _svn_grcut()
 function _svn_info()
 {
   local what=$1 line=
-  LANG=C LC_MESSAGES=C svn info --non-interactive 2> /dev/null | \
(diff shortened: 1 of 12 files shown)


blbynum (dismissed):

rrthomas:
@akinomyoga Sounds like a good idea to me!

akinomyoga (approved):
I actually wanted to discuss whether prefixing `command ` to alias might cause conflicts. I am still not sure whether it is completely fine to put `command ` to the aliases, but I do not come up with any actual cases that might cause problems so far. Probably it would be OK to put `command ` also in the aliases. I rebased the PR on top of the latest master and added a commit 1fb516d for `command` prefix in the git aliases.

I now need to merge this PR prior to another planned PR.

PR #183, opened 2020-10-26
lib/shopt: Update CDPATH

Lennart00:
Setting CDPATH to "." as a default value leads to cd showing all possible completions for the current directory twice, which leads to a non standard tab autocompletion for bash (ble.sh) showing every directory twice which is really annoying. I searched for hours to find out why my autocompletion shows me all the possible completions just for cd twice in my current directory.

### DIFF
--- a/lib/shopt.sh
+++ b/lib/shopt.sh
@@ -50,7 +50,7 @@ shopt -s cdspell 2> /dev/null
 # This defines where cd looks for targets
 # Add the directories you want to have fast access to, separated by colon
 # Ex: CDPATH=".:~:~/projects" will look for targets in the current working directory, in home and in the ~/projec
-CDPATH="."
+CDPATH=""
 
 # This allows you to bookmark your favorite places across the file system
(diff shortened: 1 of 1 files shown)


akinomyoga:
> a non standard tab autocompletion for bash (ble.sh) showing every directory twice which is really annoying.

@Lennart00 Hi, I'm the author of `ble.sh`. I noticed this PR yesterday. If a user just want the default behavior as an empty `CDPATH`, this solution works. But if the user want to set something like `CDPATH=.:<dirpath>`, the problem arises anyway, so this PR does not solve the essential problem. Note that `.` should be included in `CDPATH` when the user want to let Bash search the current directory first before the custom path.

I fixed it in the `ble.sh` side (https://github.com/akinomyoga/ble.sh/commit/5777d7f58d7d7c426b69b9c6366a9bb5107cddc1 in master branch) so that the duplicate directory names are not generated. I believe now the problem does not occur without this PR, but @Lennart00 could you check the behavior if you have time?

Thanks

PR #184, opened 2020-11-03
base.theme.sh git branch names fix

zealotous:
base theme replaces underscores in a branch name.
hello_world -> hello-world

It's inconvenient when a end user copy-pastes the branch name from PS1 and 
 impossible to force everybody to use `-` in branch names.

### DIFF
--- a/themes/base.theme.sh
+++ b/themes/base.theme.sh
@@ -132,7 +132,7 @@ function scm_prompt_info_common {
 function git_clean_branch {
   local unsafe_ref=$(command git symbolic-ref -q HEAD 2> /dev/null)
   local stripped_ref=${unsafe_ref##refs/heads/}
-  local clean_ref=${stripped_ref//[^a-zA-Z0-9\/]/-}
+  local clean_ref=${stripped_ref//[^a-z_A-Z0-9\/]/-}
   echo $clean_ref
 }
 


zealotous:
@nntoan Could you take a look?
Please let me know if need more description.

zealotous:
@chaifeng , @surendrand, @jverce Yo Guys, can you take a look at the PR

jverce:
I'm seeing in other themes that they don't even do a cleanup of the branch name (other than the obvious removal of `refs/heads/`). For example:
- [agnoster](https://github.com/ohmybash/oh-my-bash/blob/master/themes/agnoster/agnoster.theme.sh#L249-L263)
- [hawaii50](https://github.com/ohmybash/oh-my-bash/blob/545543598b073647691db4673a728b3b1ca726d7/themes/hawaii50/hawaii50.theme.sh#L121-L134)

We might get away with removing the cleanup altogether since `git` already prevents certain characters from being used in branch names:
```
jay@jay-desktop > ~/dev/dummy_repo > master$ git checkout -b 'foo:bar'
fatal: 'foo:bar' is not a valid branch name.

✘ jay@jay-desktop > ~/dev/dummy_repo > master$ git checkout -b 'foo;bar'
Switched to a new branch 'foo;bar'

jay@jay-desktop > ~/dev/dummy_repo > foo;bar$
```

What do you think?

akinomyoga:
Ah, #200 was actually a duplicate of this PR. I have already merged #200. Sorry for the missing credit.

> I'm seeing in other themes that they don't even do a cleanup of the branch name (other than the obvious removal of `refs/heads/`).

Maybe we can think of completely removing these escapes.

akinomyoga:
As I have written in the previous reply, I failed to pick up your contribution in the repository. Nevertheless, let me thank you again for opening the PR! Let me close the PR now.

For your reference, recently we have merged another commit to allow a wider range of git branch names in PR #420.

Thank you again.

PR #186, opened 2020-11-16
Added other ways to get adapter and battery status.

DTan13:
Getting data from /sys directory will make it easier for battery plugin to work even if none of the commands tried in function [ac_adapter_connected](https://github.com/ohmybash/oh-my-bash/blob/89e505415284619377df9d8dc73f033572e24621/plugins/battery/battery.plugin.sh#L6) are available, for ex. in a docker container.

### DIFF
--- a/plugins/battery/battery.plugin.sh
+++ b/plugins/battery/battery.plugin.sh
@@ -4,55 +4,41 @@ function _omb_plugin_battery__upower_print_info {
   upower -i "$(upower -e | sed -n '/BAT/{p;q;}')"
 }
 
+## @fn ac_adapter_connected
+##   @exit 0 if the adapter is disconnected, or non-zero exit status
+##     otherwise.
 function ac_adapter_connected {
-  if _omb_util_command_exists upower;
-  then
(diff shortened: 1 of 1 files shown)


plugins/battery/battery.plugin.sh
```diff
@@ -48,6 +55,13 @@ ac_adapter_disconnected(){
   then
     WMIC Path Win32_Battery Get BatteryStatus /Format:List | grep -q 'BatteryStatus=1'
     return $?
+  else
+  	if [[ "$(cat /sys/class/power_supply/ADP0/online)" == "0" ]] ;
+      then
+        true
+ 	    else
+ 		    false
+ 	  fi
```
akinomyoga, line 64: I think this will output error messages in the terminal in the systems without the file. The same applies to the other changes.

Also these six lines are actually equivalent to just
```suggestion
  	[[ "$(cat /sys/class/power_supply/ADP0/online)" == "0" ]]
```

akinomyoga (approved):

PR #191, opened 2020-12-18
Add quote around path to handle case with space in $HOME or $OSH

nmarghetti:


oh-my-bash.sh
```diff
@@ -30,56 +30,56 @@ fi
...
 # Add all defined completions to fpath. This must be done
 # before running compinit.
 for alias in ${aliases[@]}; do
-  if is_alias $OSH_CUSTOM $alias; then
-    fpath=($OSH_CUSTOM/aliases/$alias $fpath)
-  elif is_alias $OSH $alias; then
-    fpath=($OSH/aliases/$alias $fpath)
+  if is_alias "$OSH_CUSTOM" "$alias"; then
+    fpath=("$OSH_CUSTOM/aliases/$alias" "$fpath")
+  elif is_alias "$OSH $alias"; then
```
akinomyoga, line 81: Shouldn't it be `is_alias "$OSH" "$alias"`?

tools/install.sh
```diff
@@ -38,7 +38,11 @@ main() {
   fi
 
   if [ ! -n "$OSH" ]; then
-    OSH=$HOME/.oh-my-bash
+    OSH="$HOME/.oh-my-bash"
```
akinomyoga, line 41: Quotes not mandatory here.

lib/history.sh
```diff
@@ -3,7 +3,7 @@ shopt -s histappend # append to bash_history if Terminal.app quits
 
 ## Command history configuration
 if [ -z "$HISTFILE" ]; then
-  HISTFILE=$HOME/.bash_history
+  HISTFILE="$HOME/.bash_history"
```
akinomyoga, line 6: Actually, the quote is not needed here although it is harmless. The right-hand side of variable assignments are not subject to word splitting or pathname expansions.

themes/agnoster/agnoster.theme.sh
```diff
@@ -34,7 +34,7 @@
 
 # then add the following to your .bashrc:
 
-# export THEME=$HOME/.bash/themes/agnoster-bash/agnoster.bash
+# export THEME="$HOME/.bash/themes/agnoster-bash/agnoster.bash"
```
akinomyoga, line 37: Quotes not mandatory here.

templates/bashrc.osh-template
```diff
@@ -1,5 +1,5 @@
 # Path to your oh-my-bash installation.
-export OSH=$HOME/.oh-my-bash
+export OSH="$HOME/.oh-my-bash"
```
akinomyoga, line 2: Quotes not mandatory either in this context. The word splitting and pathname expansions don't occur either in argument of the form of the variable assignments for the special assignment builtins (such as `export`, `declare`, `local`, `typeset`, and `readonly`).

oh-my-bash.sh
```diff
@@ -30,56 +30,56 @@ fi
...
 done
 
 
 is_plugin() {
   local base_dir=$1
   local name=$2
-  test -f $base_dir/plugins/$name/$name.plugin.sh \
-    || test -f $base_dir/plugins/$name/_$name
+  test -f "$base_dir/plugins/$name/$name.plugin.sh" \
+    || test -f "$base_dir/plugins/$name/_$name"
```
akinomyoga, line 44: For conditionals, isn't it better to use the special construct `[[ ... ]]` rather than quoting the arguments?

tools/install.sh
```diff
@@ -38,7 +38,11 @@ main() {
...
+  fi
+
+  if [ -z "$OSH_REPOSITORY" ]; then
+    OSH_REPOSITORY="https://github.com/ohmybash/oh-my-bash.git"
```
akinomyoga, line 45: Quotes not mandatory here.

tools/install.sh
```diff
@@ -67,23 +71,23 @@ main() {
...
   printf "${BLUE}Using the Oh My Bash template file and adding it to ~/.bashrc${NORMAL}\n"
-  cp $OSH/templates/bashrc.osh-template $HOME/.bashrc
+  cp "$OSH/templates/bashrc.osh-template" "$HOME/.bashrc"
   sed "/^export OSH=/ c\\
-export OSH=$OSH
-  " $HOME/.bashrc > $HOME/.bashrc-ombtemp
-  mv -f $HOME/.bashrc-ombtemp $HOME/.bashrc
+export OSH=\"$OSH\"
+  " "$HOME/.bashrc" > "$HOME/.bashrc-ombtemp"
+  mv -f "$HOME/.bashrc-ombtemp" "$HOME/.bashrc"
```
akinomyoga, line 90: What happens if `OSH` contains double quotations `"`? What happens if `OSH` contains newlines?

akinomyoga:
I'm thinking of utilizing the CI test configuration included in this PR (which has somehow no explanation from the PR author). It currently directly describes the test in `.github/workflow/build.yml` but we should prepare test scripts.

akinomyoga:
c5eaf4a I've reverted the changes to the codes outdated in the master. These are already fixed in the master independently. I have also reverted unnecessary quotations on the right-hand side of assignments.

akinomyoga:
I have rebased it on top of the latest master, split the commit, and fixed problems in the CI test.

akinomyoga (dismissed):
I was thinking of adjusting or separating the CI test for more comprehensive tests, but now the CI test is also suggested by @nntoan in https://github.com/ohmybash/oh-my-bash/pull/355#issuecomment-1256965559, so I decided to leave the CI-test upgrading to a new issue, #365.

Then, I think the current PR is fine as is. I'll wait for a while and will merge it if there don't seem to be problems.

akinomyoga (approved):
I have rebased it.

akinomyoga:
Merged. Thanks!

PR #196, opened 2021-02-02
plugins/git: add git pull autostash aliases

matheusramos:
Add two aliases to autostash local changes while rebasing. Those were very useful when I was using oh-my-zsh, and were missed when I migrated to oh-my-bash.


matheusramos:
@nntoan @Kreyren, could you take a look?

nntoan (approved):
Looks good to me.

PR #200, opened 2021-02-20
Add '_' to git branch name character whitelist

jonware:
Underscore characters in git branch names are displayed in the scm prompt as hypens, so e.g. my_branch becomes my-branch.

This change adds the _ character to the branch name character whitelist

### DIFF
--- a/themes/base.theme.sh
+++ b/themes/base.theme.sh
@@ -132,7 +132,7 @@ function scm_prompt_info_common {
 function git_clean_branch {
   local unsafe_ref=$(command git symbolic-ref -q HEAD 2> /dev/null)
   local stripped_ref=${unsafe_ref##refs/heads/}
-  local clean_ref=${stripped_ref//[^a-zA-Z0-9\/]/-}
+  local clean_ref=${stripped_ref//[^a-zA-Z0-9\/_]/-}
   echo $clean_ref
 }
 


akinomyoga:
Thank you! LGTM, let's merge this.

mr-pmillz:
OMG this was driving me nuts! Glad I'm not the only one :) 

Is there a way we can override things in `~/.oh-my-bash/themes/powerline/powerline.base.sh` with a file not checked out in git.

I like to customize my right powerline status bar however, after making changes to `~/.oh-my-bash/themes/powerline/powerline.base.sh` it messes up source control and I can't update oh-my-bash because of my minor changes.

Is it possible to override for instance,

```shell
user_info="${USER_INFO_SSH_CHAR}${WORKSPACE_NAME} - ${WORKSPACE_PROJECT_NAME}"
```
on line 34 of `~/.oh-my-bash/themes/powerline/powerline.base.sh` without directly modifying this file?

Like a `~/.oh-my-bash/custom` directory where users can place custom overrides for specific variables for instance, `user_info` ?

This would be a great feature!!!
Anywayz, awesome project! Cheers 🍻

akinomyoga:
@mr-pmillz Thank you for your feedback!

> Is there a way we can override things in `~/.oh-my-bash/themes/powerline/powerline.base.sh` with a file not checked out in git.
> 
> I like to customize my right powerline status bar however, after making changes to `~/.oh-my-bash/themes/powerline/powerline.base.sh`

In this case, users are usually supposed to copy the theme into `~/.oh-my-bash/custom/themes/` and edit the custom theme for yourself. For example,

```bash
$ cp -r ~/.oh-my-bash/themes/powerline ~/.oh-my-bash/custom/themes/powerline
$ edit ~/.oh-my-bash/custom/themes/powerline/powerline.base.sh
```

where `edit` is your text editor.

> Like a `~/.oh-my-bash/custom` directory where users can place custom overrides for specific variables for instance, `user_info` ?

Maybe I miss the point, but would you expect some mechanism that a file placed at e.g. `.oh-my-bash/custom/variable/user_info` automatically overrides the value of `user_info` in arbitrary places in `oh-my-bash` (which I think is technically difficult and also not well-defined)? Could you describe the idea in more detail?

Issue #113, opened 2019-10-31
update script changes PWD


Akbar30Bill:
in update script we should save `$PWD` to a variable to cd to it after the upgrade is complete

akinomyoga:
This has been solved by #286. #245 was the same issue. Thank you for the report.

PR #206, opened 2021-03-17
init, upgrader: minor fixes on upgrade procedure

miniero:
oh-my-bash.sh, tools/check_for_upgrade.sh: use "source" instead of env+bash so no child is created
tools/check_for_upgrade.sh: "echo -e" so "\c" is effective
tools/upgrade.sh: set $OLDPWD, don't exec bash (only "source" .bashrc), cd to $OLDPWD

Fixes ohmybash/oh-my-bash#113 (update script changes PWD)

### DIFF
--- a/oh-my-bash.sh
+++ b/oh-my-bash.sh
@@ -8,7 +8,7 @@ esac
 
 # Check for updates on initial load...
 if [ "$DISABLE_AUTO_UPDATE" != "true" ]; then
-  env OSH=$OSH DISABLE_UPDATE_PROMPT=$DISABLE_UPDATE_PROMPT bash -f $OSH/tools/check_for_upgrade.sh
+  OSH=$OSH DISABLE_UPDATE_PROMPT=$DISABLE_UPDATE_PROMPT source $OSH/tools/check_for_upgrade.sh
 fi
 
 # Initializes Oh My Bash
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -11,7 +11,7 @@ function _update_osh_update() {
 }
 
 function _upgrade_osh() {
-  env BASH=$OSH sh $OSH/tools/upgrade.sh
+  BASH=$OSH source $OSH/tools/upgrade.sh
   # update the osh file
   _update_osh_update
 }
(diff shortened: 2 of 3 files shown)


tools/upgrade.sh
```diff
@@ -22,6 +22,7 @@ else
 fi
 
 printf "${BLUE}%s${NORMAL}\n" "Updating Oh My Bash"
+OLDPWD="$PWD"
```
akinomyoga, line 25: `OLDPWD` is a special variable of Bash, and is automatically updated on `cd`. Thus we do not need this line.

tools/upgrade.sh
```diff
@@ -34,7 +35,8 @@ then
   printf '%s\n' '                        /____/                            '
   printf "${BLUE}%s\n" "Hooray! Oh My Bash has been updated and/or is at the current version."
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on GitHub: https://github.com/ohmybash/oh-my-bash"
-  exec bash; source $HOME/.bashrc
+  source $HOME/.bashrc
 else
   printf "${RED}%s${NORMAL}\n" 'There was an error updating. Try again later?'
 fi
+cd "$OLDPWD"
```
akinomyoga, line 42: Equivalent to

```suggestion
cd -
```

oh-my-bash.sh
```diff
@@ -8,7 +8,7 @@ esac
 
 # Check for updates on initial load...
 if [ "$DISABLE_AUTO_UPDATE" != "true" ]; then
-  env OSH=$OSH DISABLE_UPDATE_PROMPT=$DISABLE_UPDATE_PROMPT bash -f $OSH/tools/check_for_upgrade.sh
+  OSH=$OSH DISABLE_UPDATE_PROMPT=$DISABLE_UPDATE_PROMPT source $OSH/tools/check_for_upgrade.sh
```
akinomyoga, line 11: Sourcing `check_for_upgrade.sh` results in importing functions and variables inside the script. I think we need to rename these functions and variables to avoid possible conflicts.

tools/check_for_upgrade.sh
```diff
@@ -42,7 +42,7 @@ if mkdir "$OSH/log/update.lock" 2>/dev/null; then
       if [ "$DISABLE_UPDATE_PROMPT" = "true" ]; then
         _upgrade_osh
       else
-        echo "[Oh My Bash] Would you like to check for updates? [Y/n]: \c"
+        echo -e "[Oh My Bash] Would you like to check for updates? [Y/n]: \c"
```
akinomyoga, line 45:
```suggestion
        echo -n "[Oh My Bash] Would you like to check for updates? [Y/n]: "
```

akinomyoga (changes requested):

akinomyoga:
The suggested change actually overlaps with #292, which correctly implements sourcing in the current shell and addresses the problems caused by it. So, now I close this PR. Thank you for creating the PR.

Issue #192, opened 2020-12-27
Function _ac_adapter_connected()_ return '0' when AC power is not connected and _upower_ is used to request battery status


DaarkMoon:
There is an issue in battery plugin (file  _plugins/battery/battery.plugin.sh_). Function _ac_adapter_connected()_ return '0' when AC power is not connected, it should return '1'. 

The issue is only present when _upower_ is used to request battery status. The issue come from the command at line 9
`upower -i $(upower -e | grep BAT) | grep 'state' | grep -q 'charging\|fully-charged'`
The final grep pattern is to lax and catch 'discharging' and 'charging', it should be replaced with :
`upower -i $(upower -e | grep BAT) | grep 'state' | grep -q '[^(dis)]charging\|fully-charged'`

akinomyoga:
This has been solved by #209.

PR #209, opened 2021-03-28
Correction for grep regex when upower is used

DaarkMoon:
Correction for issue #192

### DIFF
--- a/plugins/battery/battery.plugin.sh
+++ b/plugins/battery/battery.plugin.sh
@@ -6,7 +6,7 @@ if ! command -v "command_exists" >/dev/null; then	command_exists() { command -v 
 ac_adapter_connected(){
   if command_exists upower;
   then
-    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q 'charging\|fully-charged'
+    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q '[^(dis)]charging\|fully-charged'
     return $?
   elif command_exists acpi;
   then
--- a/plugins/battery/battery.plugin.sh
+++ b/plugins/battery/battery.plugin.sh
@@ -6,7 +6,7 @@ if ! command -v "command_exists" >/dev/null; then	command_exists() { command -v 
 ac_adapter_connected(){
   if command_exists upower;
   then
-    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q '[:[:blank:]]charging\|fully-charged'
+    upower -i $(upower -e | grep BAT) | grep -qE 'state[:[:blank:]]*(charging|fully-charged)'
     return $?
   elif command_exists acpi;
   then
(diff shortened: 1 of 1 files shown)


plugins/battery/battery.plugin.sh
```diff
@@ -6,7 +6,7 @@ if ! command -v "command_exists" >/dev/null; then	command_exists() { command -v
 ac_adapter_connected(){
   if command_exists upower;
   then
-    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q 'charging\|fully-charged'
+    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q '[^(dis)]charging\|fully-charged'
```
akinomyoga, line 9: `[^(dis)]` doesn't work as you expect. It requires one character that is not one of `(`, `d`, `i`, `s`, or `)` before the word `charging`.

By the way, this is not related to this PR, but `\|` is a non-POSIX extension to BRE. We can use newlines to separate multiple targets of `grep`. Or, we can specify the `-E` option to `grep` and use ERE.

solonovamax: Alternatively, couldn't `-e` be used?

You could do
```bash
grep -q -e '^(dis)?charging' -e 'fully-charged'
```
which would accomplish the same thing

akinomyoga: > Alternatively, couldn't `-e` be used?

Ah, yes, you are right. That is more common.

> ```
> grep -q -e '^(dis)?charging' -e 'fully-charged'
> ```

Unless you specify the `-E` option, `(dis)?` matches with the literal five characters "`(dis)?`", which I guess is not your intent. If you specify `-E`, you could have simply used `|`. If you don't specify `-E`, you need to write it in BRE, i.e., `grep -q -e '^\(dis\)\{0,1\}charging' -e 'fully-charged'`.

But actually, what we want to match here is not "`discharging` or `charging` in the beginning of line" but "`charging`, which is not a part of `discharging`, in any position in the line". The simplest solution might be adding `grep -v`, but entirely rewriting it in Awk might be better.

plugins/battery/battery.plugin.sh
```diff
@@ -6,7 +6,7 @@ if ! command -v "command_exists" >/dev/null; then	command_exists() { command -v
 ac_adapter_connected(){
   if command_exists upower;
   then
-    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q '[^(dis)]charging\|fully-charged'
+    upower -i $(upower -e | grep BAT) | grep 'state' | grep -q '[: \t]charging\|fully-charged'
```
akinomyoga, line 9: Thank you for the update! However, there is still an issue. In the POSIX regular expressions, you cannot use `\t` in bracket expressions to represent a TAB character except for `awk`, i.e., `[: \ t]` matches one of `:`, <code>&nbsp;</code> (space), `\`, or `t`.

akinomyoga:
Thank you for updating. I have rebased the changes on top of the current `master` and also worked on the POSIX compatibility issue. @DaarkMoon Could you check the change that I added to this PR?

PR #215, opened 2021-04-23
Update functions.sh

ruslan-sem:
Fix function alias_value()

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -50,8 +50,7 @@ function open_command() {
 #    1 if it does not exist
 #
 function alias_value() {
-    alias "$1" | sed "s/^$1='\(.*\)'$/\1/"
-    test $(alias "$1")
+    [ "${BASH_ALIASES[$1]}" ] && echo ${BASH_ALIASES[$1]}
 }
 
 #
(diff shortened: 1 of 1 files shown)


lib/functions.sh
```diff
@@ -50,8 +50,7 @@ function open_command() {
 #    1 if it does not exist
 #
 function alias_value() {
-    alias "$1" | sed "s/^$1='\(.*\)'$/\1/"
-    test $(alias "$1")
+    [ "${BASH_ALIASES[$1]}" ] && echo ${BASH_ALIASES[$1]}
```
akinomyoga, line 53: - `BASH_ALIASES` is only available with Bash 4+.
-  With this PR, `alias_value` will fail on empty aliases (such as `alias empty=`). I suggest the following change:
```suggestion
    [[ ${BASH_ALIASES[$1]+set} ]] && echo ${BASH_ALIASES[$1]}
```

akinomyoga: cd79ef5 I have applied the above change and also added an implementation for Bash 3.*. In addition, I have quoted `${BASH_ALIASES[$1]}` with double quotes in case it contains glob special characters and spaces.

akinomyoga (approved):
I've rebased it. I'll wait for about a week and merge it.

akinomyoga:
I've merged it. Thanks!

PR #223, opened 2021-06-22
themes/kitsune: Include short SHA commit in prompt

altemirpacheco:
Including short SHA commit to prompt


themes/kitsune/kitsune.theme.sh
```diff
@@ -28,8 +28,13 @@ function prompt_command() {
 	PS1="\n${cyan}┌─${bold_white}[\u@\h]${cyan}─${bold_yellow}(\w)$(scm_prompt_info)\n${cyan}└─${bold_green}[\A]-${green}($BC%)${bold_cyan}-[${green}${bold_green}\$${bold_cyan}]${green} "
 }
 
+function git_short_sha() {
+  SHA=$(git rev-parse --short HEAD 2> /dev/null) && echo "$SHA"
```
akinomyoga, line 32:
```suggestion
  local SHA
  SHA=$(git rev-parse --short HEAD 2> /dev/null) && echo "$SHA"
```

themes/kitsune/kitsune.theme.sh
```diff
@@ -28,8 +28,13 @@ function prompt_command() {
...
+}
+
+
 # scm theming
-SCM_THEME_PROMPT_DIRTY=" ${red}✗"
+SCM_THEME_PROMPT_DIRTY="${bold_red}[$(git_short_sha)]${red}✗"
```
akinomyoga, line 37:
```suggestion
SCM_THEME_PROMPT_DIRTY="${bold_red}[\$(git_short_sha)]${red}✗"
```

`git_short_sha`  must be executed when `PS1` is evaluated but not when the shell starts.

akinomyoga: Also, I don't think it is the right way to include other information in the *scm dirty marker*. We should prepare another variable or include it directly in `PS1`.

PR #218, opened 2021-05-31
Fix sdkman plugin comments

sethbergman:
- Update plugin comments to match sdkman
- Update sdkman plugin README links

PR #225, opened 2021-06-29
Fix: stop SDKMAN! plugin initialisation breaking NVM

### LINKED ISSUE #224: SDKMAN! plugin init messes up NVM installation
greetings.. interestingly, I only discovered this issue when switching my Pop!_OS shell from `bash` to `zsh`.

basically, the current initialisation logic in [`sdkman.plugin.sh`](https://github.com/ohmybash/oh-my-bash/blob/4db7436384e0ddfa62911a101b69dd03a0df53fe/plugins/sdkman/sdkman.plugin.sh) causes NVM to be installed to `$HOME/.sdkman` instead of `$HOME/.nvm`.

after manually fixing this file locally and moving my installed Node versions and aliases to the correct place, everything worked as expected in both Bash and ZSH.

agreeff:
closes #224

### DIFF
--- a/plugins/sdkman/sdkman.plugin.sh
+++ b/plugins/sdkman/sdkman.plugin.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
-# Set NVM_DIR if it isn't already defined
-[[ -z "$SDKMAN_DIR" ]] && export NVM_DIR="$HOME/.sdkman"
+# Set SDKMAN_DIR if it isn't already defined
+[[ -z "$SDKMAN_DIR" ]] && export SDKMAN_DIR="$HOME/.sdkman"
 
 # Try to load sdk only if command not already available
 if ! type "sdk" &> /dev/null; then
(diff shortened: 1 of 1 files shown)


agreeff:
note: I'm closing this PR because I'm ditching this @agreeff account, moving everything over to my personal @ZaLiTHkA account instead.

it seems like this is still an issue in the current SDKMAN release though, so I will sort out a new PR from the appropriate account in the near future.

akinomyoga:
FYI, this was fixed in fd6e34ea2354e8e6f66fd7e800b40aaad1a77570 as per #218.

akinomyoga:
@ZaLiTHkA I have applied the change to the master in commit 1eea954. Thank you!

ZaLiTHkA:
nice timing.. I was just about to say, this old PR of mine did actually have one change that #218 didn't.. (:

akinomyoga:
Oh, OK! Thank you for checking #218!

PR #226, opened 2021-06-29
Debian plugin

berkoezel:


plugins/debian/debian-plugin.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
akinomyoga, line 1: The name of the plugin file is supposed to be `plugins/{PluginName}/{PluginName}.plugin.sh`, so could you rename the file as `debian{- => .}plugin.sh` if you will make this a plugin?

akinomyoga (commented):
Thank you for submitting the PR! Because it seems like this plugin only contains aliases, maybe we can put this in `aliases/` directory rather than making it a plugin. In that case, we may make a new rule on the name of the README files of aliases. What do you think?

akinomyoga:
I'm thinking of making this an `aliases` module.

akinomyoga:
I have rebased and squashed.

akinomyoga (approved):
I have rebase it on top of the latest master. I'll keep this open for a while and merge it if there are no comments.

akinomyoga:
Merged. Thanks for your contribution!

PR #232, opened 2021-07-28
Grep alias

### LINKED ISSUE #229: "gpu" alias definition in git plugin gets used when not command
When I enable the `git` plugin, doing the following
```
grep gpu *.txt
```
results in
```
grep: push: No such file or directory
grep: upstream: No such file or directory
```

This is even if more than one of the `.txt` files contains the string "`gpu`".

prehensilecode:
Fixed an error in the parameter substitution from previous pull request.

----
**edit** by @akinomyoga

Fixes #229

### DIFF
--- a/lib/grep.sh
+++ b/lib/grep.sh
@@ -21,7 +21,9 @@ elif grep_flag_available --exclude=.cvs; then
 fi
 
 # export grep settings
-alias grep="grep ${GREP_OPTIONS[*]}"
+if ((${#GREP_OPTIONS[@]} > 0)); then
+  alias grep="grep ${GREP_OPTIONS[*]}"
+fi
 
 # clean up
 unset GREP_OPTIONS
--- a/lib/grep.sh
+++ b/lib/grep.sh
@@ -1,31 +1,32 @@
 #!/usr/bin/env bash
+
 # is x grep argument available?
-grep_flag_available() {
-    echo | grep $1 "" >/dev/null 2>&1
+_omb_grep_flag_available() {
+  echo | grep $1 "" >/dev/null 2>&1
 }
 
-GREP_OPTIONS=()
+_omb_grep_options=()
 
 # color grep results
-if grep_flag_available --color=auto; then
-    GREP_OPTIONS+=( "--color=auto" )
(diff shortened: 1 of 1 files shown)


lib/grep.sh
```diff
@@ -21,7 +21,12 @@ elif grep_flag_available --exclude=.cvs; then
 fi
 
 # export grep settings
-alias grep="grep $GREP_OPTIONS"
+if [ -z ${GREP_OPTIONS+x} ]; then
+    alias grep="grep $GREP_OPTIONS"
+else
+    # no-op
+    :
+fi
```
akinomyoga, line 29: Have you tested your PR in your environment? I think the condition is inverted, and also there is a quoting issue. The above new code defines the alias when the variable `GREP_OPTIONS` is defined but has an empty value, and won't define the alias when `GREP_OPTIONS` has actual contents.

```suggestion
if [[ $GREP_OPTIONS ]]; then
    alias grep="grep $GREP_OPTIONS"
fi
```

Or you may actually always define the alias but with single quotes:

```suggestion
alias grep='grep $GREP_OPTIONS'
```

I think this second version should be the right fix because the first version has another problem that the shell metacharacters in `$GREP_OPTIONS` would be interpreted in the first version and can change the syntax structure of the command.

akinomyoga:
Another tip: You have first created a PR at #230 and then created this new one #232 just to modify one line in the PR. But you can just push the new commits to the branch associated with the PR, and the new commits will be automatically reflected in the PR page.

lib/grep.sh
prehensilecode: Yes, you're right. Just changing to using single-quotes works fine. I used double-quotes just because that was what was there, originally. Please feel free to disregard this pull request, and make your changes directly. I'm a newbie at git/github, so I'm thrashing a bit.

akinomyoga:
I've rebased it on top of the latest master. See commit  cb3599a.

akinomyoga:
Refactored variable names and function names 9b24024.

akinomyoga:
Thank you! I've merged the PR.

PR #236, opened 2021-09-04
Refactored install.sh

JamesMConroy:
Now the script uses logging functions.
Use substring removal parameter expansion to get the BASH major version.
Run sed in place on the bashrc file, instead of outputting to a file and
force copying over the original.
Wrapped loose variables in '"' to prevent globbing and word splitting.

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -91,16 +91,16 @@ export OSH=$OSH
 
   # MOTD message :)
   printf '%s' "$GREEN"
-  printf '%s\n' '         __                          __               __  '
-  printf '%s\n' '  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_ '
-  printf '%s\n' ' / __ \/ __ \   / __ `__ \/ / / /  / __ \/ __ `/ ___/ __ \'
-  printf '%s\n' '/ /_/ / / / /  / / / / / / /_/ /  / /_/ / /_/ (__  ) / / /'
-  printf '%s\n' '\____/_/ /_/  /_/ /_/ /_/\__, /  /_.___/\__,_/____/_/ /_/ '
-  printf '%s\n' '                        /____/                            .... is now installed!'
-  printf "%s\n" "Please look over the ~/.bashrc file to select plugins, themes, and options"
+  printf '%s\n' \
+    '         __                          __               __  ' \
+    '  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_ ' \
+    ' / __ \/ __ \   / __ `__ \/ / / /  / __ \/ __ `/ ___/ __ \' \
+    '/ /_/ / / / /  / / / / / / /_/ /  / /_/ / /_/ (__  ) / / /' \
+    '\____/_/ /_/  /_/ /_/ /_/\__, /  /_.___/\__,_/____/_/ /_/ ' \
+    '                        /____/                            .... is now installed!' \
+    "Please look over the ~/.bashrc file to select plugins, themes, and options"
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on GitHub: https://github.com/ohmybash/oh-my-bash"
   exec bash; source ~/.bashrc
 }
-
 
 main


tools/install.sh
```diff
@@ -54,49 +53,60 @@ main() {
...
-  cp $OSH/templates/bashrc.osh-template $HOME/.bashrc
-  sed "/^export OSH=/ c\\
-export OSH=$OSH
-  " $HOME/.bashrc > $HOME/.bashrc-ombtemp
-  mv -f $HOME/.bashrc-ombtemp $HOME/.bashrc
+  msg "${BLUE}Using the Oh My Bash template file and adding it to ~/.bashrc${NORMAL}"
+  cp "${OSH}/templates/bashrc.osh-template" "${HOME}/.bashrc"
+
+  # update bashrc template to point to osh
+  sed -i "s/^export OSH=.*$/export OSH=${OSH}" "${HOME}/.bashrc"
```
akinomyoga, line 81: `sed -i` is a GNU extension. Not every `sed` implementations support `-i`. Even for the `sed` implementations that support the option `-i`, there is variations in the behavior when `-i` is used without the option argument.

tools/install.sh
```diff
@@ -54,49 +53,60 @@ main() {
...
-    exit 1
+    die "git is not installed"
   }
   # The Windows (MSYS) Git is not compatible with normal use on cygwin
   if [ "$OSTYPE" = cygwin ]; then
     if git --version | grep msysgit > /dev/null; then
-      echo "Error: Windows/MSYS Git is not supported on Cygwin"
-      echo "Error: Make sure the Cygwin git package is installed and is first on the path"
-      exit 1
+      err "Windowr/MSYS Git is not supported on Cygwin"
```
akinomyoga, line 63: Windows

akinomyoga (commented):
There are some conficts with my PR #214.

tools/install.sh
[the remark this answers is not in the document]
JamesMConroy: Fixed

tools/install.sh
```diff
@@ -54,49 +53,60 @@ main() {
...
+      die "Error: Make sure the Cygwin git package is installed and is first on the path"
```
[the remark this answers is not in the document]
JamesMConroy: You're right it is redundant. Two arguments could work, but I think it's better to remove the new line between 'ERROR:' and the argument.

Like so

``` bash
err() {
  msg "ERROR: ${*}" >&2
}
```

tools/install.sh
JamesMConroy: I did not know that. Even so there should be a better way to update the template than moving it, passing it through sed to a new file and then overwriting the original.

JamesMConroy:
Hey @akinomyoga I didn't notice your PR when I submitted this one. If you want I can merge these changes to your branch first. That way there won't be merger conflicts, and fewer duplicate PR's

akinomyoga:
I think we don't necessarily have to merge conflicting PRs of different purposes unless the maintainer requests that, but actually some changes of our PR has the same purpose, i.e.,

> Wrapped loose variables in '"' to prevent globbing and word splitting.

For this particular change, I had thought we might combine our changes, but after another look at your changes now I find that your changes of quoting seem to be the minor changes related to the other changes. So actually now I think we don't necessarily have to merge them. Sorry for the noise.

Anyway, the maintainer of this project seems to be inactive for a long time. I'm not sure if these PRs will be reviewed or merged at some time in the future.

tools/install.sh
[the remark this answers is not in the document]
akinomyoga: OK! Nice!

tools/install.sh
akinomyoga: Yeah, actually the first `cp` is unneeded. We can directly feed `sed` with `bashrc.osh-template`.

On the other hand, I believe we should finally replace `.bashrc` using `mv` as in the current master. If we directly rewrite the file contents using `sed` or `cp`, when the rewriting failed unexpectedly by some other factors, there remains an incomplete `.bashrc` file. We should first create a complete file and then, only after it succeeded, replace the old `.bashrc` with the new complete `.bashrc`. The filesystem operation of `mv` is atomic so there is a guarantee that it will not result in a broken status even on the filesystem failure.

tools/install.sh
```diff
@@ -54,49 +53,60 @@ main() {
...
+  '         __                          __               __   ' \
+  '  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_  ' \
+  ' / __ \/ __ \   / __ `__ \/ / / /  / __ \/ __ `/ ___/ __ \ ' \
+  '/ /_/ / / / /  / / / / / / /_/ /  / /_/ / /_/ (__  ) / / / ' \
+  '\____/_/ /_/  /_/ /_/ /_/\__, /  /_.___/\__,_/____/_/ /_/  ' \
+  '                        /____/                            .... is now installed!' \
+  "Please look over the ~/.bashrc file to select plugins, themes, and options"
+
+  printf "${BLUE}${BOLD}%s${NORMAL}\n"
+    "To keep up on the latest news and updates, follow us on GitHub: https://github.com/ohmybash/oh-my-bash"
```
akinomyoga, line 95: There's another thing that I noticed now. A backslash is needed after the first line (`printf ...`) here.

I think you should additionally check your install script by actually running it. There may be still other problems.

tools/install.sh
```diff
@@ -54,49 +53,60 @@ main() {
...
-  printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on GitHub: https://github.com/ohmybash/oh-my-bash"
-  exec bash; source $HOME/.bashrc
+  printf '%s\n' \
+  '         __                          __               __   ' \
+  '  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_  ' \
+  ' / __ \/ __ \   / __ `__ \/ / / /  / __ \/ __ `/ ___/ __ \ ' \
+  '/ /_/ / / / /  / / / / / / /_/ /  / /_/ / /_/ (__  ) / / / ' \
+  '\____/_/ /_/  /_/ /_/ /_/\__, /  /_.___/\__,_/____/_/ /_/  ' \
+  '                        /____/                            .... is now installed!' \
+  "Please look over the ~/.bashrc file to select plugins, themes, and options"
```
akinomyoga, line 92: I am now thinking of picking this change.

akinomyoga (commented):
I actually have joined a maintainer team last week so now can process PRs. Today I have merged #214 and then looked at this PR again. Now actually I have doubts if we should apply all of these changes.

- I have addressed the problem of the Bash-version checking in #214.
- The problem of pathname expansions and word splitting is also already addressed in #214.
- `sed -i` is not POSIX compatible and its behavior varies between GNU sed and BSD sed. Also, we should separate the generation phase (`sed`) and the replacing phase (`mv`) for the fail safety.
- I don't see the necessity of defining and using `die`, `msg`, and `err` for this small script.

What do you think?

tools/install.sh
JamesMConroy: Go right ahead. Personally I like the consistency of `msg`, `err`, `die`, but they aren't necessary.

akinomyoga: Thank you! I have rebased and merged it.

Issue #151, opened 2020-05-12
Please update license back to MIT


robbyrussell:
Hi there,

First off, we're delighted to see that you took the initiative to create a bash-equivalent clone of Oh My Zsh. We're happy to see this pattern continue to emerge and provide the community with more options to have a great terminal experience.

Having said that -- it is quite obvious that there was some strong code and content duplication, which is okay _but_ it doesn't afford you the right to change the underlying license at the same time.

Oh My Zsh is released under the MIT License and we would ask that you update yours to match.

https://github.com/ohmyzsh/ohmyzsh/blob/master/LICENSE.txt

Thanks in advance!
Robby

larson-carter:
Also maintainers of this project here is a copy of the MIT license that Oh My Zsh is licensed under: https://opensource.org/licenses/MIT

tachyons:
Since MIT is more open than GPL V2, oh my bash team is legally derive work from oh my zsh into GPL V2, right? I couldn't find any restrictions in MIT which preventing relicensing derived work in GPLv2. 

Am I missing something ?

mcornella:
Hmm you can license the *changes* under GPL, but the underlying base copied from OMZ still is MIT, and it needs to be made explicit. There has to be a prominent mention of the project it's based on. In this case, given that there weren't MIT license headers in each Oh My Zsh file, it's hard to do this properly.

Some resources:
[1] https://www.tawesoft.co.uk/kb/article/mit-license-faq
[2] https://softwareengineering.stackexchange.com/a/105926

akinomyoga:
@nntoan There is still serious copyright infringement. The license has been changed back to the MIT license, but **the original copyright notice of the author and the contributors of Oh My Zsh is removed** which explicitly violates the license requirement. The copyright holders of the portion of codes that originated from Oh My Zsh are still the author and the contributor of Oh My Zsh, so you don't have the right to remove the copyright notice of the copyright holder and to declare it as if you are the copyright holder of the entire codes. You can just "add" your name to the copyright notice.

PR #237, opened 2021-09-04
Fix the violation of the MIT license of Oh My Zsh

akinomyoga:
The MIT license does not allow to remove the original copyright notice. See #151.


akinomyoga:
@nntoan ping

akinomyoga:
I think this has a bit higher priority, so let me merge this PR.

PR #257, opened 2021-11-15
themes/font: Fix clock format supporting current timezone

KyongSik-Yoon:
Font theme has clock format is `%I:%M:%S`
But it's not recognize current system time zone maybe looks like fixed UTC.
So I update clock format to `%T`. It's print proper time of current timezone.


nntoan:
@KyongSik-Yoon, that's intentional for general purposes. You can always override it by copying the font theme to the custom directory.

akinomyoga:
rebased

RobLoach:
Sounds like this can be overriden by setting the `THEME_CLOCK_FORMAT` variable. I'd recommend closing this one then.

akinomyoga:
Yeah, I think you are right. I thought it could be also an option to allow users to directly edit the file (since most of the users of OMB do not know how to properly do the things in Bash, and I think it's just fine for those users as far as it works). But we already have descriptions in `README`, so we do not have to provide another example to change it by modifying the theme file.

Issue #260, opened 2021-11-25
GitBash on windows: syntax error


tuxerrante:
Hi
I'm getting this error 
![image](https://user-images.githubusercontent.com/8364469/143466142-5a21dd34-f555-4355-9a53-fc4a36d068fa.png)

https://github.com/ohmybash/oh-my-bash/blob/4db7436384e0ddfa62911a101b69dd03a0df53fe/themes/cupcake/cupcake.theme.sh#L74

akinomyoga:
This seems to be a duplicate of #211.

akinomyoga:
See https://github.com/ohmybash/oh-my-bash/issues/211#issuecomment-1218903376. This is a problem with Git Bash (Git for Windows). Please update your Git Bash:

```bash
$ git update-git-for-windows
```

PR #261, opened 2021-11-25
Fixing nested command

tuxerrante:
https://github.com/ohmybash/oh-my-bash/issues/260
It may need further improvement since backtick is going to be deprecated, but this solves the issue


akinomyoga:
See #260 for the discussion.

PR #262, opened 2021-11-25
rana: Fixing #248

### LINKED ISSUE #248: Strange behavior with rana theme
I'm facing some trouble with the Rana theme that I'm not having with other themes I tested (purity, agnoster, powerline, rjorgenson, roderik) and I don't know if it is to be expected: when I press arrow up to navigate through the history, at some moment the command line stays frozen or it shows half of the previous line and I can't access the command I want.

For instance: suppose I have typed the following commands, in that exact order:

```
nano ~/.bashrc
latexmk -help
cd /home/dados/Acadêmicos
``` 
If I try to go back (arrow up) to open the .bashrc file, this is what I find:

One arrow up (OK):
```
cd /home/dados/Acadêmicos
``` 
Two arrows up (but I can't run the latexmk command):

```
cd /home/dados/Acadêmicoslatexmk -help
```
Three arrows up:
```
cd /home/dados/Acadêmicosnano ~/.bashrc
```
In other words, the command line stays paralyzed in the first command and I can't run the others.

This behavior does not appear with the other themes I tested.

I'm running bash in Ubuntu 20.04.

I'm sorry if it is a silly question, but I don't know how to solve this problem.

einwickler:
Fixes #248 

I'm definitely no bash expert, so there may be a better solution to this. I took the idea of the `ansi` function in the agnoster theme.
See my comment in #248

### DIFF
--- a/themes/rana/rana.theme.sh
+++ b/themes/rana/rana.theme.sh
@@ -32,37 +32,37 @@
 #*color7: #E5E5E5
 
 # ----------------------------------------------------------------- DEF COLOR
-RCol='\e[0m'    # Text Reset
+RCol='\[\e[0m\]'    # Text Reset
 
 # Regular
-Bla='\e[0;30m';
-Red='\e[0;31m';
-Gre='\e[0;32m';
-Yel='\e[0;33m';
-Blu='\e[0;34m';
-Pur='\e[0;35m';
-Cya='\e[0;36m';
-Whi='\e[0;37m';
+Bla='\[\e[0;30m\]';
+Red='\[\e[0;31m\]';
+Gre='\[\e[0;32m\]';
+Yel='\[\e[0;33m\]';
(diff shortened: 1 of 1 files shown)


akinomyoga:
@einwickler Actually [`themes/colours.theme.sh`](https://github.com/ohmybash/oh-my-bash/blob/4cc3ce04a4f739659a7d415f377a25935ac5e39d/themes/colours.theme.sh), which is [sourced from `oh-my-bash.sh` by default](https://github.com/ohmybash/oh-my-bash/blob/4cc3ce04a4f739659a7d415f377a25935ac5e39d/oh-my-bash.sh#L130), already defines equivalent variables. I think it is better to update the PR to use these variables instead of directly defining the escape sequences inside `rana.theme.sh`. What do you think?

akinomyoga:
I've refactored and merged the PR. Thank you for your contribution.

PR #266, opened 2021-12-15
Plugins: tmux, goenv, pyenv

chopnico:
I use these programs quite extensively and these simple plugins have been quite stable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,16 @@ For example, this line might begin to look like this:
 plugins=(git bundler osx rake ruby)
 ```
 
+##### With Conditionals
+
+You may want to control when and/or how plugins should be enabled.
+
+For example, if you want the `tmux-autoattach` plugin to only run on SSH sessions, you could employ a trivial conditional that checks for the `$SSH_TTY` variable. Just make sure to remove the plugin from the larger plugin list.
+
+``` bash
+[ "$SSH_TTY" ] && plugins+=(tmux-autoattach)
+```
+
 #### Using Plugins
 
 Most plugins (should! we're working on this) include a __README__, which documents how to use them.
--- a/plugins/tmux-autoattach/README.md
+++ b/plugins/tmux-autoattach/README.md
@@ -6,7 +6,35 @@ This tmux plugin will automatically attach a tmux session to your shell session.
 
 #### OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR
 
-| Setting | Description                                                                                                                                                      |
-|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `exit`    | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
-| `detach` (default)  | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
+| Setting            | Description                                                                                                                                                      |
+|--------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `detach` (default) | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
+| `exit`             | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
+
+## Common Enable Conditions
+
+*Note* If you prefer to manually start your desktop session using tools like `startx`, you may run into problems starting your desktop session inside of a tmux terminal. You can employ conditionals to control when this plugin should be enabled.
+
+**SSH**
+
+```bash
+[ "$SSH_TTY" ] && plugins+=(tmux-autoattach)
+```
+
+**Wayland**
+
+```bash
+[ "$DISPLAY_WAYLAND" ] && plugins+=(tmux-autoattach)
+```
+
+**X11**
+
+```bash
+[ "$DISPLAY" ] && plugins+=(tmux-autoattach)
+```
+
+**Multiple**
+
+```bash
+{ [ "$DISPLAY" ] || [ "$SSH" ]; } && plugins+=(tmux-autoattach)
+```
--- a/plugins/tmux-autoattach/README.md
+++ b/plugins/tmux-autoattach/README.md
@@ -1,8 +1,12 @@
-# tmux plugin
+# tmux-autoattach.plugin
 
-The tmux plugin will either attach or create a new tmux session for each bash session. If a session already exists, it will simply attach.
+This tmux plugin will automatically attach a tmux session to your shell session.
 
-## OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR
+## Variables
 
-| exit   | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
-| detach | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
+#### OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR
+
+| Setting | Description                                                                                                                                                      |
+|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| exit    | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
+| detach  | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
--- a/plugins/tmux-autoattach/README.md
+++ b/plugins/tmux-autoattach/README.md
@@ -8,5 +8,5 @@ This tmux plugin will automatically attach a tmux session to your shell session.
 
 | Setting | Description                                                                                                                                                      |
 |---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| exit    | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
-| detach  | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
+| `exit`    | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
+| `detach` (default)  | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
--- a/plugins/tmux-autoattach/README.md
+++ b/plugins/tmux-autoattach/README.md
@@ -1,0 +1,8 @@
+# tmux plugin
+
+The tmux plugin will either attach or create a new tmux session for each bash session. If a session already exists, it will simply attach.
+
+## OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR
+
+| exit   | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
+| detach | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
--- a/plugins/tmux-autoattach/README.md
+++ b/plugins/tmux-autoattach/README.md
@@ -36,5 +36,7 @@ This tmux plugin will automatically attach a tmux session to your shell session.
 **Multiple**
 
 ```bash
-{ [ "$DISPLAY" ] || [ "$SSH" ]; } && plugins+=(tmux-autoattach)
+if [ "$DISPLAY" ] || [ "$SSH" ]; then
+  plugins+=(tmux-autoattach)
+fi
 ```
--- a/plugins/tmux-autoattach/tmux-autoattach.plugin.sh
+++ b/plugins/tmux-autoattach/tmux-autoattach.plugin.sh
@@ -19,10 +19,7 @@ if [ "$DISPLAY" ] || [ "$WAYLAND_DISPLAY" ]; then
 		"exit")
 			_osh_plugin_tmux_autoattach_exit
 			;;
-		"detach")
-			_osh_plugin_tmux_autoattach_detach
-			;;
-		*)
+		"detach" | *)
 			_osh_plugin_tmux_autoattach_detach
 			;;
 	esac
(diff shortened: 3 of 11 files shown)


plugins/tmux/tmux.plugin.sh
```diff
@@ -0,0 +1,3 @@
+# @chopnico 2021
+
+[ -z "$TMUX"  ] && { tmux -2 -u attach || exec tmux -2 -u new-session && exit;}
```
akinomyoga, line 3: May I confirm the intent of this line? In shell syntax, `||` and `&&` have effectively the same precedence and left associativity (unlike typical languages like C/C++, JavaScript, etc.), so the above code is actually interpreted as

```bash
[ -z "$TMUX"  ] && { { tmux -2 -u attach || exec tmux -2 -u new-session; } && exit; }
```

but then the question is why don't you simply write

```bash
[ -z "$TMUX"  ] && { tmux -2 -u attach || exec tmux -2 -u new-session; } && exit
```

Possibly you have intended the following code?

```bash
[ -z "$TMUX"  ] && { tmux -2 -u attach || { exec tmux -2 -u new-session && exit; } }
```

chopnico: No, you're correct.

```bash 
[ -z "$TMUX"  ] && { tmux -2 -u attach || exec tmux -2 -u new-session; } && exit
```

I wanted the bash session to completely end once all tmux sessions were terminated including closing out my terminal.

The last example brought up an interesting question, though, surrounding detaching sessions. I could add a variable such as "OMB_TMUX_BEHAVIOR" that would allow you to configure who TMUX closes.

The suggested format, and my current preferred intent, also closes the terminal if you detach from the session. If I added something like this:

```bash
[ -z "$OMB_TMUX_BEHAVIOR" ] && export OMB_TMUX_BEHAVIOR="detach"

tmux_exit(){
	[ -z "$TMUX"  ] && { tmux -2 -u attach || exec tmux -2 -u new-session; } && exit
}
tmux_detach(){
	[ -z "$TMUX"  ] && { tmux -2 -u attach || tmux -2 -u new-session; }
}

case "$OMB_TMUX_BEHAVIOR" in
	"exit")
		tmux_exit
		;;
	"detach")
		tmux_dettach
		;;
	*)
		tmux_dettach
		;;
esac
```

We could allow for that behavior to change.

akinomyoga: Thank you for your reply! I now see your original intent.

----

> ```bash
> tmux_exit(){
> 	[ -z "$TMUX"  ] && { tmux -2 -u attach || exec tmux -2 -u new-session; } && exit
> }
> tmux_detach(){
> 	[ -z "$TMUX"  ] && { tmux -2 -u attach || tmux -2 -u new-session; }
> }
> ```

Are those functions also intended to be directly called by users?

----

> "OMB_TMUX_BEHAVIOR"

- Currently, there is no explicit naming convention for the config variables of `oh-my-bash`, but it seems like `OSH_` is the prefix used by `oh-my-bash`. The main script `oh-my-bash.sh` and a few themes use these variables. There are no existing instances of variables named `OMB_*`. The theme variable names seem to have the form `OSH_THEME_${ThemeName^^}_${ConfigName^^}`.
- `BEHAVIOR` seems too general.
- Actually, the plugin name `tmux` feels also too general. This plugin specifically attaches to a `tmux` session on the startup of the shell.

I'd suggest changing the plugin name from `tmux` to `tmux-attach`, `tmux-autoattach`, or else. For the configuration name, I'd suggest something like `OSH_PLUGIN_TMUX_ATTACH_BEHAVIOR`. Maybe we need to shift from `OSH_*` to `OMB_*`, but I feel that that should be discussed in a separate PR or issue.

----

By the way, there seems like [an option `-A` for `new-session`](https://stackoverflow.com/a/49134974/4908404) in tmux 1.8 (2013/03/29), so that we may just write `tmux -2u new -A` instead of `tmux -2 -u attach || tmux -2 -u new-session`. But, of course, that doesn't work for older `tmux`.

chopnico: Awesome.

I think I'll remove the functions. I forget we're sourcing these so maybe let's not unnecessarily clutter the session?

I like the OSH_PLUGIN_TMUX_ATTACH_BEHAVIOR, so I'll continue with that naming convention. I'll also update the name to tmux-autoattach as I think that best represents the action here.

I think I'm OK with moving to `tmux -2u new -A`. I'll make a note that it only supports 1.8 and up.

akinomyoga: > I think I'll remove the functions. I forget we're sourcing these so maybe let's not unnecessarily clutter the session?

Or, maybe we can use `_osh_plugin_tmux_autoattach_{exit,detach}` for the function names.

> I like the OSH_PLUGIN_TMUX_ATTACH_BEHAVIOR, so I'll continue with that naming convention.

OK! I think I'll later open a discussion on the naming convention.

> I'll also update the name to tmux-autoattach as I think that best represents the action here.

In that case, I'd think the appropriate name should be `OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR`. I'd like to confine the related variables in a plugin-specific namespace (`TMUX_AUTOATTACH` in this case).

> I think I'm OK with moving to tmux -2u new -A. I'll make a note that it only supports 1.8 and up.

Oh, sorry. I think I need to make it clear that I didn't intend to request shifting to `tmux -2u new -A`, but I just found it by searching. It's completely up to you; you can choose either way as you like.

chopnico: Excellent.

I think ` _osh_plugin_tmux_autoattach_{exit,detach}` would work just wonderfully. 

> In that case, I'd think the appropriate name should be OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR. I'd like to confine the related variables in a plugin-specific namespace (TMUX_AUTOATTACH in this case).

Fantastic. 

> Oh, sorry. I think I need to make it clear that I didn't intend to request shifting to tmux -2u new -A, but I just found it by searching. It's completely up to you; you can choose either way as you like.

I like the `tmux -2u new -A`. That was a wonderful find and it cleans it up a bit.

akinomyoga: > I think ` _osh_plugin_tmux_autoattach_{exit,detach}` would work just wonderfully.

OK! Thank you! I have asked the original maintainer @nntoan about the naming convention at https://github.com/ohmybash/oh-my-bash/discussions/268#discussioncomment-1866155. (I have just become a new maintainer yesterday as @nntoan seems to be busy). I'm willing to shift to `OMB_*` finally if there are no problems, so maybe these functions and variables will be renamed again later.

> I like the tmux -2u new -A. That was a wonderful find and it cleans it up a bit.

Great!

chopnico: Awesome. Welcome aboard and thank you for all your help!

plugins/tmux-autoattach/tmux-autoattach.plugin.sh
```diff
@@ -0,0 +1,29 @@
...
+	case "$OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR" in
+		"exit")
+			_osh_plugin_tmux_autoattach_exit
+			;;
+		"detach")
+			_osh_plugin_tmux_autoattach_detach
+			;;
+		*)
+			_osh_plugin_tmux_autoattach_detach
+			;;
```
akinomyoga, line 27:
```suggestion
		"detach" | *)
			_osh_plugin_tmux_autoattach_detach
			;;
```

plugins/tmux-autoattach/README.md
```diff
@@ -0,0 +1,12 @@
...
+This tmux plugin will automatically attach a tmux session to your shell session.
+
+## Variables
+
+#### OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR
+
+| Setting | Description                                                                                                                                                      |
+|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| exit    | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
+| detach  | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
```
akinomyoga, line 12:
```suggestion
| `exit`    | This will completely close out your shell session, including your terminal, but keep your tmux sessions intact. This will also close your session if you detach. |
| `detach` (default)  | This will allow you to detach from the tmux screen without closing the terminal or shell session.                                                                |
```

plugins/tmux-autoattach/tmux-autoattach.plugin.sh
```diff
@@ -0,0 +1,29 @@
+# @chopnico 2021
+#
+# tmux-autoattach.plugin.sh
+#
+# A tmux plugin that will automatically attach itself to a bash session.
+
+[ -z "$OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR" ] && export OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR="detach"
```
akinomyoga, line 7: I feel we do not have to make it an environment variable as the variable is only used in this Bash session.

akinomyoga (changes requested):
Now it looks good. I have a few minor comments.

plugins/tmux-autoattach/tmux-autoattach.plugin.sh
chopnico: Good point.

chopnico: Would you like this types of vars upper or lower case?

akinomyoga: No specific preferences, but I think it needs to be consistent with other similar variables in `oh-my-bash`, so the uppercase is fine.

plugins/tmux/tmux.plugin.sh
akinomyoga: Thank you!

plugins/tmux-autoattach/tmux-autoattach.plugin.sh
```diff
@@ -14,7 +14,7 @@ _osh_plugin_tmux_autoattach_detach() {
 	[ -z "$TMUX" ] && tmux -2u new -A
 }
 
-if [ "$DISPLAY" ] || [ "$WAYLAND_DISPLAY" ]; then
+if [ "$DISPLAY" ] || [ "$WAYLAND_DISPLAY" ] || [ "$SSH_TTY" ]; then
```
chopnico, line 17: Also added SSH. These checks are mostly here to ensure this doesn't run on a console. I typically login using xinitrc, so this works for me. But if this becomes important, maybe we add another variable that could be used to ignore these checks.

akinomyoga: I was about to write that we should make it configurable thinking that other users will be confused by the fact that it won't attach to `tmux` just by enabling the plugin, ...

But wait... such an option would just turn on/off the plugin actually. Then, I think we can just write

```bash
[ "$SSH_TTY" ] && plugins+=(tmux-autoattach)
```

at the user's side if the user would like to require conditions to turn on the plugin. We can optionally write about this type of setup in `tmux-autoattach/README.md` so that the users can specify any conditions for turning on the plugin on their own.

chopnico: This is an interesting idea. I assume we could generalize the application a little more so that this format could be used with any plugin and mention something in the larger OSH documentation. The individual plugin documentation could have a blurb about specific use cases for such conditionals.

I'm going to remove all checks and make note of this in the documentation.

akinomyoga: Ah, yes. That's a good idea. Please go ahead and update the `oh-my-bash` documentation! Thank you!

README.md
```diff
@@ -50,6 +50,16 @@ For example, this line might begin to look like this:
...
+##### With Conditionals
+
+You may want to control when and/or how plugins should be enabled.
+
+For example, if you want the `tmux-autoattach` plugin to only run on SSH sessions, you could employ a trivial conditional that checks for the `$SSH_TTY` variable. Just make sure to remove the plugin from the larger plugin list.
+
+``` bash
+[ "$SSH_TTY" ] && plugins+=(tmux-autoattach)
+```
+
```
akinomyoga, line 62: Maybe we can also add the same explanation in [templates/bashrc.osh-template](https://github.com/ohmybash/oh-my-bash/blob/master/templates/bashrc.osh-template). What do you think?

plugins/tmux-autoattach/README.md
```diff
@@ -0,0 +1,40 @@
...
+**X11**
+
+```bash
+[ "$DISPLAY" ] && plugins+=(tmux-autoattach)
+```
+
+**Multiple**
+
+```bash
+{ [ "$DISPLAY" ] || [ "$SSH" ]; } && plugins+=(tmux-autoattach)
```
akinomyoga, line 39: I usually think it is time to switch to `if ... then ... fi` when we start to use `{ ... }` for conditions.

```suggestion
if [ "$DISPLAY" ] || [ "$SSH" ]; then
  plugins+=(tmux-autoattach)
fi
```

akinomyoga (commented):
Thank you for the updates! I think this is now almost ready to be merged, but I have two additional suggestions. These are just optional, so you do not have to adopt them. I'd like to hear your opinion on them if any. Thank you!

plugins/tmux-autoattach/README.md
chopnico: I like this. There could be a possibility that some plugins require the same conditions, so this could minimize code and it makes it a little clearer.

akinomyoga:
Thank you! Is there anything else? If not, I will merge it after adding some trivial changes.

akinomyoga:
Thank you! I have merged it!

FWIW, I have noticed that you have actually worked in the `master` branch. It's fine for me, but it is generally nicer to work in a branch of a different name for a PR. Otherwise, the maintainer needs to temporarily rename or delete the true `master` branch in their local repository in order to edit the PR branch (or need to create another local repository).

Thank you again for making the PR, for your swift replies, and for your patience with a number of requests.

PR #239, opened 2021-09-24
Replace `which` with `command -v`

mchaker:
Debian has deprecated usage of `which`, as per:
https://salsa.debian.org/debian/debianutils/-/commit/3a8dd10b4502f7bae8fc6973c13ce23fc9da7efb

The recommendation is now to use `command -v`.

PR #270, opened 2021-12-24
OMB Improvements

nntoan:
- [x] bash-3.2 is the minimal version support. (_macOS users no longer need to install another version of Bash like the first version of OMB_)
- [x] Rename all `.sh` into `.bash` and add shebang line
- [ ] Refactor to use `OMB` instead of `OSH`
- [ ] Performance improvements for Git
- [ ]


lib/base.bash
```diff
@@ -326,7 +326,7 @@ httpHeaders () { /usr/bin/curl -I -L "$@" ; }             # httpHeaders:      Gr
 
 #   to mount a read-only disk image as read-write:
 #   ---------------------------------------
-#   hdiutil attach example.dmg -shadow /tmp/example.shadow -noverify
+#   hdiutil attach example.dmg -shadow /tmp/example.bashadow -noverify
```
akinomyoga, line 329: This is an oversight.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+    _omb::help
+    return 1
+  }
+
+  local command="$1"
+  shift
+
+  # Subcommand functions start with _ so that they don't
+  # appear as completion entries when looking for `omb`
+  (( $+functions[_omb::$command] )) || {
```
akinomyoga, line 14: This new file relies on many Zsh-specific features.
```suggestion
  declare -f "_omb::$command" &>/dev/null || {
```

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+
+function omb {
+  [[ $# -gt 0 ]] || {
+    _omb::help
```
akinomyoga, line 5: The shell function `_omb::help` is undefined.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+    'plugin:Manage plugins'
+    'pr:Manage Oh My Bash Pull Requests'
+    'reload:Reload the current bash session'
+    'theme:Manage themes'
+    'update:Update Oh My Bash'
+    'version:Show the version'
+  )
+
+  if (( CURRENT == 2 )); then
+    _describe 'command' cmds
```
akinomyoga, line 36: `_describe` is undefined.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+  elif (( CURRENT == 3 )); then
+    case "$words[2]" in
+      changelog) local -a refs
+        refs=("${(@f)$(cd "$OMB"; command git for-each-ref --format="%(refname:short):%(subject)" refs/heads refs/tags)}")
```
akinomyoga, line 40: `${(@f)...}` is also a Zsh feature.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+      )
+        _describe 'command' subcmds ;;
+      pr) subcmds=('clean:Delete all Pull Request branches' 'test:Test a Pull Request')
+        _describe 'command' subcmds ;;
+      theme) subcmds=('list:List themes' 'set:Set a theme in your .zshrc file' 'use:Load a theme')
+        _describe 'command' subcmds ;;
+    esac
+  elif (( CURRENT == 4 )); then
+    case "${words[2]}::${words[3]}" in
+      plugin::(disable|enable|load))
```
akinomyoga, line 57:
```suggestion
      plugin::@(disable|enable|load))
```
We also need `extglob`. There are other similar cases.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+        local -aU valid_plugins
```
akinomyoga, line 58: `-U` is Zsh specific. The same also applies to the other similar lines.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+  local -a cmds subcmds
+  cmds=(
+    'changelog:Print the changelog'
+    'help:Usage information'
+    'plugin:Manage plugins'
+    'pr:Manage Oh My Bash Pull Requests'
+    'reload:Reload the current bash session'
+    'theme:Manage themes'
+    'update:Update Oh My Bash'
+    'version:Show the version'
```
akinomyoga, line 32: We don't have any of these `_omb::xxxx` commands.

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+          # if command is "enable", remove already enabled plugins
+          [[ "${words[3]}" = enable ]] && valid_plugins=(${valid_plugins:|plugins})
+        fi
+
+        # Remove plugins already passed as arguments
+        # NOTE: $(( CURRENT - 1 )) is the last plugin argument completely passed, i.e. that which
+        # has a space after them. This is to avoid removing plugins partially passed, which makes
+        # the completion not add a space after the completed plugin.
+        local -a args
+        args=(${words[4,$(( CURRENT - 1))]})
```
akinomyoga, line 98: This is also Zsh syntax.

```suggestion
        args=("${COMP_WORDS[@]:3:COMP_CWORD}")
```

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+    'reload:Reload the current bash session'
+    'theme:Manage themes'
+    'update:Update Oh My Bash'
+    'version:Show the version'
+  )
+
+  if (( CURRENT == 2 )); then
+    _describe 'command' cmds
+  elif (( CURRENT == 3 )); then
+    case "$words[2]" in
```
akinomyoga, line 38: This is also a Zsh syntax.

```suggestion
    case "${words[2]}" in
```

lib/cli.bash
```diff
@@ -0,0 +1,108 @@
...
+        valid_plugins=(${valid_plugins:|args})
+
+        _describe 'plugin' valid_plugins ;;
+    esac
+  fi
+
+  return 0
+}
+
+compdef _omb omb
```
akinomyoga, line 108: `compdef` is also a Zsh feature.
```suggestion
complete -F _omb omb
```

- `_describe` should be properly defined for Bash `complete -F`. Such as `_describe() { COMPRELY+=("$(compgen -W '"$1"' -- "${COMP_WORDS[COMP_CWORD]}")"); }`.
- The array variable `words` in the body of `_omb` should be replaced by `COMP_WORDS`. Also, be careful that the index is shifted by 1 in Zsh compared to Bash; The array index of Zsh starts from 1 while that of Bash starts from 0.
- The variable `CURRENT` should be replaced by `COMP_CWORD+1`.

themes/rjorgenson/rjorgenson.theme.bash
```diff
@@ -49,8 +49,8 @@ function is_integer() { # helper function for todo-txt-count
 }
 
 todo_txt_count() {
-    if `hash todo.sh 2>&-`; then # is todo.sh installed
-        count=`todo.sh ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }'`
+    if `hash todo.bash 2>&-`; then # is todo.bash installed
+        count=`todo.bash ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }'`
```
akinomyoga, line 53: `todo.sh` is not a part of `oh-my-bash`, so we can't simply replace them.

themes/brainy/brainy.theme.bash
```diff
@@ -136,10 +136,10 @@ ___brainy_prompt_ruby() {
 
 ___brainy_prompt_todo() {
 	[ "${THEME_SHOW_TODO}" != "true" ] ||
-	[ -z "$(which todo.sh)" ] && return
+	[ -z "$(which todo.bash)" ] && return
 	color=$bold_white
 	box="[|]"
-	info="t:$(todo.sh ls | egrep "TODO: [0-9]+ of ([0-9]+)" | awk '{ print $4 }' )"
+	info="t:$(todo.bash ls | egrep "TODO: [0-9]+ of ([0-9]+)" | awk '{ print $4 }' )"
```
akinomyoga, line 142: The same as the previous comment. `todo.sh` is not a part of `oh-my-bash`.

tools/git-prompt.bash
```diff
@@ -9,7 +9,7 @@
 #
 #    1) Copy this file to somewhere (e.g. ~/.git-prompt.sh).
```
akinomyoga, line 10:
```suggestion
#    1) Copy this file to somewhere (e.g. ~/.git-prompt.bash).
```

tools/install.bash
```diff
@@ -27,23 +27,21 @@ main() {
...
   # Checks the minium version of bash (v4) installed, 
-  # stops the installation if check fails
+  # tell the user Bash 4 is nice to have
   if [ -n $BASH_VERSION ]; then
      bash_major_version=$(echo $BASH_VERSION | cut -d '.' -f 1)
      if [ "${bash_major_version}" -lt "4" ]; then
-        printf "Error: Bash 4 required for Oh My Bash.\n"
-        printf "Error: Upgrade Bash and try again.\n"
-        exit 1
+        printf "Warning ;) Bash >=4 is no longer required for Oh My Bash. But it's cool to have, do you want to try?\n"
```
akinomyoga, line 34: We anyway should check for Bash 3.2.

akinomyoga (changes requested):
Thank you for making the PR! In addition to the code comments,

- I would like to notice that this major change will probably introduce conflicts to many existing PRs. We might need to notify the PR authors. Maybe we can resolve the conflicts at our side.
- I think maybe we should change `OMB` to `OMB_PATH`, `OMB_DIR`, `OMBDIR`, or something more descriptive and unlikely to conflict with the variables defined by users and other people.
- Just to confirm, is the shebang `#!/usr/bin/env bash` in plugins/themes/completions intended for text editors to correctly detect the language?
- We want to import old configuration names as I have described in [a reply at \#268](https://github.com/ohmybash/oh-my-bash/discussions/268#discussioncomment-1866155): e.g. `: "${OMB_CUSTOM=${OSH_CUSTOM}}"`.
- We want to keep `oh-my-bash.sh` as a symbolic link to `oh-my-bash.bash` for a while until the users complete the migration.

akinomyoga:
If you are busy to address all of these comments, maybe I can work on them when I have time.

> * [ ]  Refactor to use `OMB` instead of `OSH`
> * [ ]  Performance improvements for Git

I feel we should separate the PRs for each point because I am likely to make many comments for each change.

akinomyoga:
I found that the extension of the main script of `oh-my-zsh` is `.sh` despite that the extensions of the other scripts are `.zsh`. This is probably because `oh-my-zsh` already contains the shell name so `oh-my-zsh.zsh` would appear to be verbose. In this sense, `oh-my-bash.bash` also feels wordy/redundant.

nntoan:
> 

Agreed.

akinomyoga:
@nntoan Thanks. I think I need to share my current thinking for the plan to process this PR.

Currently, the author of PRs #288, #277, and #239 are working on polishing them up. On the other hand, this PR includes large changes `*.{sh => bash}` which will cause conflicts with these PRs. The change `*.{sh => bash}` in this PR is just renaming so trivial to rework it. On the other side, reworking these PRs #288, #277, and #239 requires effort. So I need to process the rename `*.{sh => bash}` after these PRs are merged.

Nevertheless, maybe the other changes in this PR can be processed separately (maybe in separate PRs). I'll take a look

lib/base.bash
akinomyoga:
```suggestion
#   hdiutil attach example.dmg -shadow /tmp/example.shadow -noverify
```

themes/brainy/brainy.theme.bash
akinomyoga, line 142:
```suggestion
	info="t:$(todo.sh ls | egrep "TODO: [0-9]+ of ([0-9]+)" | awk '{ print $4 }' )"
```

themes/rjorgenson/rjorgenson.theme.bash
akinomyoga, line 53:
```suggestion
    if `hash todo.sh 2>&-`; then # is todo.bash installed
        count=`todo.sh ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }'`
```

akinomyoga:
I've separated the commits.

akinomyoga:
https://github.com/ohmybash/oh-my-bash/pull/270/commits/af20bfbf7d6640f521c06b0430558e2fca6f3edf I'm not sure if we should use `#!/usr/bin/env bash` as the shebang of all these script files that are supposed to be sourced from other script files. The point of shebang is to tell `exec` how to execute the file, but these script files are not intended to be executed as executable files, so they will fail if it is directly executed using Bash. Here's a related discussion:

- [shell - Do you need shebang in all bash scripts? - Stack Overflow](https://stackoverflow.com/questions/7615430/do-you-need-shebang-in-all-bash-scripts)

What's the point of adding `#!/usr/bin/env bash` to those files that are intended to be sourced from interactive session of Bash? I searched what people do for this case and found an interesting solution using `/bin/echo`:

- [linux - What are all options to execute and source a bash file? Why shebang is not checked? - Stack Overflow](https://unix.stackexchange.com/a/432080/121088)

I suggest adding `#!/bin/echo bash: usage: source `, if we want to stick with shebang, or simply putting a normal comment `# bash script to source`. With the former suggestion, we can prevent the script from being directly executed so the script becomes safer. But actually, it shouldn't be a problem because the executable bit is unset in the file permission for these script files. Also, it is a hack and its intent is not clear at a glance. i actually prefer the latter.

akinomyoga (approved):

akinomyoga:
- 2414c46 (Bash version check in `install.sh`) is included in PR #295.
- b066746 (shebang `#!/usr/bin/env bash` for scripts to source) is dropped.

akinomyoga:
Cross-Reference: This PR was mentioned in #591. (It is sad that the automatic cross-referencing doesn't work with GitHub Discussion)

akinomyoga:
I will pick the first three commits in the master.

obnoxxx (commented):
I think this PR does very many things at once.

Ideaaly, for my taste, a big feature like the introduction of a òmb`cli tool should not be hidden under a general ïmprovements subject but would deserve a dedicated PR title and commit message header.

Just my 2 cents as a user who is also a developer.

akinomyoga:
Yes, we already *cropped* commits from this PR several times...

I now picked three more commits from this PR and pushed them to the master branch. However, the implementation is actually incomplete. We now have a skeleton `omb` implementation in the `master` branch, but we need to implement each subcommand. Contributions are welcome.

To clarify the current status of this PR, although there is one remaining commit in the PR (766692e735d9a6ee4117b419060e54e16abd1757), there are also two commits 4a608c7ea1e73a73981e7fde75713732f73f350d and 2e90f1fd4461fa33d7a92ca29e39c9874ce401bf that are *temporarily dropped* to avoid unnecessary conflicts to any changes to the codebase.

PR #273, opened 2021-12-27
Add conda python venv prompt

Honghe:


themes/font/font.theme.sh
```diff
@@ -37,12 +37,24 @@ THEME_CLOCK_FORMAT=${THEME_CLOCK_FORMAT:-"%I:%M:%S"}
...
+
+    if [[ -n "${CONDA_DEFAULT_ENV}" ]]; then
+        python_venv="${CONDA_DEFAULT_ENV}"
+    elif [[ -n "${VIRTUAL_ENV}" ]]; then
+        python_venv=$(basename "${VIRTUAL_ENV}")
+    fi
+
+    [[ -n "${python_venv}" ]] && echo -e "$VIRTUALENV_THEME_PROMPT_PREFIX$python_venv$VIRTUALENV_THEME_PROMPT_SUFFIX"
+}
+
```
akinomyoga, line 51: Thanks for the PR! There seems like already a function [`python_version_prompt `](https://github.com/ohmybash/oh-my-bash/blob/39af832b766d83510a2e1a164c2ac9e9b37cb862/themes/base.theme.sh#L388-L408). Why don't you use `python_version_prompt` instead of redefining a similar function?

PR #284, opened 2022-01-05
Do not overwrite backup bashrc

### LINKED ISSUE #115: Install script overwrites .bashrc file
## Steps to reproduce

- start with a non empty .bashrc file (paths, aliases, etc...)
- run the install script

## Expected result 

The oh my bash contents are append to the end of the file.

## Actual result

The .bashrc is overwritten

*Note : Maybe it is good practise to have the aliases and paths in ~/.profile in which case I apologize for my ignorance, but a little warning would go a long way*

### LINKED ISSUE #267: oh-my-bash: We should not overwrite existing `.bashrc.pre-oh-my-bash`
couldnt run commands like `iwconfig` or `ipconfig` without being root after installing, tried removing the bash file and remaking it myself and never ended up working. it annoys me, i dont want to be root everytime i poweroff my machine which is not necessary, basic commands like wireshark, owasp-zap, sqlmap, iwconfig, dir, pwd, ifconfig, etc etc will resolve in the error `command not found` i have not installed anything other than this bash theme and apparently it likes to kill shells

akinomyoga:
This fixes the issue reported at #115 and #267. We should not overwrite the existing backup file. In this PR, I take the approach suggested by @alexpfx in #115; The backup filename now contains the timestamp. Also, in recovering the original bashrc, we search for the latest backed-up `.bashrc` that does not contain `source oh-my-bash.sh`. Additionally, the code of `chsh -s bash` in uninstallation, which makes sense only for `oh-my-zsh`, has been removed.

----
Resolve #267 / Resolve #115

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -4,7 +4,7 @@ function bash_stats() {
 }
 
 function uninstall_oh_my_bash() {
-  env OSH=$OSH sh $OSH/tools/uninstall.sh
+  source "$OSH"/tools/uninstall.sh
 }
 
 function upgrade_oh_my_bash() {
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -143,8 +143,9 @@ _omb_install_main() {
 
   printf "${BLUE}Looking for an existing bash config...${NORMAL}\n"
   if [[ -f ~/.bashrc || -h ~/.bashrc ]]; then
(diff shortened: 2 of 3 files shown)


jeffmcneill:
This new version should be released, since all distributions still have the old 7.1.0 version, and hence this mindlessly damaging overwrite without notice or backup of .bashrc is still affecting everyone who is downloading the latest release.

PR #286, opened 2022-01-05
tools/upgrade: Return to the original working directory after upgrading

### LINKED ISSUE #245: $PWD remains ~/.oh-my-bash
image

akinomyoga:
Fixes #245. See also the discussion in 028d2537392a5e1ec8166600105840e107eb65c9.

### DIFF
--- a/tools/upgrade.sh
+++ b/tools/upgrade.sh
@@ -22,9 +22,8 @@ else
 fi
 
 printf "${BLUE}%s${NORMAL}\n" "Updating Oh My Bash"
-cd "$OSH"
-if git pull --rebase --stat origin master
-then
+
+if git -C "$OSH" pull --rebase --stat origin master; then
   printf '%s' "$GREEN"
   printf '%s\n' '         __                          __               __  '
   printf '%s\n' '  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_ '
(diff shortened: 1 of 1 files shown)


nntoan (approved):

tools/upgrade.sh
```diff
@@ -34,7 +33,7 @@ then
   printf '%s\n' '                        /____/                            '
   printf "${BLUE}%s\n" "Hooray! Oh My Bash has been updated and/or is at the current version."
   printf "${BLUE}${BOLD}%s${NORMAL}\n" "To keep up on the latest news and updates, follow us on GitHub: https://github.com/ohmybash/oh-my-bash"
-  exec bash; source $HOME/.bashrc
+  exec bash; source ~/.bashrc
```
nntoan, line 36: @akinomyoga We can change like this to replace the current shell instead of adding a new shell on top of the current.

```suggestion
  exec -l bash;
```

akinomyoga: > We can change like this to replace the current shell instead of adding a new shell on top of the current.

`exec bash` already does it.

- Is there any reason for marking the replaced process as the login shell by `-l`?
- `source ~/.bashrc` is a fallback code in case that `bash` is not found in `PATH` and `exec bash` fails.

akinomyoga: Since this is actually unrelated to the change of this PR, I now merge this PR. I address this point in another PR #292.

akinomyoga:
Hmm, actually #113 is also the same issue as #245 so is also solved by this PR.

PR #290, opened 2022-01-08
Add logging functions (by @Kreyren)

akinomyoga:
These are some logging functions that were suggested in #73 but are pending. The related discussion starts from https://github.com/ohmybash/oh-my-bash/pull/73#issuecomment-1002121916. See also Discussion #280.


akinomyoga:
rebased

PR #292, opened 2022-01-12
Upgrade/Uninstall/Install OMB in the interactive shell process

akinomyoga:
This PR allows installing/upgrading/uninstalling OMB within the interactive shell process. This was mentioned in https://github.com/ohmybash/oh-my-bash/commit/028d2537392a5e1ec8166600105840e107eb65c9#commitcomment-62954551 and also addressed in [#206](https://github.com/ohmybash/oh-my-bash/pull/206/files#diff-4b98dbe69610bd8f3e7440e55255e2fd22da4659b509cefbe976c3e03d6458e0L37-R38) (incompletely/incorrectly).

- To avoid multiple PROMPT_COMMAND handlers after updating and loading OMB again, the name of `PROMPT_COMMAND` handlers are made consistent over different themes, and its registration is refactored.
- Protecting the executions of `grep`, `sed`, `mkdir`, etc. inside installing/upgrading scripts from aliases by prefixing them by `command`.
- Namespacing shell functions and shell variables used in installing/upgrading scripts to avoid conflicts.
- Other various fixes.

### DIFF
--- a/lib/functions.sh
+++ b/lib/functions.sh
@@ -8,7 +8,7 @@ function uninstall_oh_my_bash() {
 }
 
 function upgrade_oh_my_bash() {
-  env OSH=$OSH sh $OSH/tools/upgrade.sh
+  source "$OSH"/tools/upgrade.sh
 }
 
 function take() {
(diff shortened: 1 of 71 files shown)


akinomyoga:
Oh, actually this PR overlaps with #206.

PR #300, opened 2022-01-30
Refactored rjorgenson theme to reduce subshells

JamesMConroy:
- Removed several unnecessary branches
- Improved indentation
- No longer pipes grep output into awk


themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
...
-
-my_prompt_char() {
-    if [[ $OSTYPE =~ "darwin" ]]; then
-        echo "${BRACKET_COLOR}➞  ${normal}"
-    else
-        echo "${BRACKET_COLOR}➞ ${normal}"
-    fi
+    CHAR=$(scm_char)
+    [ "${CHAR}" = "${SCM_NONE_CHAR}" ] \
+        || echo "${BRACKET}[${CHAR}${BRACKET}][${STRING_COLOR}$(scm_prompt_info)${BRACKET}]$normal"
```
akinomyoga, line 67:
```suggestion
    [ "${CHAR}" = "${SCM_NONE_CHAR}" ] ||
        echo "${BRACKET}[${CHAR}${BRACKET}][${STRING_COLOR}$(scm_prompt_info)${BRACKET}]$normal"
```

This might depend on the coding style, but we can place `||` at the end of the line so that Bash knows that the line is continued without adding an extra backslash.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
...
 
 _omb_theme_PROMPT_COMMAND() {
+    my_host="${STRING_COLOR}\h${normal}";
+    my_user="${STRING_COLOR}\u${normal}";
+    my_path="${STRING_COLOR}\w${normal}";
 
-    my_ps_host="${STRING_COLOR}\h${normal}";
-    my_ps_user="${STRING_COLOR}\u${normal}";
-    my_ps_root="${bold_red}\u${normal}";
-    my_ps_path="${STRING_COLOR}\w${normal}";
```
akinomyoga, line 84: I appreciate these shorter names, but in addition, we should make them `local` to avoid conflicts with the global variables that might be used by the user. `line2` should also be made `local`.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
...
-        TITLEBAR="\[\033]0;\w\007\]"
-        ;;
-        *)
-        TITLEBAR=""
-        ;;
+        xterm*) TITLEBAR="\[\033]0;\w\007\]" ;;
 esac
 
-PS3=">> "
+if [[ $OSTYPE =~ "darwin" ]]; then
```
akinomyoga, line 27:
```suggestion
if [[ $OSTYPE == *darwin* ]]; then
```

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
...
+    echo "${BRACKET}➞  ${normal}"
```
akinomyoga, line 28: This seems to output the above string on the startup of the shell session. Is this intended? The original code was used to put the string in the prompt.

Ah, OK. I think you have forgot to replace `echo` with `PROMP_CHAR=`. We can move this code just below the assignment of the default value to `PROMP_CHAR`.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
...
+fi
 
 __my_rvm_ruby_version() {
-    local gemset=$(echo $GEM_HOME | awk -F'@' '{print $2}')
-  [ "$gemset" != "" ] && gemset="@$gemset"
-    local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $2}')
-    local full="$version$gemset"
-  [ "$full" != "" ] && echo "${BRACKET_COLOR}[${STRING_COLOR}$full${BRACKET_COLOR}]${normal}"
+    local gemset ; gemset=$(echo "${GEM_HOME}" | awk -F'@' '{print $2}')
```
akinomyoga, line 32: I'm interested in why you have separated the declaration and the assignments here as `local gemset ; gemset=...`.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
...
+# XXX do we need/want to integrate with todo.sh? We don't provide it and I
+# can't find a version online that accepts ls as an input
 todo_txt_count() {
     if _omb_util_command_exists todo.sh; then # is todo.sh installed
-        count=$(todo.sh ls | egrep "TODO: [0-9]+ of ([0-9]+) tasks shown" | awk '{ print $4 }')
-        if is_integer $count; then # did we get a sane answer back
-            echo "${BRACKET_COLOR}[${STRING_COLOR}T:$count${BRACKET_COLOR}]$normal"
+        count=$(todo.sh ls \
+          | awk '/TODO: [0-9]+ of ([0-9]+) tasks shown/ { print $4 }')
+        if [ "${count}" -eq "${count}" ] &> /dev/null
```
akinomyoga, line 57: Why don't you use `is_integer`? I haven't tested it, but is there any problem with using `is_integer`? If there is any, we should rather fix `is_integer` instead of inlining the function body of `is_integer`.

akinomyoga (commented):
Thank you for your contribution! Could you check these comments?

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: I did not know BASH could do that.

I still prefer the the explicit line continue. It clearly indicates that the line starting with `||` or `&&` is a continuation of the test.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -3,7 +3,7 @@
 
 # set colors for use throughout the prompt
 # i like things consistent
-BRACKET_COLOR=${blue}
+BRACKET=${blue}
```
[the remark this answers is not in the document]
JamesMConroy: Just to make the PS1 assignment more readable. I don't really think we need this variable, but maybe we could rename it to something like `B_COLOR`?

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: You are correct. I don't have/use a Mac so I didn't catch the issue.

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: Shellcheck SC2155: Declare and assign separately to avoid masking return values.

You don't need to always separate the declaration and assignment, just when assigning results of subshells or external commands.

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: Why check the return code of a single use function when we can just compare the input to itself.

I don't see a reason to keep it around unless we need to check in multiple places.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -14,88 +14,82 @@ SCM_THEME_PROMPT_CLEAN=" ${bold_green}✓${normal}"
 SCM_GIT_CHAR="${STRING_COLOR}±${normal}"
 SCM_SVN_CHAR="${bold_cyan}⑆${normal}"
 SCM_HG_CHAR="${bold_red}☿${normal}"
+PROMP_CHAR="${BRACKET}➞ ${normal}"
```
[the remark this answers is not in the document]
JamesMConroy: fixed

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: fixed

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: fixed

themes/rjorgenson/rjorgenson.theme.sh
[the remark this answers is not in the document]
akinomyoga: Are shorter names or abbreviations more readable in your theory?

To be honest, I don't think we should clutter with this kind of short name variables in the global scope which is primarily for interactive uses in interactive shell sessions, so I'm actually thinking of gradually moving to the variable names with namespaces like `_omb_theme_rjorgenson_border_color` or else, which seems to be completely opposite direction to this change.

themes/rjorgenson/rjorgenson.theme.sh
akinomyoga: Thanks, but in this case, we do not use the exit status, and also we will never turn on `set -e` in the interactive shells. I must say this is some false-positive warning by shellcheck. But maybe we should do it to work around shellcheck.

themes/rjorgenson/rjorgenson.theme.sh
akinomyoga: I actually wanted to make the style consistent across the entire codebase, but it seems to be not the case at the moment (I searched them and found 36 cases of `command1 Op <newline> command2`, 18 cases of `command1 Op \<newline> command2`, and 4 cases of `command1 \<newline> Op command2`, where Op is `||` or `&&`). So that's fine if this is your preference. Nevertheless, I don't think the other patterns are less clear than this style. The line continuation is visibly indicated by the indentation after all.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -15,87 +15,82 @@ SCM_GIT_CHAR="${STRING_COLOR}±${normal}"
...
+    local my_host="${STRING_COLOR}\h${normal}";
+    local my_user="${STRING_COLOR}\u${normal}";
+    local my_path="${STRING_COLOR}\w${normal}";
 
-    my_ps_host="${STRING_COLOR}\h${normal}";
-    my_ps_user="${STRING_COLOR}\u${normal}";
-    my_ps_root="${bold_red}\u${normal}";
-    my_ps_path="${STRING_COLOR}\w${normal}";
-
+     line2="${BRACKET}└─$(todo_txt_count)${PROMPT_CHAR}"
```
akinomyoga, line 76:
```suggestion
     local line2="${BRACKET}└─$(todo_txt_count)${PROMPT_CHAR}"
```

themes/rjorgenson/rjorgenson.theme.sh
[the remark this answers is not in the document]
akinomyoga: Thanks for updating.

themes/rjorgenson/rjorgenson.theme.sh
akinomyoga: Because the intent of `[ "${count}" -eq "${count}" ] &> /dev/null` is not so obvious. We might have added a comment here if necessary, but if we can *name* the code fragment by making it a function, that's better than a comment.

themes/rjorgenson/rjorgenson.theme.sh
akinomyoga: Anyway, let's keep the current style. I wouldn't request changing the style.

themes/rjorgenson/rjorgenson.theme.sh
akinomyoga: This is also acceptable as the workaround for shellcheck.

themes/rjorgenson/rjorgenson.theme.sh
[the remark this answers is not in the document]
JamesMConroy: When lines are 150+ columns long I definitely have a harder time reading them.

I like the idea of `OMB_THEME_BRACKET_COLOR`, maybe change the var to that and then do something like

```
BRACKET_R="${OMB_THEME_BRACKET_COLOR}]"
BRACKET_L="${OMB_THEME_BRACKET_COLOR}["
```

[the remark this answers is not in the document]
akinomyoga: Yeah, that is what I also thought of. We can first import the variables locally in the functions with shorter names, and then we can use it in the actual code in the same function.

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: I still don't think it's the best idea, unless we use the function in more places, but I added the function back in.

akinomyoga: Maybe the existing approach is not the best approach, but I don't think directly writing this rarely used trick inline is the *best idea* either. Then, I don't see any reason to move from one not-best approach to another.

themes/rjorgenson/rjorgenson.theme.sh
akinomyoga: I have actually suggested surrounding the string by `*`s because that is equivalent to the existing matching by the regular expression (i.e. the regular expression "darwin" matches any substring of the matched string). But in the updated code, I actually see that `darwin` is still quoted by the double quotations, which changes the semantics from the partial match to the exact match. Is this intended?

According to this [Q&A]([os detection - How to detect the OS from a Bash script? - Stack Overflow](https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script)), `OSTYPE` in macOS can be something link `darwin9.0` so we need to accept `darwin*` at least.

themes/rjorgenson/rjorgenson.theme.sh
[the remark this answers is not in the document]
JamesMConroy: The variable was also used to color chars like `┌` so I couldn't do exactly that, but the var will use an environment var if present.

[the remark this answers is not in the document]
akinomyoga: Thank you for updating! Now the code looks perfect.

themes/rjorgenson/rjorgenson.theme.sh
JamesMConroy: Woops I missed the `*` when I was looking at this before.

akinomyoga:
Thanks! Now I think we can merge this PR.

themes/rjorgenson/rjorgenson.theme.sh
```diff
@@ -3,99 +3,103 @@
...
+is_integer() { # helper function for todo-txt-count
+    [ "$1" -eq "$1" ] > /dev/null 2>&1
+        return $?
 }
 
-function is_integer() { # helper function for todo-txt-count
+is_integer() { # helper function to make sure input is an integer
     [ "$1" -eq "$1" ] > /dev/null 2>&1
         return $?
 }
```
akinomyoga, line 55: Ah, we now actually have double definitions of `is_integer`. I'll remove it at my side.

akinomyoga:
I have merged it. Thank you again for creating the PR and your prompt responses!

PR #305, opened 2022-02-11
Fixes user defined aliases being overridden by OMB 

### LINKED ISSUE #304: Overwrites ls alias
oh-my-bash currently overrides any aliases that has been set for `ls`, if `/usr/bin/dircolors` exists.

This happens because of [lines 61 to 70 in `$OSH/lib/bourne-shell.sh`](https://github.com/ohmybash/oh-my-bash/blob/master/lib/bourne-shell.sh#L61-L70).

An alternative for this would be to do
```bash
if [ -x /usr/bin/dircolors ]; then
  test -r ~/.dircolors && eval "$(dircolors -b ~/.dircolors)" || eval "$(dircolors -b)"
  if ! alias ls >/dev/null 2>&1; then
    alias ls='ls --color=auto'
  fi
  #if ! alias dir >/dev/null 2>&1; then
  #  alias dir='dir --color=auto'
  #fi
  #if ! alias vdir; then
  #  alias vdir='vdir --color=auto'
  #fi

  if ! alias grep >/dev/null 2>&1; then
    alias grep='grep --color=auto'
  fi
  if ! alias fgrep >/dev/null 2>&1; then
    alias fgrep='fgrep --color=auto'
  fi
  if ! alias egrep >/dev/null 2>&1; then
    alias egrep='egrep --color=auto'
  fi
fi
```
This will only set the aliases if they don't already exist.

Will submit PR in a few.

solonovamax:
Fixes #304.

Currenty, oh-my-bash currently overrides any aliases that has been set for `ls`, `grep`, `egrep`, and `fgrep`, if the file `/usr/bin/dircolors` exists.
This means if a user specifies any aliases before OMB is sourced, they will be overwritten in a manner which is non transparent to the user (It is not stated anywhere, and the user cannot explicitly disabled it).

The goal of this PR is to not overwrite the aliases if they already exist, preferring the pre-defined aliases instead.
Alternatively, a means of disabling the new OMB aliases could be introduced through an environment variable, to allow the user to not have them set.

Currently, it wraps each alias call in an `if` statement, to check if they exist prior to applying them.
But, as per suggestions from akinomyoga, a function shall be introduced to streamline this process, in addition to refactors to all the other aliases which are defined by OMB that are *not* in the aliases folder or the plugins folder.

### DIFF
--- a/lib/bourne-shell.sh
+++ b/lib/bourne-shell.sh
@@ -60,13 +60,25 @@ esac
 # enable color support of ls and also add handy aliases
 if [ -x /usr/bin/dircolors ]; then
   test -r ~/.dircolors && eval "$(dircolors -b ~/.dircolors)" || eval "$(dircolors -b)"
-  alias ls='ls --color=auto'
-  #alias dir='dir --color=auto'
-  #alias vdir='vdir --color=auto'
+  if ! alias ls >/dev/null 2>&1; then
+    alias ls='ls --color=auto'
+  fi
+  #if ! alias dir >/dev/null 2>&1; then
+  #  alias dir='dir --color=auto'
+  #fi
+  #if ! alias vdir; then
+  #  alias vdir='vdir --color=auto'
+  #fi
 
-  alias grep='grep --color=auto'
-  alias fgrep='fgrep --color=auto'
-  alias egrep='egrep --color=auto'
+  if ! alias grep >/dev/null 2>&1; then
+    alias grep='grep --color=auto'
+  fi
+  if ! alias fgrep >/dev/null 2>&1; then
+    alias fgrep='fgrep --color=auto'
+  fi
+  if ! alias egrep >/dev/null 2>&1; then
+    alias egrep='egrep --color=auto'
+  fi
 fi
 
 # colored GCC warnings and errors
--- a/lib/bourne-shell.sh
+++ b/lib/bourne-shell.sh
@@ -60,38 +60,25 @@ esac
 # enable color support of ls and also add handy aliases
 if [ -x /usr/bin/dircolors ]; then
   test -r ~/.dircolors && eval "$(dircolors -b ~/.dircolors)" || eval "$(dircolors -b)"
-  if ! alias ls >/dev/null 2>&1; then
-    alias ls='ls --color=auto'
-  fi
-  #if ! alias dir >/dev/null 2>&1; then
-  #  alias dir='dir --color=auto'
-  #fi
-  #if ! alias vdir; then
-  #  alias vdir='vdir --color=auto'
-  #fi
-
-  if ! alias grep >/dev/null 2>&1; then
-    alias grep='grep --color=auto'
-  fi
-  if ! alias fgrep >/dev/null 2>&1; then
-    alias fgrep='fgrep --color=auto'
-  fi
-  if ! alias egrep >/dev/null 2>&1; then
-    alias egrep='egrep --color=auto'
-  fi
+  _omb_util_alias ls='ls --color=auto'
+  #_omb_util_alias ls='dir --color=auto'
+  #_omb_util_alias vdir='vdir --color=auto'
+  _omb_util_alias grep='grep --color=auto'
+  _omb_util_alias fgrep='fgrep --color=auto'
+  _omb_util_alias egrep='egrep --color=auto'
 fi
 
 # colored GCC warnings and errors
 #export GCC_COLORS='error=01;31:warning=01;35:note=01;36:caret=01;32:locus=01:quote=01'
 
 # some more ls aliases
-alias ll='ls -alF'
-alias la='ls -A'
-alias l='ls -CF'
+_omb_util_alias ll='ls -alF'
+_omb_util_alias la='ls -A'
+_omb_util_alias l='ls -CF'
 
 # Add an "alert" alias for long running commands.  Use like so:
 #   sleep 10; alert
-alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'
+_omb_util_alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'
 
 # Alias definitions.
 # You may want to put all your additions into a separate file like
--- a/lib/bourne-shell.sh
+++ b/lib/bourne-shell.sh
@@ -63,18 +63,14 @@ if [ -x /usr/bin/dircolors ]; then
   _omb_util_alias ls='ls --color=auto'
   #_omb_util_alias ls='dir --color=auto'
   #_omb_util_alias vdir='vdir --color=auto'
-  _omb_util_alias grep='grep --color=auto'
-  _omb_util_alias fgrep='fgrep --color=auto'
-  _omb_util_alias egrep='egrep --color=auto'
+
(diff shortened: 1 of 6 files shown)


akinomyoga:
- I don't know if there is any such workflow of creating an issue with the description and a PR with no description at the same time, but why don't you put all the explanations in the PR itself?
- I'm not sure if the existing aliases should be preferred to the ones defined by OMB *in general*. Aren't there any cases that the aliases in OMB preferred to the existing ones? I feel the situation is usually opposite to this PR: *First, some system-default aliases are defined by `/etc/profile.d/*`. Next, each user overwrites the aliases with the custom ones in `~/.bashrc` if necessary.* If we extend this, shouldn't the custom aliases by the user be defined after sourcing `oh-my-bash.sh`? If the user first defines the aliases and then sources `oh-my-bash.sh`, OMB cannot distinguish whether the existing aliases are the system-default one (that can be overwritten) or the user-defined ones (that shouldn't be overwritten). What is your use case?
- There seem to be still other places that overwrite these names with aliases: [lib/directories.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/directories.sh#L27-L30), [lib/grep.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/grep.sh#L26). If we should apply the change of this PR, don't you also need to check these aliases?
- If we should apply the change of this PR, why do you check only a part of the aliases defined in `lib/*.sh`? There are also similar aliases even in the same file ([`bourne-shell.sh`#L76-L78](https://github.com/ohmybash/oh-my-bash/blob/master/lib/bourne-shell.sh#L76-L78)). There are many others in other files. What are the differences between the ones checked for the existing definition and the ones unchecked? If there is no difference, maybe we should add the same checks to all the aliases.
- Maybe we may define a new function for this purpose.

solonovamax:
Updated PR description to contain more information

solonovamax:
Further, should the aliases in the [grep.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/grep.sh#L26), [directories.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/directories.sh#L27-L30), and [bourne-shell.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/bourne-shell.sh#L60-L82) not be moved into the [aliases folder](https://github.com/ohmybash/oh-my-bash/tree/master/aliases), so that the user can explicitly enable/disable them?

That seems much more reasonable to me, as I don't believe these aliases are actually used anywhere in OMB components.

akinomyoga:
Thank you for updating the description. Additionally, you may edit the PR title.

> Because, if it was done in the `ls.aliases.sh` file, I would find this much more reasonable, as the user explicitly enabled the ls aliases. But, with the current implementation, it is overwritten in a core component of OMB, which means it cannot be disabled by the user if they desire.

It's a very fair argument.

> Further, should the aliases in the [grep.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/grep.sh#L26), [directories.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/directories.sh#L27-L30), and [bourne-shell.sh](https://github.com/ohmybash/oh-my-bash/blob/master/lib/bourne-shell.sh#L60-L82) not be moved into the [aliases folder](https://github.com/ohmybash/oh-my-bash/tree/master/aliases), so that the user can explicitly enable/disable them?

Yeah, I agree. I also think we should clean up these things. One thing that I'm not sure of is whether we can introduce such breaking changes without notice. I guess there are a few hundred thousand users of OMB (at least, there were about twenty thousand clones by ten thousand unique cloners in the last two weeks), and I guess most of them are using OMB with the default configuration plus a theme. I doubt most of them actually check what aliases are defined in each `*.aliases.sh` and can recover the aliases moved to e.g. `ls.aliases.sh` by turning on `aliases+=(ls)` when they have been liked it. But probably, we don't have to care about the users of the default configuration too much.  If you could clean them up, it would be very helpful.

solonovamax:
Yes, the breaking change is my one concern.
One possible idea would be to instead introduce them via an environment variable instead?

solonovamax:
Also, I am converting the PR to a draft until we decide which option to go with.

akinomyoga:
Thank you for the reply. It's actually hard for me to judge whether we can break the existing definitions.

> One possible idea would be to instead introduce them via an environment variable instead?

Can you describe it in more detail? Is it the same environment variable as the one mentioned in the cover (as quoted below)?

> Alternatively, a means of disabling the new OMB aliases could be introduced through an environment variable, to allow the user to not have them set.

My current consideration is this: For example, if that variable contains the value `no`, new aliases are not defined by `lib/*.sh` at all. If the value is `yes`, the new aliases in `lib/*.sh` always override existing ones. If the value is `check`, the new aliases are defined only when there is no existing definition. The "default behavior" when the variable is not defined is `yes` so that the behavior in the existing environment does not change by upgrading. We put `OMB_XXXX=check` in `templates/bashrc.osh-template` so that the existing aliases are checked in the new environments. After a long time, we may switch the default behavior from `yes` to `check`. But maybe we can just forcibly change the default behavior from `yes` to `check` from the beginning, which is maybe cleaner.

Or maybe you are thinking of introducing a variable to control the aliases defined in `directories.sh` and `ls.aliases.sh` in a different way?

solonovamax:
> Thank you for the reply. It's actually hard for me to judge whether we can break the existing behavior.
> 
> > One possible idea would be to instead introduce them via an environment variable instead?
> 
> Can you describe it in more detail? Is it the same environment variable as the one mentioned in the cover (as quoted below)?
> 
> > Alternatively, a means of disabling the new OMB aliases could be introduced through an environment variable, to allow the user to not have them set.

Yes, that is what I meant.

An environment variable like `DISABLE_AUTO_UPDATE`, `COMPLETION_WAITING_DOTS`, `OSH_THEME`, etc. could be introduced, with the default set to `true`.


> My current consideration is this: For example, if that variable contains the value `no`, new aliases are not defined by `lib/*.sh` at all. If the value is `yes`, the new aliases in `lib/*.sh` always override existing ones. If the value is `check`, the new aliases are defined only when there is no existing definition. The "default behavior" when the variable is not defined is `yes` so that the behavior in the existing environment does not change by upgrading. We put `OMB_XXXX=check` in `templates/bashrc.osh-template` so that the existing aliases are checked in the new environments. After a long time, we may switch the default behavior from `yes` to `check`. But maybe we can just forcibly change the default behavior from `yes` to `check` from the beginning, which is maybe cleaner.
> 
> Or maybe you are thinking of introducing a variable to control the aliases defined in `directories.sh` and `ls.aliases.sh` in a different way?

That is what I was thinking, although it would not control any aliases in `ls.aliases.sh`, only those in `lib/*.sh`.

Also, `enabled`, `disabled`, and `check` might be better.

akinomyoga:
I thought, ideally, it'd be better if we could finally reorganize things in `ls.aliases.sh` and other `*.aliases.sh`, but the suggested shell variable (to control the aliases in `lib/*.sh`) may be the practical solution. You can decide which option to take.

> Also, `enable`, `disable`, and `check` might be better.

Yes, though maybe it depends on the choice of the variable name. We have been defining variable names like `DISABLE_AUTO_UPDATE` and `COMPLETION_WAITING_DOTS`, but now we want to prefix all the OMB variables with `OMB_`. (If you are interested the related discussion is #280 and there is a pending fix at PR #270)

akinomyoga:
Can I take over this PR and implement the suggested configuration variable (say, `OMB_DEFAULT_ALIASES={enable,disable,check}`)?

solonovamax:
Of course!

I don't mind at all if you take over this & PR it, as I haven't been working on it at all in quite a while

akinomyoga:
Thank you for your response! I'll later work on it when I have time!

akinomyoga:
I've implemented `OMB_DEFAULT_ALIASES={enable,disable,check}` and pushed it to the PR branch.

akinomyoga:
@solonovamax I've merged the PR. Thank you for raising the issue and attempting to solve the issue!

PR #310, opened 2022-02-22
Add asdf plugin support

prodrigues1912:


plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,27 @@
...
+ASDF_COMPLETIONS="$ASDF_DIR/completions"
+
+# If not found, check for archlinux/AUR package (/opt/asdf-vm/)
+if [[ ! -f "$ASDF_DIR/asdf.sh" || ! -f "$ASDF_COMPLETIONS/asdf.bash" ]] && [[ -f "/opt/asdf-vm/asdf.sh" ]]; then
+  ASDF_DIR="/opt/asdf-vm"
+  ASDF_COMPLETIONS="$ASDF_DIR"
+fi
+
+# If not found, check for Homebrew package
+if [[ ! -f "$ASDF_DIR/asdf.sh" || ! -f "$ASDF_COMPLETIONS/asdf.bash" ]] && (( $+commands[brew] )); then
```
akinomyoga, line 12: ` $+commands[brew]` is Zsh syntax. Could you please update it for Bash?

akinomyoga (changes requested):
Thank you for the PR. Do you have the copyright of these codes? This seems like just copied from oh-my-zsh and the license of oh-my-zsh is the MIT license. We actually already included the copyright notice of oh-my-zsh in our LICENCE, so there is no problem in merging this PR, but, nevertheless. you actually should mention it if you are not the copyright holder of these codes in general. Thank you!

plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Load command
+if [[ -f "$ASDF_DIR/asdf.sh" ]]; then
+  . "$ASDF_DIR/asdf.sh"
+
+  # Load completions
+  if [[ -f "$ASDF_COMPLETIONS/asdf.bash" ]]; then
+    . "$ASDF_COMPLETIONS/asdf.bash"
+  fi
+fi
```
akinomyoga, line 27: We want a newline at the end of the file.

PR #319, opened 2022-03-17
Added the possibility to have an own .bashrc.local

mnlwldr:


akinomyoga:
Thank you for creating a PR! I guess the local settings have been supposed to be placed in `.oh-my-bash/custom/lib/*.sh` in oh-my-bash, but I also agree that it would be useful to automatically load `~/.bashrc.local`!

Actually, there is a similar code in [`lib/bourne-shell.sh`](https://github.com/ohmybash/oh-my-bash/blob/master/lib/bourne-shell.sh#L89-L91) for `~/.bash_aliases`. [ Note: all of `lib/*.sh` are automatically loaded by oh-my-bash. ] I think it is better to put these codes in the same place. What do you think?

mnlwldr:
Hi :) Absolutly. I moved that from `templates/bashrc.osh-template` to `lib/bourne-shell.sh`

akinomyoga:
Thank you! I have merged it!

PR #320, opened 2022-03-22
Add --autostash to the git pull command

yelkarama:
Add --autostash to the git pull command to allow upgrading when some files were changed locally.


akinomyoga:
Thank you! If we specify `--autostash` here, the changes made by users would be lost without notice. The local changes are supposed to be stashed, reverted, or committed at the user's discretion. I think not all users would like to stash the local changes. At least, this option `--autostash` should be specified only when the user explicitly requests it.

akinomyoga:
This can be useful for some users who want to make temporary changes to the files that can be reverted anytime later, but I think it is dangerous for most users to make it default. So, I'm thinking of not including this change in the upstream OMB. What do you think?

PR #331, opened 2022-05-05
vscode theme

leofdss:
Theme based on the "dev container" generated by the "Remote - Containers" plugin for Visual Studio code.

Reference: [vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers)


akinomyoga:
Thank you for your contribution! There is one thing I want to confirm before merging.

> Theme based on the "dev container" generated by the "Remote - Containers" plugin for Visual Studio code.

Does this mean the code in this PR is generated by `vscode-dev-container` or have you implemented it from scratch by mimicking the appearance?

If it is the former, I would like to check the license of the codes generated by `vscode-dev-container`. If the license of `vscode-dev-container` also applies to the generated code (for example, for the reason that the generated strings are actually originally embedded in the source code of `vscode-dev-container`), we need to retain the original copyright notice by [Microsoft Coorporation](https://github.com/microsoft/vscode-dev-containers/blob/172a918f40e31bd24da8e64135026ec9f26c91b0/LICENSE#L3), in which case you can simply include the content of [vscode-dev-containers/LICENSE](https://github.com/microsoft/vscode-dev-containers/blob/172a918f40e31bd24da8e64135026ec9f26c91b0/LICENSE#L3) (particularly the copyright notice of Microsoft) in the theme file `vscode.theme.sh`.

leofdss:
The code was generated by [vscode-dev-containers](https://github.com/microsoft/vscode-dev-containers).

[Original string](https://github.com/microsoft/vscode-dev-containers/blob/172a918f40e31bd24da8e64135026ec9f26c91b0/containers/javascript-node/.devcontainer/library-scripts/common-debian.sh#L303-L320)

[Reference to license](https://github.com/microsoft/vscode-dev-containers/blob/172a918f40e31bd24da8e64135026ec9f26c91b0/containers/javascript-node/.devcontainer/library-scripts/common-debian.sh#L3-L4)

I added one more commit with the existing license in the original code.

akinomyoga:
Thank you! I'll later adjust the styles and merge it!

akinomyoga:
I have added adjustments. Could you check them?

themes/vscode/vscode.theme.sh
```diff
@@ -26,4 +26,4 @@ _omb_theme_PROMPT_COMMAND() {
     unset -f _omb_theme_PROMPT_COMMAND
 }
 _omb_theme_PROMPT_COMMAND
-export PROMPT_DIRTRIM=4
\ No newline at end of file
+export PROMPT_DIRTRIM=4
```
akinomyoga: We want to add a newline at the end of the file to make it better work with POSIX utilities.

themes/vscode/vscode.theme.sh
```diff
@@ -5,7 +5,7 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-_omb_theme_PROMPT_COMMAND() {
+_omb_theme_vscode_initialize() {
```
akinomyoga, line 8: The function name `_omb_theme_PROMPT_COMMAND` is used for the function that is supposed to be called after every execution of a user command. This function seems to be only executed once in the initialization time and soon delete by `unset -f`, so the name does not match with the function purpose. Let me change the function name.

themes/vscode/vscode.theme.sh
```diff
@@ -26,4 +26,4 @@ _omb_theme_vscode_initialize() {
     unset -f _omb_theme_vscode_initialize
 }
 _omb_theme_vscode_initialize
-export PROMPT_DIRTRIM=4
+PROMPT_DIRTRIM=4
```
akinomyoga, line 31: The export attribute is not needed because the variable `PROMPT_DIRTRIM` is basically a shell variable that is used to configure the current process of Bash but is not an environmental variable for external commands.

themes/vscode/vscode.theme.sh
```diff
@@ -26,4 +26,6 @@ _omb_theme_vscode_initialize() {
     unset -f _omb_theme_vscode_initialize
 }
 _omb_theme_vscode_initialize
+
+_omb_theme_PROMPT_COMMAND() { true; }
```
akinomyoga, line 30: This change is related to the internals of OMB, but we need to clear `_omb_theme_PROMPT_COMMAND` set by another theme in case that the theme has been switched from another to `vscode` in a single session.

leofdss:
I agree with all changes

akinomyoga:
Thank you for checking! I'll rebase, squash, and merge it! Thank you again for opening the PR and for your quick response!

PR #332, opened 2022-05-18
Fix Tmux-autoattach plugin to reattach to a session is there is one

ThePhilderbeast:
this corrects the plugin not connecting back to the session by default

### DIFF
--- a/plugins/tmux-autoattach/tmux-autoattach.plugin.sh
+++ b/plugins/tmux-autoattach/tmux-autoattach.plugin.sh
@@ -7,6 +7,14 @@
 
 [ -z "$OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR" ] && OSH_PLUGIN_TMUX_AUTOATTACH_BEHAVIOR="detach"
 
+# Note on the option "-As0": tmux-3.0a and before does not attach to a session
+# as far as the session name is not given by "-s SESSION_NAME" [1].  From
+# tmux-3.1, it attaches to an existing session, if any, without specifying the
+# session.  Here, we assume the session name to be "0", which is the default
+# name for the first session.
+#
+# [1] https://github.com/ohmybash/oh-my-bash/pull/332
+
 _osh_plugin_tmux_autoattach_exit() {
 	[ -z "$TMUX" ] && tmux -2u new -As0 && exit
 }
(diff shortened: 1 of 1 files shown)


akinomyoga:
Thank you for the PR and sorry for the delay. Could you describe more context on this PR? Maybe I miss the point as I don't usually use `tmux`, but

1. It seems to me that specifying a session is only needed when there are multiple sessions, but if we are forced by bashrc to use only one specific session `0`, creating multiple sessions seems to be useless. What would be the use case for restricting to only a single session while creating multiple sessions?
2. As far as I try `tmux -2u new -A`, it already seems to anyway attach to a session even when there is more than one session.
3. When there is only a single session `1` (after the session `0` is closed), `tmux -2u new -As0` fails to attach to the session `1` and creates another session.

ThePhilderbeast:
In my testing it was never reconnecting back to the existing session, specifying session 0 meant that it would come back to the original session as desired, this could be a version issue as you suggest, I am using 3.0 on ubuntu 20.04, generally if we are auto attaching I would suggest that we do not want to create multiple sessions, but rather want to reuse the same one when we open a new shell.

> When there is only a single session 1 (after the session 0 is closed), tmux -2u new -As0 fails to attach to the session 1 and creates another session.
this would only be the case if they are creating a session out side of this plugin, which would generally mean they are creating a session in a session, something that is unlikely enough that I believe its worth the trade off, particularly over ending up with a number of sessions that never get reattached to if you have it set to exit.

akinomyoga:
Thank you for your reply!

> it was never reconnecting back to the existing session, specifying session 0 meant that it would come back to the original session as desired, this could be a version issue as you suggest, I am using 3.0 on ubuntu 20.04,

Thank you for the information! I'll later check the behavior of `tmux-3.0`.

> if we are auto attaching I would suggest that we do not want to create multiple sessions, but rather want to reuse the same one when we open a new shell.

Oh, does it mean `tmux` in your environment creates a new session with `tmux -2u new -A` even when there is only one session?

akinomyoga:
Thank you! I've confirmed the behavior. In `tmux-3.0a` and before, `tmux -2u new -A` doesn't seem to attach to an existing session even when it is unique unless the session name is specified through the `-s` option. I'll merge it after I add some comments. Thank you!

akinomyoga:
I have merged it. Thank you again for your contribution!

PR #335, opened 2022-06-15
themes/sirup: Show VirtualEnv

juanswan13:
Added the ability to show the current Virtual Env when using Python Virtual Env. 
If there's not vitual env, the theme works normal.

### DIFF
--- a/themes/sirup/sirup.theme.sh
+++ b/themes/sirup/sirup.theme.sh
@@ -1,8 +1,8 @@
 #! bash oh-my-bash.module
 # For unstaged(*) and staged(+) values next to branch name in __git_ps1
 GIT_PS1_SHOWDIRTYSTATE="enabled"
-OMB_PROMPT_VIRTUALENV_FORMAT='[%s] '
-OMB_PROMPT_CONDAENV_FORMAT='[%s] '
+OMB_PROMPT_VIRTUALENV_FORMAT=' [%s]'
+OMB_PROMPT_CONDAENV_FORMAT=' [%s]'
 OMB_PROMPT_SHOW_PYTHON_VENV=${OMB_PROMPT_SHOW_PYTHON_VENV:=false}
 
 function _omb_theme_sirup_rubygem {
@@ -20,7 +20,7 @@ function _omb_theme_PROMPT_COMMAND {
   # Check http://github.com/Sirupsen/dotfiles for screenshot
   local python_venv
   _omb_prompt_get_python_venv
-  PS1="$_omb_prompt_purple$python_venv$_omb_prompt_navy\W/$_omb_prompt_bold_navy$(_omb_theme_sirup_rubygem)$_omb_prompt_bold_green$(__git_ps1 " (%s)") ${_omb_prompt_normal}$ "
+  PS1="$_omb_prompt_navy\W/$_omb_prompt_bold_navy$(_omb_theme_sirup_rubygem)$_omb_prompt_bold_purple$python_venv$_omb_prompt_bold_green$(__git_ps1 " (%s)") ${_omb_prompt_normal}$ "
 }
 
 _omb_util_add_prompt_command _omb_theme_PROMPT_COMMAND
--- a/themes/sirup/sirup.theme.sh
+++ b/themes/sirup/sirup.theme.sh
@@ -3,6 +3,7 @@
 GIT_PS1_SHOWDIRTYSTATE="enabled"
 OMB_PROMPT_VIRTUALENV_FORMAT='[%s] '
 OMB_PROMPT_CONDAENV_FORMAT='[%s] '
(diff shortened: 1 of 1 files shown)


themes/sirup/sirup.theme.sh
```diff
@@ -15,7 +25,7 @@ function _omb_theme_sirup_rubygem {
  
 function _omb_theme_PROMPT_COMMAND {
   # Check http://github.com/Sirupsen/dotfiles for screenshot
-  PS1="$_omb_prompt_navy\W/$_omb_prompt_bold_navy$(_omb_theme_sirup_rubygem)$_omb_prompt_bold_green$(__git_ps1 " (%s)") ${_omb_prompt_normal}$ "
+  PS1="$_omb_prompt_purple$(omb_theme_syrup_python_venv)$_omb_prompt_navy\W/$_omb_prompt_bold_navy$(_omb_theme_sirup_rubygem)$_omb_prompt_bold_green$(__git_ps1 " (%s)") ${_omb_prompt_normal}$ "
```
akinomyoga, line 28: There seems to be already a better implementation `_omb_prompt_get_python_venv` that also checks `CONDA_DEFAULT_ENV` in addition to `VIRTUAL_ENV`.

https://github.com/ohmybash/oh-my-bash/blob/cb7238fb8f18a2df22b3dfb6f14e35229b5fc56c/lib/omb-prompt-base.sh#L468-L474

`omb-prompt-base.sh` is loaded by default, so the function `_omb_prompt_get_python_venv` is readily available in the theme files:

```suggestion
  local python_venv; _omb_prompt_get_python_venv
  PS1="$_omb_prompt_purple$python_venv$_omb_prompt_navy\W/$_omb_prompt_bold_navy$(_omb_theme_sirup_rubygem)$_omb_prompt_bold_green$(__git_ps1 " (%s)") ${_omb_prompt_normal}$ "
```

akinomyoga (commented):

Actually, this type of adjustment of themes depends on personal preferences, so I wouldn't merge it soon unless there is a strong reason that we need to merge it. Rather, we would recommend users customize it personally following [the corresponding section in README](https://github.com/ohmybash/oh-my-bash#customization-of--plugins-and-themes).

Nevertheless, this PR can be used to demonstrate how to customize the theme, so let me add a label to this PR.

themes/sirup/sirup.theme.sh
shelfofclub: Some themes like "font" use `_omb_prompt_get_python_venv`. Some themes like "powerline" use `__powerline_python_venv_prompt` defined by itself.
`_omb_prompt_get_python_venv` is only used by garo, font and hawaii50. It's needed to unify all themes.

akinomyoga: As for the duplication of `__powerline_python_venv_prompt` vs `_omb_prompt_get_python_venv`, I currently think I wouldn't unify it.

- The `powerline` theme in this repository, unlikely the other ones, has its own independent and sophisticated theming system. It's hard to unify the two completely different theming systems (omb-prompt vs powerline) into one. I'm also not sure if there is a motivation to invent a heterogeneous amalgam of two different theming systems.
- I'm not sure which is the exact origin of the theme `powerline` in this repository, but there is likely an upstream somewhere. I wouldn't like to introduce structural changes on the powerline theme independently from the upstream if any. At least, I'd like to be careful before deciding on the large edits on the `powerline` theme.

What I'm thinking for the refactoring is to unify the access to `VIRTUAL_ENV` and other environment variables within a few interfaces, `_omb_prompt_get_python_venv` or similar ones. There are still themes that directly reference `VIRTUAL_ENV` such as `agnoster`, `duru`, `mairan`, `slick` and `zork`, though I haven't decided what to do with each one because I haven't checked the insides of them.

shelfofclub: Yeah, you are right.

akinomyoga: ping @juanswan13 Can I apply to the PR branch the change I suggested? If you think there is a reason that we should keep the current version of PR, please let us know!

akinomyoga: @juanswan13 There seems to be no reply, so I decided to push the change.

akinomyoga:
I changed my mind and decided to merge this into `master`. The background is explained [here](https://github.com/ohmybash/oh-my-bash/pull/393#issuecomment-1416783536).

- I added a commit 3cee7307bcd2358bdeb007e52baa06b416189fb8 and rebased.
- It seems there is already a segment of context information about `ruby_gem`, so it is more reasonable to put the `python_venv` information next to the ruby_gem information. I have moved the position of `$python_venv` to honor the original theme in commit 71d9a1b7dac65c7bcda07f0befb0d8b04ac472d4. I also made the style `bold` to make it match with `ruby_gem`.

If there are no further problems or suggestions, I'll wait for a few days and merge this. Thank you!

akinomyoga (approved):

akinomyoga:
Thanks, I have merged it.

PR #344, opened 2022-07-07
Vagrant plugin creation for OMB

oscarenzo:
Create a new plugin to work with vagrant environments, its easy to use.

### DIFF
--- a/plugins/vagrant/vagrant.plugin.sh
+++ b/plugins/vagrant/vagrant.plugin.sh
@@ -1,21 +1,23 @@
 #! bash oh-my-bash.module
+
 # Functions
-function vagrant-version(){
+function vagrant-version() {
     vagrant --version
 }
 
-function vagrant-init(){
-    if [ -n "$1" ] ; then
+function vagrant-init() {
+    if [[ $1 ]]; then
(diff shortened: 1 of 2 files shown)


plugins/vagrant/vagrant.plugin.sh
```diff
@@ -0,0 +1,70 @@
...
+alias va='vagrant'
+alias vaver='vagrant-version'
+alias vaconf='vagrant ssh-config'
+alias vastat='vagrant-status'
+alias vacheck='vagrant validate'
+alias vaport='vagrant port'
+alias vapvm='vagrant-plugin-vm'
+alias vapi='vagrant plugin install'
+alias vapr='vagrant plugin uninstall'
+alias vau='vagrant-up '
```
akinomyoga, line 63: Is this space after `vagrant-up` intentional? In Bash, aliases that end with space have a special meaning that the first argument should also become subject to the alias expansions. However, as far as I search, `vagrant-up` doesn't seem to receive commands as arguments or some other types of arguments that can be usefully expanded by aliases.

oscarenzo:
thats ok, thank you!

akinomyoga:
Thanks! I'll merge it!

PR #345, opened 2022-07-13
Feature to connect over vagrant servers

oscarenzo:
Feature to connect over vagrant servers
Fix in Ansible plugin description

### DIFF
--- a/plugins/vagrant/vagrant.plugin.sh
+++ b/plugins/vagrant/vagrant.plugin.sh
@@ -1,4 +1,5 @@
 #! bash oh-my-bash.module
+# Author: Enzo Arroyo <enzo@arroyof.com>
 
 # Functions
 function vagrant-version() {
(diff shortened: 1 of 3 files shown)


akinomyoga:
Thanks. I have merged it.

plugins/vagrant/vagrant.plugin.sh
```diff
@@ -55,22 +55,24 @@ function vagrant-status() {
...
         vagrant ssh
-    elif [[ $1 ]] && [ "$VMCOUNT" -gt 1 ]; then
+    elif [[ $1 ]] && ((VMCOUNT > 1)); then
         echo "Login to : $1 Server...."
         vagrant ssh "$1"
-    elif [ "$VMCOUNT" == 0 ]; then
+    elif ((VMCOUNT == 0)); then
         echo "Seems like that not there running servers" >&2
-        return 2
+        return 1
```
akinomyoga, line 73: Note: `return 2` is used for the wrong usage of the command. The runtime error can be represented by `return 1`.

plugins/vagrant/vagrant.plugin.sh
```diff
@@ -55,22 +55,24 @@ function vagrant-status() {
 
 function vagrant-ssh() {
     local VMCOUNT
-    VMCOUNT="$(vagrant status | grep -c running)"
+    VMCOUNT=$(vagrant status | grep -c running)
     local VMDEFAULT
-    VMDEFAULT="$(vagrant status | grep -w default | grep -c running)"
+    VMDEFAULT=$(vagrant status | grep -w default | grep -c running)
```
akinomyoga, line 60: Note: We actually do not need to quote them in the right-hand side of the variable assignments (unless we want to include literal shell special characters) because the pathname expansions and word splitting does not occur. An exception is a case that the right-hand side contains `$*` or `${arr[*]}` with which some older versions of Bash have bugs.

oscarenzo: > Note: We actually do not need to quote them in the right-hand side of the variable assignments (unless we want to include literal shell special characters) because the pathname expansions and word splitting does not occur. An exception is a case that the right-hand side contains `$*` or `${arr[*]}` with which some older versions of Bash have bugs.

Okay, thank you by the advice.

akinomyoga: Yeah, this is just a matter of style after all. You can use either style in your scripts. BTW, `shellcheck` (of at least recent versions) doesn't warn about this. Anyway, thank you for your contribution!

PR #352, opened 2022-08-28
Add theme: rr inspired by robbyrussell from ohmyzsh themes

anish-mandal:
The PR adds a new simple theme `rr` inspired from [robbyrussell](https://github.com/ohmyzsh/ohmyzsh/wiki/Themes#robbyrussell) from ohmyzsh themes.

Take a look:

Its just simple and cool I think.


themes/rr/rr.theme.sh
```diff
@@ -0,0 +1,29 @@
...
+
+function _omb_theme_RR_PROMPT() {
+    # configs
+    user_name="${_omb_prompt_white}\u${_omb_prompt_reset_color}"
+    base_directory="${_omb_prompt_bold_blue}\W${_omb_prompt_reset_color}"
+    SCM_THEME_PROMPT_PREFIX="${_omb_prompt_bold_purple}git:(${_omb_prompt_reset_color}"
+    SCM_THEME_PROMPT_SUFFIX="${_omb_prompt_bold_purple})${_omb_prompt_reset_color}"
+    SCM_THEME_PROMPT_CLEAN="${_omb_prompt_bold_green} ✓${_omb_prompt_reset_color}"
+    SCM_THEME_PROMPT_DIRTY="${_omb_prompt_bold_red} ✗${_omb_prompt_reset_color}"
+    arrow="${_omb_prompt_bold_purple}➜${_omb_prompt_reset_color}"
```
akinomyoga, line 20: We want to make the variables used only in the present function call local.

```suggestion
    local user_name="${_omb_prompt_white}\u${_omb_prompt_reset_color}"
    local base_directory="${_omb_prompt_bold_blue}\W${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_PREFIX="${_omb_prompt_bold_purple}git:(${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_SUFFIX="${_omb_prompt_bold_purple})${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_CLEAN="${_omb_prompt_bold_green} ✓${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_DIRTY="${_omb_prompt_bold_red} ✗${_omb_prompt_reset_color}"
    local arrow="${_omb_prompt_bold_purple}➜${_omb_prompt_reset_color}"
```

themes/rr/rr.theme.sh
```diff
@@ -0,0 +1,29 @@
...
+
+    PS1="${arrow}  ${user_name} ${base_directory} "
+
+    if [[ "$(scm_prompt_info)" ]]; then
+        PS1="${PS1}$(scm_prompt_info) "
```
akinomyoga, line 25:
```suggestion
    local prompt_info=$(scm_prompt_info)
    PS1+=${prompt_info:+$prompt_info }
```

themes/rr/rr.theme.sh
```diff
@@ -0,0 +1,29 @@
...
+    fi
+}
+
+_omb_util_add_prompt_command _omb_theme_RR_PROMPT
```
akinomyoga, line 29: We are using `_omb_theme_PROMPT_COMMAND` for the prompt setting of all the themes. This is because, when we switch the theme in the same session, we want to overwrite the function for `PROMPT_COMMAND` set by the previous theme. Please also update the name in the function definition.

```suggestion
_omb_util_add_prompt_command _omb_theme_PROMPT_COMMAND
```

akinomyoga (changes requested):
Thank you for your contribution! Could you check these points?

anish-mandal:
I committed all the requested changes. Now you can merge.

themes/rr/rr.theme.sh
```diff
@@ -0,0 +1,28 @@
...
+    local arrow="${_omb_prompt_bold_purple}➜${_omb_prompt_reset_color}"
+
+    PS1="${arrow}  ${user_name} ${base_directory} "
+
+    local scm_info=$(scm_prompt_info)
+
+    PS1+=${scm_info:+$scm_info }
+}
+
+_omb_util_add_prompt_command _omb_theme_PROMPT_COMMAND
```
akinomyoga, line 28: Thanks for updating!

An additional point is that we need a newline character at the end of the file. Missing the newline character may cause problems with other command line tools that handle this file since the line needs to be always terminated by a newline according to the POSIX standard. I think this doesn't need to be discussed, so I'll just fix it.

akinomyoga (approved):

akinomyoga:
Thank you! I have merged it!

PR #355, opened 2022-09-10
Add robbyrussell theme

gjurgensen:
This theme attempts to recreate the default oh-my-zsh theme of the same name. It is a modification of the existing "rr" theme.


themes/robbyrussell/robbyrussell.theme.sh
```diff
@@ -0,0 +1,25 @@
+#! bash oh-my-bash.module
```
akinomyoga, line 1: Could you mention in a comment that the design is taken from robbyrussell of ohmyzsh?

akinomyoga (changes requested):
Thank you for your contribution. So, this is a more faithful reproduction of the appearance of the original robbyrussell theme in ohmyzsh? I think I'm going to merge the PR.

- Could you also add an entry to [`themes/THEMES.md`](https://github.com/ohmybash/oh-my-bash/blob/master/themes/THEMES.md)?
- Also, I would like to normalize the sizes of new images. Could you make the image smaller? The typical size is about 700x290.

gjurgensen:
Yes, the "rr" theme was evidently inspired by the robbyrussell theme, whereas this attempts to recreate it as closely as possible. In particular, that means dropping the user name, matching the colors, color-coding the prompt arrow based on `$?`, and moving the "dirty prompt" signifier to the very end.

I believe I incorporated all of your requested changes. Let me know if I should change anything else.

akinomyoga (approved):

akinomyoga:
Thank you! I'll squash the commit of image change and merge it. This is because one of the reasons to normalize the size is to keep the repository data size small.

akinomyoga (approved):

akinomyoga (dismissed):

akinomyoga:
I have also adjusted the commit messages. I'll later merge it. Thank you!

gjurgensen:
I had to make a small change, adding `${_omb_prompt_normal}` to the end of the prompt. I realized that without this, the terminal output was being made bold when outside a git/svn repository.

themes/robbyrussell/robbyrussell.theme.sh
```diff
@@ -0,0 +1,29 @@
...
+
+function _omb_theme_PROMPT_COMMAND() {
+    if [[ "$?" == 0 ]]; then
+        local arrow_color="${_omb_prompt_bold_green}"
+    else
+        local arrow_color="${_omb_prompt_bold_brown}"
+    fi
+
+    local base_directory="${_omb_prompt_bold_teal}\W${_omb_prompt_reset_color}"
+    local SCM_THEME_PROMPT_PREFIX="${_omb_prompt_bold_navy}git:(${_omb_prompt_bold_brown}"
```
akinomyoga, line 14: > Also, I realized that when inside an svn repository, the repo information is still decorated with "git:(...)". If there is a way to check whether we are in a git vs svn repository, I could change this to say "svn" when appropriate.

Can you test this?

```suggestion
    local GIT_THEME_PROMPT_PREFIX="${_omb_prompt_bold_navy}git:(${_omb_prompt_bold_brown}"
    local SVN_THEME_PROMPT_PREFIX="${_omb_prompt_bold_navy}svn:(${_omb_prompt_bold_brown}"
    local HG_THEME_PROMPT_PREFIX="${_omb_prompt_bold_navy}hg:(${_omb_prompt_bold_brown}"
```

akinomyoga (commented):
> I had to make a small change, adding `${_omb_prompt_normal}` to the end of the prompt. I realized that without this, the terminal output was being made bold when outside a git/svn repository.

Thanks, good point. I have amended the commit to apply the same change to `rr` (748b584).

themes/robbyrussell/robbyrussell.theme.sh
gjurgensen: Yes, that produces the desirable behavior, thanks!

akinomyoga: Thanks for checking!

akinomyoga: I've also applied the same change to the `rr` theme and squashed it into your new commit. I have also edited the commit message.

akinomyoga (approved):

gjurgensen:
Hi, may I ask the status of this PR? Will it be merged soon now that is has been approved?

akinomyoga:
I'm keeping this open for at least a week in case the owner @nntoan would like to merge or have comments.

The repository setting has been recently changed so that at least one approval is needed before the merging. Because I'm not the owner of this repository and the owner is not responsive these years, I'm not sure if this is an intended change of the setting by the owner and what is the background of the change of the setting.

@nntoan May I ask whether you have intentionally changed the setting to request an approval for merging? If so, could I have an explanation about the background of the change of the setting and what I should do for that? I have been thinking that the setting to request an approval is used to double-check the PR by at least two people, i.e., a reviewer to approve and a maintainer to merge. However, currently there is only one active member, so the things don't work in that way. I can still self-approve the PR and merge it soon, but that seems like a meaningless procedure, and what we are doing is effectively the same as before the setting change. I guessed you would like to change something, but I'm not sure currently what it is. The setting change is now just blocking merging PRs.

nntoan:
@akinomyoga I only enable these changes to wipe out GitHub recommendations on every PRs, I know it is pretty meaningless at this time as you are the most active maintainer here. However, in the future when we have other people joining the team, it will work similar with what you mention above, member review PRs and approve then the maintainer will do the merging... I also want to integrate some CI for shellcheck into the project, that's one of reasons I turned it on.

Let me know what's your thoughts about this. I can turning these off if it's blocking you guys.

akinomyoga:
Hi @nntoan, Thank you for your reply! OK! I now see the background of the setting change. I was just afraid that it might not be what you expect that I approve and merge PRs. So, I hesitated to merge PRs after the setting change. Considering the current situation of only one active member, I think I'll continue to approve and merge PRs until some other people join us as you say. Thank you again for your reply!

akinomyoga:
> I also want to integrate some CI for shellcheck into the project, that's one of reasons I turned it on.

There is some stub of CI in #191. I think we may add the `shellcheck` stuff there.

> Let me know what's your thoughts about this. I can turning these off if it's blocking you guys.

Now I see the backgrounds, so I can handle things in accordance with it. Then, it is fine to keep the current setting as is.

PR #356, opened 2022-09-13
Changes and betters on script to manager the vagrant environment usin…

oscarenzo:
Changes and betters on script to manager the vagrant environment using OMB vagrant plugin

### DIFF
--- a/plugins/vagrant/README.md
+++ b/plugins/vagrant/README.md
@@ -20,7 +20,7 @@ plugins=(... vagrant)
 | `vastat`           | command `vagrant global-status`                              |
 | `vacheck`          | command `vagrant validate`                                   |
 | `vaport`           | command `vagrant port`                                       |
-| `vapvm`            | command `vagrant plugin install *[ virtualbox \| libvirt ]*` |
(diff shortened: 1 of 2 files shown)


akinomyoga:
Thank you for your contribution! Could you explain about the background of renaming the aliases? It seems just renaming of the aliases to me, and there is no essential changes in the behavior. I feel it will just confuse the existing users who might get get used to the existing alias names.

oscarenzo:
Hello!,

Is for linter and standard purposes, I thought that is better work in the logic of the shortcut name and it should be forward the same pattern.

Thank you.

akinomyoga:
Hello!

> Is for linter and standard purposes,

Could you make it a bit clear? Does it mean some linter warns about these names? What is the *standard*? If you have references, could you provide me with them?

> I thought that is better work in the logic of the shortcut name and it should be forward the same pattern.

What is the logic of the shortcut names? I feel the existing names are already sufficiently logical and predictable from the long name versions. On the other hand, the suggested name doesn't seem to be predictable and logical compared to the existing ones.

Thank you.

oscarenzo:
Hi!

I cancelled the pull request, thank you.

PR #358, opened 2022-09-17
Fixed unbound variable error

Charadon:
git_status_flags variable wasn't in the function git_prompt_vars, so if `set -u` was set in bashrc, it would give unbound variable error constantly.

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -233,9 +233,10 @@ function git_status_summary {
 
 function git_prompt_vars {
   local details=''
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):

akinomyoga:
Thanks for your contribution. I have slightly modified it so that `local` in a later assignment to the variable is removed. If there is no other problems, I'll merge it later.

akinomyoga (approved):

akinomyoga:
I merged it. Thank you again for your contribution!

PR #367, opened 2022-10-24
Terraform aliases

loafoe:
- Initial set of aliases


akinomyoga:
Thank you for your contribution! Could you consider adding this as an `aliases` module instead of as a `plugin` module because this seems to contain only aliases?

loafoe:
Yes sure, i was not aware of alias modules. Will refactor

akinomyoga (approved):

PR #368, opened 2022-10-24
README: add a block with similar projects

vitaly-zdanevich:


akinomyoga:
Thank you for your contribution! I feel there are still other similar projects. For example, oh-my-zsh and bash-it are the important ones that largely affected this project, though the former is not for Bash. I think there are still other similar projects. Also, oh-my-bash, bash-it, and oh-my-zsh provide aliases, plugins, and completions in addition to themes, but at a glance, bash-git-prompt seems to only provide themes. If we also count the collection of themes as similar projects, I think there are even more similar projects. If we add a list of similar projects in README, we would like to maintain it unbiased and comprehensive, or do you have any specific reason that "bash-git-prompt" would be outstanding compared to any other similar projects?

vitaly-zdanevich:
> do you have any specific reason that "bash-git-prompt" would be outstanding compared to any other similar projects?

Just because I use `bash-git-prompt` and recently found this repo.

I have added this repo link to https://github.com/magicmonty/bash-git-prompt/commit/0101261948e2e3b946484990e95a9f8700ace156

I agree that we can increase that list.

vitaly-zdanevich:
[Bash-it ](https://github.com/Bash-it/bash-it) looks like not about bash prompt, maybe we can have multiple categories of similar projects?

akinomyoga:
> [Bash-it ](https://github.com/Bash-it/bash-it) looks like not about bash prompt,

Bash-it calls it *theme*.

If you are talking about prompt settings, there are even more projects of configurable prompt/theme settings for Bash, such as sbp, starship, liquidprompt, etc. These are just examples I know, but I believe there are many others that I do not know.

> maybe we can have multiple categories of similar projects?

If we start to collect the projects of something different types of bash configurations, does that mean we summarize all the types of Bash configurations regardless that it is similar to oh-my-bash? The similarity between oh-my-bash and bash-git-prompt seems to be just both set prompts, but that is the common aspect of most bash configurations. I'm not sure if we should maintain such a list for wild themes/prompts in our README. Shouldn't that be maintained in dedicated projects like [awesome-bash](https://github.com/awesome-lists/awesome-bash#customization)?

PR #374, opened 2022-12-04
plugins/nvm: add nvmrc autoload option

### LINKED ISSUE #371: feature request: automatic nvm use when entering directories with .nvmrc
Hi, thanks a lot for this great project! I started using a few days ago, and I really like the fact that I can get most of the oh-my-zsh feeling without switching the default shell (I ditched Mac OS a long time ago).

I'm using the nvm plugin, and I would like to have the automatic switch of nodejs version when I cd to a directory that has a `.nvmrc` file in it.

I solved this thanks to [this great answer from stackoverflow](https://stackoverflow.com/a/50378304/4149512), adding the following at the end of my `.bashrc`

```
enter_directory() {
  if [[ $PWD == "$PREV_PWD" ]]; then
    return
  fi

  if [[ "$PWD" =~ $PREV_PWD && ! -f ".nvmrc" ]]; then
    return
  fi

  PREV_PWD=$PWD
  if [[ -f ".nvmrc" ]]; then
    nvm use
    NVM_DIRTY=true
  elif [[ $NVM_DIRTY = true ]]; then
    nvm use default
    NVM_DIRTY=false
  fi
}

export PROMPT_COMMAND="$PROMPT_COMMAND; enter_directory"
```

could this be embedded in the nvm plugin?
would you accept a PR with that?

sanzoghenzo:
fixes #371

### DIFF
--- a/plugins/nvm/nvm.plugin.sh
+++ b/plugins/nvm/nvm.plugin.sh
@@ -10,59 +10,90 @@
 # Try to load nvm only if command not already available
 if ! _omb_util_command_exists nvm; then
   # shellcheck source=/dev/null
-  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+  [[ -s $NVM_DIR/nvm.sh ]] && source "$NVM_DIR/nvm.sh"
 fi
 
+#------------------------------------------------------------------------------
 # Optional .nvmrc autoload
-if _omb_util_command_exists nvm && [[ ${OMB_PLUGIN_NVM_AUTO_USE} == true ]]; then
-  find-up () {
-    path=$(pwd)
-    while [[ "$path" != "" && ! -e "$path/$1" ]]; do
-        path=${path%/*}
+#
+# This part is originally from the official nvm documentation
+# (README.md).  Here is the license of the original project:
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2010 Tim Caswell
+# Copyright (c) 2014 Jordan Harband
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+# NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+# BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+if _omb_util_command_exists nvm && [[ ${OMB_PLUGIN_NVM_AUTO_USE-} == true ]]; then
+  _omb_plugin_nvm_find_up() {
+    local path=$PWD
+    while [[ $path && ! -e $path/$1 ]]; do
+      path=${path%/*}
     done
     echo "$path"
   }
   
-  cdnvm(){
-    cd "$@" || return $?
-    nvm_path=$(find-up .nvmrc | tr -d '\n')
+  _omb_plugin_nvm_cd(){
+    cd "$@" || return "$?"
+    local nvm_path=$(_omb_plugin_nvm_find_up .nvmrc)
 
     # If there are no .nvmrc file, use the default nvm version
-    if [[ ! $nvm_path = *[^[:space:]]* ]]; then
+    if [[ $nvm_path != *[^[:space:]]* ]]; then
 
-      declare default_version;
-      default_version=$(nvm version default);
+      local default_version
+      default_version=$(nvm version default)
 
       # If there is no default version, set it to `node`
       # This will use the latest version on your machine
       if [[ $default_version == "N/A" ]]; then
-          nvm alias default node;
-          default_version=$(nvm version default);
+        nvm alias default node
+        default_version=$(nvm version default)
       fi
 
       # If the current version is not the default version, set it to use the default version
       if [[ $(nvm current) != "$default_version" ]]; then
-          nvm use default;
+        nvm use default
       fi
 
     elif [[ -s $nvm_path/.nvmrc && -r $nvm_path/.nvmrc ]]; then
-      declare nvm_version
-      nvm_version=$(<"$nvm_path"/.nvmrc)
+      local nvm_version
+      nvm_version=$(< "$nvm_path"/.nvmrc)
 
       # Add the `v` suffix if it does not exists in the .nvmrc file
       if [[ $nvm_version != v* ]]; then
-          nvm_version="v""$nvm_version"
+        nvm_version=v$nvm_version
       fi
 
       # If it is not already installed, install it
-      if [[ $(nvm ls "$nvm_version" | tr -d '[:space:]') == "N/A" ]]; then
-          nvm install "$nvm_version";
+      if nvm ls "$nvm_version" | grep -qx '[[:space:]]*N/A[[:space:]]*'; then
+        nvm install "$nvm_version"
       fi
 
       if [[ $(nvm current) != "$nvm_version" ]]; then
-          nvm use "$nvm_version";
+        nvm use "$nvm_version"
       fi
     fi
   }
-  alias cd='cdnvm'
+  alias cd='_omb_plugin_nvm_cd'
 fi
+#------------------------------------------------------------------------------
--- a/plugins/nvm/nvm.plugin.sh
+++ b/plugins/nvm/nvm.plugin.sh
@@ -1,4 +1,7 @@
 #! bash oh-my-bash.module
+# Description: automatically load nvm 
+#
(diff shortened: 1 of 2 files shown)


plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+    path=$(pwd)
+    while [[ "$path" != "" && ! -e "$path/$1" ]]; do
+        path=${path%/*}
+    done
+    echo "$path"
+  }
+  
+  cdnvm(){
+    cd "$@" || return $?
+    nvm_path=$(find-up .nvmrc | tr -d '\n')
```
akinomyoga, line 28:
```suggestion
    local nvm_path=$(find-up .nvmrc | tr -d '\n')
```

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+
+    # If there are no .nvmrc file, use the default nvm version
+    if [[ ! $nvm_path = *[^[:space:]]* ]]; then
+
+      declare default_version;
```
akinomyoga, line 33:
```suggestion
      local default_version;
```
`declare` in a function scope is equivalent to `local`, but I prefer to unify the uses to `local` in the codebase of OMB.

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+          default_version=$(nvm version default);
+      fi
+
+      # If the current version is not the default version, set it to use the default version
+      if [[ $(nvm current) != "$default_version" ]]; then
+          nvm use default;
+      fi
+
+    elif [[ -s $nvm_path/.nvmrc && -r $nvm_path/.nvmrc ]]; then
+      declare nvm_version
```
akinomyoga, line 49:
```suggestion
      local nvm_version
```

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
-  # [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+  # shellcheck source=/dev/null
+  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+fi
+
+# Optional .nvmrc autoload
+if _omb_util_command_exists nvm && [[ ${OMB_PLUGIN_NVM_AUTO_USE} == true ]]; then
+  find-up () {
+    path=$(pwd)
+    while [[ "$path" != "" && ! -e "$path/$1" ]]; do
```
akinomyoga, line 20:
```suggestion
    while [[ $path && ! -e $path/$1 ]]; do
```

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+    done
+    echo "$path"
+  }
+  
+  cdnvm(){
+    cd "$@" || return $?
+    nvm_path=$(find-up .nvmrc | tr -d '\n')
+
+    # If there are no .nvmrc file, use the default nvm version
+    if [[ ! $nvm_path = *[^[:space:]]* ]]; then
```
akinomyoga, line 31:
```suggestion
    if [[ $nvm_path != *[^[:space:]]* ]]; then
```

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+      declare nvm_version
+      nvm_version=$(<"$nvm_path"/.nvmrc)
+
+      # Add the `v` suffix if it does not exists in the .nvmrc file
+      if [[ $nvm_version != v* ]]; then
+          nvm_version="v""$nvm_version"
+      fi
+
+      # If it is not already installed, install it
+      if [[ $(nvm ls "$nvm_version" | tr -d '[:space:]') == "N/A" ]]; then
```
akinomyoga, line 58:
```suggestion
      if nvm ls "$nvm_version" | grep -qx '[[:space:]]*N/A[[:space:]]*'; then
```

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+          nvm use default;
+      fi
+
+    elif [[ -s $nvm_path/.nvmrc && -r $nvm_path/.nvmrc ]]; then
+      declare nvm_version
+      nvm_version=$(<"$nvm_path"/.nvmrc)
+
+      # Add the `v` suffix if it does not exists in the .nvmrc file
+      if [[ $nvm_version != v* ]]; then
+          nvm_version="v""$nvm_version"
```
akinomyoga, line 54:
```suggestion
          nvm_version=v$nvm_version
```

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
 [[ -z "$NVM_DIR" ]] && export NVM_DIR="$HOME/.nvm"
 
 
 # Try to load nvm only if command not already available
 if ! _omb_util_command_exists nvm; then
-  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
-  # This is done as part of completions!!!
-  # [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+  # shellcheck source=/dev/null
+  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```
akinomyoga, line 13:
```suggestion
  [[ -s $NVM_DIR/nvm.sh ]] && source "$NVM_DIR/nvm.sh"
```
I think it is good to update it to use `[[ ... ]]` in this chance.

akinomyoga (commented):
Thank you for your contribution. Looks good. I suggest only cosmetic changes.

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+fi
+
+# Optional .nvmrc autoload
+if _omb_util_command_exists nvm && [[ ${OMB_PLUGIN_NVM_AUTO_USE} == true ]]; then
+  find-up () {
```
akinomyoga, line 18: Could you rename this function to something like `_omb_plugin_nvm_find_up`?

Or, is this function supposed to be also directly used by users in the command line? OMB tries not to affect the global namespace of the commands unless it is not the command intended to be directly used by the users. Instead, we confine internal functions and commands in the `_omb_*` namespace.

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+    if [[ ! $nvm_path = *[^[:space:]]* ]]; then
+
+      declare default_version;
+      default_version=$(nvm version default);
+
+      # If there is no default version, set it to `node`
+      # This will use the latest version on your machine
+      if [[ $default_version == "N/A" ]]; then
+          nvm alias default node;
+          default_version=$(nvm version default);
```
akinomyoga, line 40: Why do these lines explicitly terminated by `;`? Can you remove these semicolons?

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+  find-up () {
+    path=$(pwd)
+    while [[ "$path" != "" && ! -e "$path/$1" ]]; do
+        path=${path%/*}
+    done
+    echo "$path"
+  }
+  
+  cdnvm(){
+    cd "$@" || return $?
```
akinomyoga, line 27:
```suggestion
    cd "$@" || return "$?"
```

In case `IFS` has an unusual value like `IFS=0`.

akinomyoga (commented):
These are additional points.

plugins/nvm/nvm.plugin.sh
akinomyoga: By the way, why do we remove newlines here?

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+  # shellcheck source=/dev/null
+  [[ -s $NVM_DIR/nvm.sh ]] && source "$NVM_DIR/nvm.sh"
+fi
+
+# Optional .nvmrc autoload
+if _omb_util_command_exists nvm && [[ ${OMB_PLUGIN_NVM_AUTO_USE-} == true ]]; then
+  _omb_plugin_nvm_find_up () {
+    local path=$PWD
+    while [[ $path && ! -e $path/$1 ]]; do
+        path=${path%/*}
```
akinomyoga, line 21:
```suggestion
      path=${path%/*}
```

I noticed that the indentation is different from the upper level. There are also similar inconsistencies in the function `cdnvm`.

sanzoghenzo:
Thanks for the review, I didn't put much thought on the PR, and I also am not too experienced with bash script, so I just copied the example from the official nvm documentation!
I fixed everything!

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+    done
+    echo "$path"
+  }
+  
+  cdnvm(){
```
akinomyoga, line 26: Maybe we can also rename this function `cdnvm` to `_omb_plugin_nvm_cd` or something because it is not directly input by the user?

plugins/nvm/nvm.plugin.sh
```diff
@@ -1,12 +1,68 @@
...
+      fi
+
+      # If the current version is not the default version, set it to use the default version
+      if [[ $(nvm current) != "$default_version" ]]; then
+          nvm use default
+      fi
+
+    elif [[ -s $nvm_path/.nvmrc && -r $nvm_path/.nvmrc ]]; then
+      local nvm_version
+      nvm_version=$(<"$nvm_path"/.nvmrc)
```
akinomyoga, line 50:
```suggestion
      nvm_version=$(<"$nvm_path"/.nvmrc)
```

This is really cosmetic but I would like to add a space here. This is just my preference, but what do you think?

akinomyoga (commented):
Thank you for updating! I gave another round. Could you check these?

akinomyoga (dismissed):

akinomyoga:
> so I just copied the example from the official nvm documentation!

Ah, is this from [README of nvm-sh/nvm](https://github.com/nvm-sh/nvm#bash). I'm not sure if the license of nvm (the MIT license) applies to the code example in README, but if it is the case, we need to retain the copyright notice. I'll add it.

akinomyoga:
I have edited the commit message and also squashed additional commits.

akinomyoga (approved):

akinomyoga:
If there are no further points, I'll keep this open for a few days and merge it later. @sanzoghenzo Thank you again for your contribution and quick responses!

PR #376, opened 2022-12-06
lib: provide a prompt function for the dirstack

jjmcdn:
Providing this additional function for prompts allows anyone using pushd/popd/dirs to navigate directory trees to quickly view their stack and find which entry they want to switch to.

This degrades gracefully for anyone not using the directory stack since dirs will always display the same as `\w` when the stack is only one entry deep.

The bakke theme has been tweaked to use the new function as an example.

Signed-off-by: Joe MacDonald <joe@deserted.net>

For what it looks like when it use, I'm adding a screenshot here.

2022-12-06-123744_3840x1080_scrot

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -52,6 +52,9 @@ SCM_SVN_CHAR='⑆'
 SCM_NONE='NONE'
 SCM_NONE_CHAR='○'
 
+DIRSTACK_EXPAND_TILDE=${DIRSTACK_EXPAND_TILDE:=false}
+DIRSTACK_LIMIT=${DIRSTACK_LIMIT:=0}
+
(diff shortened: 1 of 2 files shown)


akinomyoga:
I have a question. Is it different from directly putting `$(dirs)` in `PS1`?

akinomyoga:
OK. Thank you for the clarification! I think I will keep this PR open as an "example configuration".

jjmcdn:
> OK. Thank you for the clarification! I think I will keep this PR open as an "example configuration".

Okay, sorry about the delayed follow up on this.  I added the full feature as I'd intended it, with the optional `-l` expansion and limiting the number of dirstack entries from an environment variable.  It's only slightly more complex than the simple `dirs` one I'd initially proposed, so it's probably still useful as an example for others.

Thanks!

Issue #379, opened 2022-12-12
Missing terraform completions an `t` alias for terraform


TechDufus:
Can we get tab completion on the `t` alias instead of just terraform? I assume some sort of detection will need to go [here](https://github.com/ohmybash/oh-my-bash/blob/60332ed8fb954f6e6c0af243933e67d570ffc2cc/completions/terraform.completion.sh#L26) to ensure that the `t` alias is defined before leading completion for it....

akinomyoga:
Closing in favor of #380

PR #380, opened 2022-12-12
adding t alias conditianal completion

TechDufus:
This PR closes https://github.com/ohmybash/oh-my-bash/issues/379

Since aliases are sourced before completions, we can check for the existence of the alias at this point and conditionally load completion for it here.

### DIFF
--- a/completions/terraform.completion.sh
+++ b/completions/terraform.completion.sh
@@ -24,3 +24,8 @@ _terraform()
         return 0
 } &&
 complete -F _terraform terraform
+
+# If the terraform alias is defined, add completion for the alias
+if alias -p | grep -q "t='terraform'"; then
+    complete -F _terraform t
+fi


TechDufus:
I think you're right.. I'm just having this setting clash with `complete -D` which is why I'm having issues... This can probably just be dropped...

akinomyoga:
Thank you for your reply. I now see the problem. I guess this should be fixed at the side of Bash or the default completion function of `complete -D -F func`.

I checked the mailing list of Bash and found a thread [1] where the discussion seemed to have stopped in the middle, and the change did not happen. I also checked the implementation of `bash-completion`'s `__load_completion` [2] (which is used by bash-completion's default completion), but it doesn't seem to handle the aliases. There is actually a Bash configuration `cykerway/complete-alias` [3] to handle aliases, and it is suggested to merge it into `bash-completion` [4], which haven't yet happened actually. On the other hand, `progcomp_alias` has a higher priority than the default completion of `complete -D` in ble.sh [5], 

- [1] https://lists.gnu.org/archive/html/bug-bash/2019-01/msg00101.html
- [2] [bash-completion/bash_completion at bbe2f099 · scop/bash-completion](https://github.com/scop/bash-completion/blob/bbe2f0995b99cebf1c67fb963b67c5e14bb3583d/bash_completion#L2612)
- [3] [cykerway/complete-alias: automagical shell alias completion;](https://github.com/cykerway/complete-alias)
- [4] [adding an alias completion helper · Issue #521 · scop/bash-completion](https://github.com/scop/bash-completion/issues/521)
- [5] [ble.sh/core-complete.sh at b568ade0 · akinomyoga/ble.sh](https://github.com/akinomyoga/ble.sh/blob/b568ade0c7a6b85b8fe98cb56182f139c75ffbff/lib/core-complete.sh#L3805-L3827)

PR #383, opened 2023-01-07
Allow CASE_SENSITIVE and HYPHEN_INSENSITIVE to work as documented

underwoo:
Adjusted shopt.sh to use CASE_SENSITIVe and HYPHEN_INSENSITIVE variables to affect completion.

Note, this changes the old default hyphen insensitive completion behavior, the user will need to set HYPHEN_INSENSITIVE="true" to have the old default behavior

### DIFF
--- a/lib/shopt.sh
+++ b/lib/shopt.sh
@@ -46,7 +46,13 @@ else
 
 	# Treat hyphens and underscores as equivalent
 	# CASE_SENSITIVE must be off
-	if [[ ${OMB_HYPHEN_INSENSITIVE:-${HYPHEN_INSENSITIVE:-false}} == true ]]; then
+	if [[ ! ${OMB_HYPHEN_SENSITIVE-} && ${HYPHEN_INSENSITIVE} ]]; then
+		case $HYPHEN_INSENSITIVE in
+		(true)  OMB_HYPHEN_SENSITIVE=true ;;
+		(false) OMB_HYPHEN_SENSITIVE=false ;;
+		esac
+	fi
+	if [[ ${OMB_HYPHEN_SENSITIVE-} == false ]]; then
 		bind "set completion-map-case on"
 	fi
 fi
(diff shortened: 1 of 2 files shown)


lib/shopt.sh
```diff
@@ -27,10 +27,17 @@ shopt -s nocaseglob;
...
-
-# Treat hyphens and underscores as equivalent
-bind "set completion-map-case on"
+if [ ${CASE_SENSITIVE:-false} != "true" ]
+then
+  bind "set completion-ignore-case on"
+
+  # Treat hyphens and underscores as equivalent
+  # CASE_SENSITIVE must be off
+  if [ ${HYPHEN_INSENSITIVE:-false} = "true" ]
```
akinomyoga, line 36: Although ohmyzsh has the setting with the name `HYPHEN_INSENSITIVE`, I feel we could use `HYPHEN_SENSITIVE` to `HYPHEN_"IN"SENSITIVE` because it is consistent with `CASE_SENSITIVE` and also negative boolean variables should be generally avoided. Or another way is to use the names like `IGNORE_CASE` and `IGNORE_HYPHEN` (and revert the meaning of `CASE_SENSITIVE`).

akinomyoga (commented):
Thanks for your contribution!

This is partially a duplicate of #180, which also attempts to fix `CASE_SENSITIVE`. I'll also look at #180 again.

Another thing to discuss is that we would like to gradually move to the `OMB_*` as discussed in #280. However, the original ohmyzsh seems to have the setting with the unprefixed `CASE_SENSITIVE`.

lib/shopt.sh
akinomyoga: I have changed the variable name to `HYPHEN_SENSITIVE`.

I decided to keep `SENSITIVE` instead of switching to `IGNORE` for now.

akinomyoga (approved):

PR #384, opened 2023-01-07
Add check for notify-send command prior to setting alert alias

underwoo:


akinomyoga (dismissed):
Thanks for your contributions.

I'll later check whether there are other similar cases in the codebase. Or have you already checked the codebase and found that it is the only occurrence?

akinomyoga (approved):

akinomyoga:
There seem to be not so many similar cases. Some basic commands such as `sudo` are not always available in all the systems, but I decided not to check the existence of `sudo` for the aliases that use `sudo` because it is hard to tell the reason when the aliases are not actually defined.

I have edited the commit message of the original commit and force-pushed.

akinomyoga:
I'll merge this PR after waiting for a few days.

underwoo:
@akinomyoga thank you for reviewing and approving this PR.  Removing the hardcoded paths for the aliases is a good idea.  As for `sudo` in those aliases, It might be better to use the `OMB_USE_SUDO` option to decide if `sudo` is added to the alias or not.

I will check through to see if I find any additional locations that should check if the application exists.  

Although it is beyond the scope of this PR, the network interface names in the `ipInfo[12]` are not universal.  And, it may be worth splitting the `aliases/misc.aliases.sh` aliases into smaller, more specific alias groups (one file per section currently in the file).  To keep the same behavior, you could then have `misc.aliases.sh` source all the new files.  This would allow a user to enable a subset of the misc.aliases.

akinomyoga:
> I will check through to see if I find any additional locations that should check if the application exists.

If there are no further updates, I'll consider merging this PR.

> Although it is beyond the scope of this PR, the network interface names in the `ipInfo[12]` are not universal. And, it may be worth splitting the `aliases/misc.aliases.sh` aliases into smaller, more specific alias groups (one file per section currently in the file). To keep the same behavior, you could then have `misc.aliases.sh` source all the new files. This would allow a user to enable a subset of the misc.aliases.

These settings stemmed from [a blog post](https://natelandau.com/my-mac-osx-bash_profile/). I'm not sure if we should keep maintaining these aliases actively. To be honest, we might remove too specific aliases, but I just keep them for backward compatibility. The new users just do not need to load `misc` aliases.

akinomyoga (approved):

PR #386, opened 2023-01-12
feat: adding gcloud plugin

ninoCan:
This plugin provides support for gcloud and loads its completions

### DIFF
--- a/plugins/gcloud/README.md
+++ b/plugins/gcloud/README.md
@@ -1,0 +1,25 @@
+# gcloud
+
+This plugin provides completion support for the
+[Google Cloud SDK CLI](https://cloud.google.com/sdk/gcloud/).
+Based on [gcloud plugin for oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/gcloud).
+
+To use it, add `gcloud` to the plugins array in your bashrc file.
+
+```bash
+plugins=(... gcloud)
+```
+
+It relies on you having installed the SDK using one of the supported options
+listed [here](https://cloud.google.com/sdk/install).
+
+## Plugin Options
+
+* Set `OMB_PLUGIN_GCLOUD_HOME` in your `bashrc` file before you load oh-my-bash if you have
+your GCloud SDK installed in a non-standard location. The plugin will use this
+as the base for your SDK if it finds it set already.
+
+* If you do not have a `python2` in your `PATH` you'll also need to set the
+`CLOUDSDK_PYTHON` environment variable at the end of your `.bashrc`. This is
+used by the SDK to call a compatible interpreter when you run one of the
+SDK commands.
--- a/plugins/gcloud/gcloud.plugin.sh
+++ b/plugins/gcloud/gcloud.plugin.sh
@@ -1,0 +1,56 @@
+#!/bin/bash oh-my-bash.module
+# gcloud.plugin.sh
+# Author: Ian Chesal (github.com/ianchesal) -- ohmyzsh/gloud.plugin.zsh
+# Author: Antonino Cangialosi (github.com/ninoCan)
+# Fork of oh-my-zsh gcloud plugin
+
+function _omb_plugin_gcloud_set_home_var() {
+  if [[ ! ${OMB_PLUGIN_GCLOUD_HOME-} ]]; then
+    local -a search_locations=(
+      "$HOME/google-cloud-sdk"
+      "/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk"
+      "/opt/homebrew/Caskroom/google-cloud-sdk/latest/google-cloud-sdk"
+      "/usr/share/google-cloud-sdk"
+      "/snap/google-cloud-sdk/current"
+      "/usr/lib/google-cloud-sdk"
+      "/usr/lib64/google-cloud-sdk"
+      "/opt/google-cloud-sdk"
+      "/opt/local/libexec/google-cloud-sdk"
+    )
+
+    local gcloud_sdk_location
+    for gcloud_sdk_location in "${search_locations[@]}"; do
+      if [[ -d $gcloud_sdk_location ]]; then
+        OMB_PLUGIN_GCLOUD_HOME=$gcloud_sdk_location
+        break
+      fi
+    done
+  fi
+}
+
+function _omb_plugin_gcloud_set_completions() {
+  if [[ ${OMB_PLUGIN_GCLOUD_HOME-} ]]; then
+    # Only source this if gcloud isn't already on the path
+    if _omb_util_binary_exists gcloud; then
+      if [[ -f $OMB_PLUGIN_GCLOUD_HOME/path.bash.inc ]]; then
+        source "$OMB_PLUGIN_GCLOUD_HOME/path.bash.inc"
+      fi
+
+      # Look for completion file in different paths
+      local -a completion_filepath_candidates=(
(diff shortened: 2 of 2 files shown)


plugins/gcloud/gcloud.plugin.sh
```diff
@@ -0,0 +1,49 @@
+#! bash oh-my-bash.module
+# gcloud.plugin.sh
+# Author: Antonino Cangialosi (github.com/ninoCan)
+# Fork of oh-my-bash gcloud plugin
```
akinomyoga, line 4: Did you mean the following?

```suggestion
# Fork of oh-my-zsh gcloud plugin
```

plugins/gcloud/gcloud.plugin.sh
```diff
@@ -0,0 +1,49 @@
+#! bash oh-my-bash.module
+# gcloud.plugin.sh
+# Author: Antonino Cangialosi (github.com/ninoCan)
```
akinomyoga, line 3: The original code is a part of Oh My Zsh, which is licensed with MIT License, so you need to retain the original copyright notice. The MIT License isn't meant to waive the copyright.
```suggestion
# Author: Ian Chesal (github.com/ianchesal) -- ohmyzsh/gloud.plugin.zsh
# Author: Antonino Cangialosi (github.com/ninoCan)
```

plugins/gcloud/gcloud.plugin.sh
```diff
@@ -0,0 +1,49 @@
...
+            break
+        fi
+    done
+    unset search_locations gcloud_sdk_location
+fi
+
+if [[ -n $CLOUDSDK_HOME ]]; then
+    which gcloud &> /dev/null
+    # Only source this if gcloud isn't already on the path
+    if [[ $? -ne 0 ]]; then
```
akinomyoga, line 31:
```suggestion
    # Only source this if gcloud isn't already on the path
    if _omb_util_binary_exists gloud; then
```

plugins/gcloud/README.md
```diff
@@ -0,0 +1,25 @@
...
+```bash
+plugins=(... gcloud)
+```
+
+It relies on you having installed the SDK using one of the supported options
+listed [here](https://cloud.google.com/sdk/install).
+
+## Plugin Options
+
+* Set `CLOUDSDK_HOME` in your `bashrc` file before you load oh-my-bash if you have
```
akinomyoga, line 18: As far as I search for the string `CLOUDSDK_HOME`, the variable seems to be specific to the oh-my-zsh plugin. For the shell variables used by Oh My Bash, we are gradually applying the naming convention discussed in #280. In particular, we would like to adopt the naming convention from the beginning to avoid unnecessary later migrations. For the present purpose, for example, `OMB_PLUGIN_GCLOUD_HOME` follows the naming convention.
```suggestion
* Set `OMB_PLUGIN_GCLOUD_HOME` in your `bashrc` file before you load oh-my-bash if you have
```

(The other occurrences of the name `CLOUDSDK_HOME` also need to be updated.)

plugins/gcloud/gcloud.plugin.sh
```diff
@@ -0,0 +1,49 @@
...
+        "/opt/local/libexec/google-cloud-sdk"
+    )
+
+    for gcloud_sdk_location in $search_locations; do
+        if [[ -d "${gcloud_sdk_location}" ]]; then
+            CLOUDSDK_HOME="${gcloud_sdk_location}"
+            break
+        fi
+    done
+    unset search_locations gcloud_sdk_location
```
akinomyoga, line 25: Could you put the entire code in a shell function and declare these variables as `local` so that we do not affect the global variables? FWIW, the naming convention #280 suggests the function name of the form `_omb_plugin_gloud_xxxxx` (such as `_omb_plugin_gloud_initialize`).

In that way, we do not need to unset the used variables and also can avoid possible conflicts with the global variables of the same names defined in other places. Or otherwise, please rename these temporary global variables to have the names of `_omb_plugin_gloud_xxxx` (#280).

plugins/gcloud/gcloud.plugin.sh
```diff
@@ -0,0 +1,49 @@
...
+            "${CLOUDSDK_HOME}/completion.bash.inc"             # default location
+            "/usr/share/google-cloud-sdk/completion.bash.inc"  # apt-based location
+        )
+        for comp_file in "$completion_filepath_candidates"; do
+            if [[ -f "$comp_file" ]]; then
+                source "$comp_file"
+                break
+            fi
+        done
+        unset comp_file
```
akinomyoga, line 47: The variable `completion_filepath_candidates` also needs to be unset or declared to be `local` after putting the code inside a function.

plugins/gcloud/gcloud.plugin.sh
```diff
@@ -0,0 +1,49 @@
...
+    fi
+fi
```
akinomyoga, line 49: Please add a newline at the end of the file.

akinomyoga (commented):
Thank you for your contribution! I'd like to merge this after the adjustments suggested below.

plugins/gcloud/gcloud.plugin.sh
ninoCan: Oh, thanks! I didn't pay attention to the license. My bad!

plugins/gcloud/gcloud.plugin.sh
ninoCan: Yes, I did. It was most likely overwritten by a batch substitution!

plugins/gcloud/README.md
ninoCan: You are correct! I did assume `CLOUDSDK_HOME` was an internal variable to `gcloud`. Yet, perusing the source code of the installation script, it can be seen they define it as `CLOUDSDK_ROOT_DIR` and it's used only as a local variable in the installation phase. I think sticking to the namespacing it's a very sensible choice!

plugins/gcloud/gcloud.plugin.sh
ninoCan: I will asap!

plugins/gcloud/gcloud.plugin.sh
ninoCan: Nice one!

plugins/gcloud/gcloud.plugin.sh
ninoCan: will do

plugins/gcloud/gcloud.plugin.sh
ninoCan: :+1:

ninoCan:
I apologize for the delay! Here's the modifications you asked!

akinomyoga:
I applied the change 97e8fb64635e446545a80f57b104488f6cafc211 to make it work with `set -u` (`set -o nounset`). When the user set `set -u`, referencing undefined variables as `$var` causes errors. Instead, the construct `${var-}` can be used too make it explicit that the undefined variables should be treated as an empty value.

akinomyoga:
Thanks! I also added other commits, 9085cac, 33e6823, and 1a3cff2, but these are all cosmetic style changes. I'll later squash these commits and merge this PR if there are no further problems.

akinomyoga (approved):

akinomyoga (approved):

akinomyoga (approved):

akinomyoga:
Thanks. I have merged it!

PR #387, opened 2023-01-17
themes/axin: Show Virtual Env

AhmadMWaddah:
Edit terminal prompt to add Python Virtual Environment name before terminal prompt for developers when Virtual Environment is activated.

PR #373, opened 2022-11-29
themes/sexy: display active conda environment's name in bash prompt

walidabualafia:
Added functionality to display the name of an active Conda environment for `themes/sexy`. Functionality is generalizable to all other themes.

- Edited `_omb_theme_PROMPT_COMMAND()` to check for whether a Conda environment is active (If there is a conda environment active, then `$CONDA_SHLVL == 1`)
- Resolved environment name in the `$PS1` variable by invoking `$(basename $CONDA_DEFAULT_ENV)`
- colorized it to match the theme

This is what it looks like when you activate a Conda environment, and after you deactivate a Conda environment:
Screenshot 2022-11-29 at 06 43 48

Issue #372, opened 2022-11-23
add virtualenv rr theme


gab-25:
hello all, it is possible to add the visualization of the virtual environment in the rr theme

gab-25:
thus fixing the function I was able to add the virtual environment.
function _omb_theme_PROMPT_COMMAND() {
    local user_name="${_omb_prompt_white}\u${_omb_prompt_reset_color}"
    local base_directory="${_omb_prompt_bold_blue}\W${_omb_prompt_reset_color}"
    local GIT_THEME_PROMPT_PREFIX="${_omb_prompt_bold_purple}git:(${_omb_prompt_reset_color}"
    local SVN_THEME_PROMPT_PREFIX="${_omb_prompt_bold_purple}svn:(${_omb_prompt_reset_color}"
    local HG_THEME_PROMPT_PREFIX="${_omb_prompt_bold_purple}hg:(${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_SUFFIX="${_omb_prompt_bold_purple})${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_CLEAN="${_omb_prompt_bold_green} ✓${_omb_prompt_reset_color}"
    local SCM_THEME_PROMPT_DIRTY="${_omb_prompt_bold_red} ✗${_omb_prompt_reset_color}"
    local arrow="${_omb_prompt_bold_purple}➜${_omb_prompt_reset_color}"

    PS1="${arrow}  ${user_name} ${base_directory} "

    local VIRTUALENV_THEME_PROMPT_PREFIX="${_omb_prompt_bold_purple}vitualenv:(${_omb_prompt_reset_color}"
    local VIRTUALENV_THEME_PROMPT_SUFFIX="${_omb_prompt_bold_purple}) ${_omb_prompt_reset_color}"
    local python_venv; _omb_prompt_get_python_venv
  
    PS1+=${python_venv}

    local scm_info=$(scm_prompt_info)

    PS1+=${scm_info:+$scm_info }
    PS1+=${_omb_prompt_normal}
}

akinomyoga:
I picked the suggestion in PR #400.

PR #282, opened 2022-01-04
themes/agnoster: Add conda env prompt

KimBioInfoStudio:

PR #393, opened 2023-02-04
themes/luan: Show Python virtualenv

Askaiant:
Edited the prompt so it shows the active python virtualenv to the user when it is active.

### DIFF
--- a/themes/luan/luan.theme.sh
+++ b/themes/luan/luan.theme.sh
@@ -14,6 +14,10 @@ RVM_THEME_PROMPT_PREFIX=""
 RVM_THEME_PROMPT_SUFFIX=""
 
 function _omb_theme_PROMPT_COMMAND() {
+    python_venv=""
+    if [[ -n "${VIRTUAL_ENV}" ]]; then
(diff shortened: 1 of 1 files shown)


akinomyoga:
Thanks for the contribution. I have been thinking that what would be shown in the prompt would largely depend on personal preferences, so I have been marking the PRs that try to add an element to the prompt as customization examples. However, particularly for the virtualenv and the condaenv, we have so many similar PRs including this one, #387, #373, #335, and #282.

Now maybe we should introduce a configuration variable like `OMB_PROMPT_SHOW_PYTHON_VENV=true`, etc., and include the venv elements in all the themes. One non-trivial stuff is that we need to carefully choose the design of the new element that matches each theme one by one. Hmm, maybe we can give up supporting it all at once and separately apply PRs to the themes that the PR has appeared.

Askaiant:
I see what you mean. I will close this PR in place of #394. This way we could give the option to personalize whether the prompt is shown or not. As for including the element in the theme, this could be done progressively per request (e.g. #372 , #117 , #210 , #387 , #373 and #335 ) but would have to somehow match the style, would be happy to help.

PR #394, opened 2023-02-07
Define python venv as optional

Askaiant:
PR in place of #393.


lib/omb-prompt-base.sh
```diff
@@ -85,6 +85,8 @@ CONDAENV_THEME_PROMPT_SUFFIX='|'
 PYTHON_THEME_PROMPT_PREFIX=' |'
 PYTHON_THEME_PROMPT_SUFFIX='|'
 
+OMB_PROMPT_SHOW_PYTHON_VENV=${OMB_PROMPT_SHOW_PYTHON_VENV:=true}
```
akinomyoga, line 88: I would like to make each theme selects the default. Currently, some themes show the information while others do not. I would like to leave the default behavior rather than make all the themes show the information. I wouldn't like to change the default of the themes, which is related to the reason that I mark similar PRs as "Customization Examples" and kept open.

You can put this line in the theme files that use the python venv information.

akinomyoga (commented):
Great! I think this is a good starting point.

lib/omb-prompt-base.sh
akinomyoga: I'll fix it.

themes/luan/luan.theme.sh
```diff
@@ -10,20 +10,23 @@ GIT_THEME_PROMPT_CLEAN=" ${_omb_prompt_bold_green}✓"
...
+  local ruby_env python_venv
+  _omb_prompt_get_ruby_env && ruby_env="$_omb_prompt_bold_brown$ruby_env$_omb_prompt_normal "
+  _omb_prompt_get_python_venv
+  local git_branch="$(scm_prompt_info)${_omb_prompt_normal}"
+  local prompt="${_omb_prompt_bold_green}\$${_omb_prompt_normal} "
+  local arrow="${_omb_prompt_bold_white}▶${_omb_prompt_normal} "
+  local prompt="${_omb_prompt_bold_green}\$${_omb_prompt_normal} "
 
-    PS1="${dtime}${user_host}:${current_dir} ${rvm_ruby} ${git_branch}
+  PS1="${dtime}${python_venv}${user_host}:${current_dir} ${ruby_env}${git_branch}
```
akinomyoga, line 29: @Askaiant The `luan` theme already seems to have the ruby-env information in the later part after `${current_dir}`. It seems to me that there is no reason to include two similar segments in different positions in the prompt. Can we move the python-env information so that `${ruby_env}` and `${python_venv}` are located next to each other?

lib/omb-prompt-base.sh
akinomyoga: I have applied changes. Also, I added a commit 86ca670af7b550481ab8da2a90d34c6f5d8d8e5d that disables the python venv in `luan` by default, but I haven't yet actually determined which should be the default.

Askaiant: Sounds good to me, makes sense since the themes are personalized. My thought process was to have a global setting, or rather something that could be shared between themes.

themes/luan/luan.theme.sh
Askaiant: Agree that that would be the most natural spot perhaps for this theme, this was kinda my own personal preference as I liked the default bash format of venv shown like [this](https://i.stack.imgur.com/3zUIW.png). Just to pick your brain a bit, what's your thought on adding venv on start of the prompt (like the default bash) for themes that haven't implemented the "show_python_venv"?

akinomyoga: Yes! I think it would be a good default to put it at the start of the prompt by default unless there are no conflicts or a design issue like this.

Anyway, I would like to first merge this PR and then pick up the other PRs https://github.com/ohmybash/oh-my-bash/pull/387, https://github.com/ohmybash/oh-my-bash/pull/373, https://github.com/ohmybash/oh-my-bash/pull/335, and https://github.com/ohmybash/oh-my-bash/pull/282 after adjustments. I think this is a good way to count the contributions of these people. I'd like to process the other themes after that if necessary.

akinomyoga (approved):

themes/luan/luan.theme.sh
akinomyoga: I have moved `${python_venv}` just before `${ruby_env}`

akinomyoga (approved):
I have rebased and squashed some commits.

akinomyoga:
Thanks! I've merged it! I'll proceed to process other PRs https://github.com/ohmybash/oh-my-bash/pull/387, https://github.com/ohmybash/oh-my-bash/pull/373, https://github.com/ohmybash/oh-my-bash/pull/335, and https://github.com/ohmybash/oh-my-bash/pull/282.

PR #399, opened 2023-02-15
Rework PR#180 to address comments and add case sensitivity to command completion

jjmcdn:
The CASE_SENSITIVE setting was not being checked when setting case-insensitive globbing / matching.  Add checks in the appropriate places, preserving the default (insensitive) behaviour unless explicitly
changed.

Further, this will also apply to expansions so that case-sensitivity will apply in an intuitive way, treating paths, filename completions and command completions consistently.

Signed-off-by: Joe MacDonald <joe.macdonald@siemens.com>

This is a rework of #180 attempting to simplify the code there a bit, squash the commits together and address the remaining open suggestion in it.  It's not intended to address the hyphenation / underscore feature in #383 since that seems like a separate discussion and probably warrants a separate setting.  I don't see hyphenation and case sensitivity being related except by coincidence.

### DIFF
--- a/lib/shopt.sh
+++ b/lib/shopt.sh
@@ -19,18 +19,26 @@ PROMPT_DIRTRIM=2
 bind Space:magic-space
 
 # Turn on recursive globbing (enables ** to recurse all directories)
-shopt -s globstar 2> /dev/null                                                                                   
+shopt -s globstar 2> /dev/null
 
-# Case-insensitive globbing (used in pathname expansion)
-shopt -s nocaseglob;
+# Case-sensitive globbing (used in pathname expansion) and matching
+# (used in case, [[]], word expansions and command completions)
+if [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == true ]]; then
+   shopt -u nocaseglob
+   shopt -u nocasematch
+else
+   shopt -s nocaseglob
+   shopt -s nocasematch
+fi
 
 ## SMARTER TAB-COMPLETION (Readline bindings) ##
 
-# Perform file completion in a case insensitive fashion
+# Conditionally perform file completion in a case insensitive fashion.
+# Setting OMB_CASE_SENSITIVE to 'true' will switch from the default,
+# case insensitive, matching to the case-sensitive one
+#
 # Note: CASE_SENSITIVE is the compatibility name
-if [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == false ]]; then
-	bind "set completion-ignore-case on"
-elif [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == true ]]; then
+if [[ ${OMB_CASE_SENSITIVE:-${CASE_SENSITIVE:-}} == true ]]; then
 	bind "set completion-ignore-case off"
 else
 	# By default, case sensitivity is disabled.


lib/shopt.sh
```diff
@@ -19,15 +19,27 @@ PROMPT_DIRTRIM=2
...
+
+# Case-insensitive globbing (used in pathname expansion) and matching (used in
+# case, [[]], word expansions and command completions)
+if [[ ${CASE_SENSITIVE} == true ]]; then
+   shopt -u nocaseglob
+   shopt -u nocasematch
+else
+   shopt -s nocaseglob
+   shopt -s nocasematch
+fi
```
akinomyoga, line 32: This is a good fix.

akinomyoga (approved):

Hmm, I would like to keep the credit to the original submitter, but there arose the second PR #383. I intended first to merge #180 after modification and then to apply the additional changes about `HYPHEN_INSENSITIVE` of #383 after rebasing and conflict resolution, which is a bit cumbersome. But now things have become even more complicated.

> This is a rework of https://github.com/ohmybash/oh-my-bash/pull/180 attempting to simplify the code there a bit, squash the commits together

In the squashing process, why wasn't the original commit author @nextia365 kept? Anyway, I'll merge #180 first and will rebase this PR.

> and address the remaining open suggestion in it.

I also wanted to discuss the naming of the variables `CASE_SENSITIVE` and `HYPHEN_INSENSITIVE` in #383 before starting to process #180.

> I don't see hyphenation and case sensitivity being related except by coincidence.

Readline relates them. The Readline manual [§8.3.1](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html) says

> completion-map-case
>
> If set to ‘on’, and completion-ignore-case is enabled, Readline treats hyphens (‘-’) and underscores (‘_’) as equivalent when performing case-insensitive filename matching and completion. The default value is ‘off’.

akinomyoga:
There seems to be no response from @underwoo about the naming convention, so maybe I can just apply the prefix `OMB_`.

akinomyoga:
I've merged #180. Would you mind rebasing your change on top of the lateset `master`?

There seems to be an unresolved conversation in https://github.com/ohmybash/oh-my-bash/pull/180#discussion_r501544834. The line doesn't seem to be modified, but what do you think of it? I'm not sure what was the exact intent of @sirikid, but I guess we might mention that we *conditionally* perform case-insensitive completions, where `OMB_CASE_SENSITIVE` controls it.

jjmcdn:
> Thank you for your contribution!
> 
> Hmm, I would like to keep the credit to the original submitter, but there arose the second PR #383. I intended first to merge #180 after modification and then to apply the additional changes about `HYPHEN_INSENSITIVE` of #383 after rebasing and conflict resolution, which is a bit cumbersome. But now things have become even more complicated.
> 
> > This is a rework of #180 attempting to simplify the code there a bit, squash the commits together
> 
> In the squashing process, why wasn't the original commit author @nextia365 kept? Anyway, I'll merge #180 first and will rebase this PR.

Oh, the original author, as I read it, was @saul365, not @nextia365.  Since there wasn't a signed-off-by line in the original commit, I wasn't going to add one, but by the time I was done reworking it and squashing the commits, the changes were significant enough that I wanted to reach out to them before putting their name on it.

With #180 merged, though, that resolves the core issue I'd hoped to have fixed, that's great!

> > and address the remaining open suggestion in it.
> 
> I also wanted to discuss the naming of the variables `CASE_SENSITIVE` and `HYPHEN_INSENSITIVE` in #383 before starting to process #180.

Yeah, reading the comments on #383, I agree that the variables should be aligned (both should be `_SENSITIVE` or `_INSENSITIVE`, otherwise it just looks wrong), given that the readline settings are 'set nocaseglob' it may have been better to do them both as `INSENSITIVE` but my rationale for going with `CASE_SENSITIVE` was that the variable was already there and had some history behind it (even if it wasn't being used) and the majority of users seem to be happy with the default being what it is, so having a `false` setting to turn something off that's always been there seems more sensible and intuitive to me.

> > I don't see hyphenation and case sensitivity being related except by coincidence.
> 
> Readline relates them. The Readline manual [§8.3.1](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html) says
> 
> > completion-map-case
> > If set to ‘on’, and completion-ignore-case is enabled, Readline treats hyphens (‘-’) and underscores (‘_’) as equivalent when performing case-insensitive filename matching and completion. The default value is ‘off’.

Sure, I think that's still a sensible discussion to have in #383 rather than trying to do anything with it in this PR.

jjmcdn:
> I've merged #180. Would you mind rebasing your change on top of the lateset `master`?

Sure thing, I just did that and pushed a new version of the commit with an updated commit log.

> There seems to be an unresolved conversation in [#180 (comment)](https://github.com/ohmybash/oh-my-bash/pull/180#discussion_r501544834). The line doesn't seem to be modified, but what do you think of it? I'm not sure what was the exact intent of @sirikid, but I guess we might mention that we _conditionally_ perform case-insensitive completions, where `OMB_CASE_SENSITIVE` controls it.

Oh, yeah, so that part I'd thought I'd addressed in the previous PR, but I realize now I'd only explained things in the set option code and not in the readline part.  I hope that's all clear now in the current commit.

lib/shopt.sh
```diff
@@ -19,18 +19,26 @@ PROMPT_DIRTRIM=2
...
 
 ## SMARTER TAB-COMPLETION (Readline bindings) ##
 
-# Perform file completion in a case insensitive fashion
+# Conditionally perform file completion in a case insensitive fashion.
+# OMB_CASE_SENSITIVE to 'true' will switch from the default, case
+# insensitive, matching.
```
akinomyoga, line 38:
```suggestion
# insensitive, matching to the case-sensitive one.
```
It is confusing to only mention the behavior before applying the setting.

lib/shopt.sh
```diff
@@ -19,18 +19,26 @@ PROMPT_DIRTRIM=2
...
+else
+   shopt -s nocaseglob
+   shopt -s nocasematch
+fi
 
 ## SMARTER TAB-COMPLETION (Readline bindings) ##
 
-# Perform file completion in a case insensitive fashion
+# Conditionally perform file completion in a case insensitive fashion.
+# OMB_CASE_SENSITIVE to 'true' will switch from the default, case
```
akinomyoga, line 37:
```suggestion
# Setting OMB_CASE_SENSITIVE to 'true' will switch from the default, case
```

lib/shopt.sh
```diff
@@ -19,18 +19,26 @@ PROMPT_DIRTRIM=2
 bind Space:magic-space
 
 # Turn on recursive globbing (enables ** to recurse all directories)
-shopt -s globstar 2> /dev/null                                                                                   
+shopt -s globstar 2> /dev/null
 
-# Case-insensitive globbing (used in pathname expansion)
-shopt -s nocaseglob;
+# Case-insensitive globbing (used in pathname expansion) and matching
```
akinomyoga, line 24:
```suggestion
# Case-sensitive globbing (used in pathname expansion) and matching
```

akinomyoga:
> Oh, the original author, as I read it, was @saul365, not @nextia365.

Yeah, that was the tricky point. I think @saul365 and @nextia365 are actually the accounts of the same person. The *PR submitter* is @saul365, but the email address of the commit author points @nextia365 while the author name is `saul365`. Anyway, I wouldn't try to rewrite the commit author, and GitHub will finally relate it to the account @nextia365.

> Since there wasn't a signed-off-by line in the original commit, I wasn't going to add one, but by the time I was done reworking it and squashing the commits,

As far as I understand, the "Signed-off-by" line is used to declare that the commit is made by the own work of the commit author or at least the commit does the correct thing for the copyright matter, e.g., under copyright assignments or licensing. It is subtle how we consider the license of the PR commits submitted on GitHub, but we usually consider that the submitter agrees to provide the piece of code with the same license as the project, the MIT license in the case of OMB, unless otherwise specified. The MIT license requires us to retain the original copyright notice, which means that the copyright of the submitted *lines* is still held by the original author and that information should be kept. In this case, we usually track the copyright of each line through the commit history, so the commit author or `Co-authored-by` line is important. If you want to include the "Signed-off-by" line, you need to be more careful about the copyright treatment because the line indicates that you certify that.

If you think your change is significant, you wouldn't have had to squash the commits. You could include multiple commits made by different people in one PR. 

----

Thanks for all the other updates! I left some suggestions for the code comments. Could you check them? Other than that there are no further points to discuss, I guess. Thank you.

lib/shopt.sh
jjmcdn: > It is confusing to only mention the behavior before applying the setting.

Sure, that makes sense to me.

jjmcdn:
> Thanks for all the other updates! I left some suggestions for the code comments. Could you check them? Other than that there are no further points to discuss, I guess. Thank you.

Thank you for the detailed responses and explanations of the project's workflow and policies.  I think I've merged all the suggested changes and I've update the PR.

akinomyoga:
Thank you! Hmm, I guess you forgot to push the commit? Nothing seems to be pushed since the last review.

jjmcdn:
> Thank you! Hmm, I guess you forgot to push the commit? Nothing seems to have been pushed since the last review.

Oh, yeah, I guess that hadn't gotten pushed for some reason, I don't know what happened there, but I just did another force push and confirmed that my current changes (based on the suggestions you'd made above) are the current head commit.  Sorry about that!

akinomyoga (approved):
Thanks!

akinomyoga:
I think I'll merge it right now.

Issue #365, opened 2022-09-29
Integrate some CI for shellcheck into the project


akinomyoga:
_Originally posted by @nntoan in https://github.com/ohmybash/oh-my-bash/issues/355#issuecomment-1256965559_

        @akinomyoga [...] I also want to integrate some CI for shellcheck into the project, that's one of reasons I turned it on.

Issue #83, opened 2019-07-24
REQUEST: Adapt spellcheck for merge requests


Kreyren:
All bash/shell scripts should pass spellcheck (https://github.com/koalaman/shellcheck) and there are so many issues missing quotation, unexpected variable manipulation, etc.. 

Requsting to adapt spellcheck to prevent these issues in the future and to force quality assurance (spellcheck is free for open-source projects).

Kreyren:
@nntoan Requesting approval.

nntoan:
Hi @Kreyren,

Actually, the first time when I created this project I also adapt `shellcheck` into every pushes/pull request through Codacy.

However, there's inactive from me (no PRs merged) for such a long time and Codacy has set all projects on Codacy into the inactive state.

![Screenshot 2019-08-14 11 03 57](https://user-images.githubusercontent.com/3891429/62993955-ca90ef00-be83-11e9-8afb-a5d32fe75dda.png)

I've opened a new ticket on Codacy ask them to re-activate OhMyBash project. And I will update the status ASAP here.

All best,

Kreyren:
Added shellcheck for ohmybash/oh-my-bash

@nntoan Check your access and verify access to other members.

Kreyren:
Assigned P1 since code quality of this repository is poor and it should be prioritized.

Kreyren:
Configuration of shellcheck is needed, investigating atm

nntoan:
@Kreyren I've sent an invitation to our Codacy project to the email address on your GitHub profile. Please let me know if you need an invitation for other email address.

Thanks,
Toan

Kreyren:
@nntoan i make some configuration already on my end, recommends your codacy to be closed and i transfer ownership of the one that i've configured

Also it's beeing discussed on Codacy some standartization for shellcheck since default configuration is not sufficient and configuring it is pita cherrypicking.

PR #402, opened 2023-02-20
Apply ShellCheck to a few files in the `tools/` directory

neuroevolutus:
This is a PR addressing ShellCheck suggestions for a few files in the `tools/` directory of this project. As there is some existing interest in improving code with ShellCheck (#365, #83), I thought it would be helpful to begin applying some of these suggestions. @akinomyoga, if you are able to review my work to check for improvements/mistakes, I would be extremely grateful!


tools/git-prompt.sh
```diff
@@ -109,7 +109,8 @@ __git_ps1_show_upstream ()
 
 	svn_remote=()
 	# get some config options from git-config
-	local output="$(git config -z --get-regexp '^(svn-remote\..*\.url|bash\.showupstream)$' 2>/dev/null | tr '\0\n' '\n ')"
+	local output
+	output="$(git config -z --get-regexp '^(svn-remote\..*\.url|bash\.showupstream)$' 2>/dev/null | tr '\0\n' '\n ')"
```
akinomyoga, line 113: This file is taken from the upstream git ([`/contrib/completion/git-prompt.sh`](https://github.com/git/git/blob/d9d677b2d8cc5f70499db04e633ba7a400f64cbf/contrib/completion/git-prompt.sh)). We wouldn't like to introduce unneeded diffs so that it makes it easier to later merge the new version from the upstream.

By the way, we do not use the exit status of this line, and also this is the configuration for the interactive session. so `set -e` is irrelevant. For these reasons, we do not need to separate these two lines.

tools/install.sh
```diff
@@ -6,15 +6,16 @@ if [ -z "${BASH_VERSION-}" ]; then
   printf "Error: Bash 3.2 or higher is required for Oh My Bash.\n"
   printf "Error: Install Bash and try running this installation script with Bash.\n"
   if command -v bash >/dev/null 2>&1; then
+    # shellcheck disable=SC2016
     printf 'Example: \033[31;1mbash\033[0;34m -c "$(curl -fsSL https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/tools/install.sh)"\n'
   fi
-  return 1 >/dev/null 2>&1 || exit 1
+  return 1 >/dev/null 2>&1
```
akinomyoga, line 12: This is the idiom when the file can be used as either of `source tools/install.sh` or `bash tools/install.sh`.

tools/install.sh
```diff
@@ -65,15 +67,15 @@ function _omb_install_readargs {
           install_prefix=$arg
         else
           install_opts+=:error
-          printf 'install (oh-my-bash): %s\n' "$RED$BOLD[Error]$NORMAL ${RED}the option argument for '--prefix' is empty.$NORMAL" >&2
+          printf 'install (oh-my-bash): %s\n' "$RED${BOLD}[Error]$NORMAL ${RED}the option argument for '--prefix' is empty.$NORMAL" >&2
```
akinomyoga, line 70: This is a workaround for Zsh, but this script is intended to be a Bash script, and there is already a check for the Bash version.

tools/install.sh
```diff
@@ -113,28 +115,33 @@ function _omb_install_run {
...
+  local BANNER
+  BANNER=$(cat <<-'EOF'
+	         __                          __               __  
+	  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_ 
+	 / __ \/ __ \   / __ `__ \/ / / /  / __ \/ __ `/ ___/ __ \
+	/ /_/ / / / /  / / / / / / /_/ /  / /_/ / /_/ (__  ) / / /
+	\____/_/ /_/  /_/ /_/ /_/\__, /  /_.___/\__,_/____/_/ /_/ 
+	                        /____/                            .... is now installed!
+EOF
+  )
```
akinomyoga, line 127: We are using `printf '%s\n'` to avoid the fork cost. By the way, is this really reported by `shellcheck`?

tools/install.sh
```diff
@@ -113,28 +115,33 @@ function _omb_install_run {
...
 ##   @var[in] install_opts
 ##   @var[in] OSH
 function _omb_install_user_bashrc {
-  printf "${BLUE}Looking for an existing bash config...${NORMAL}\n"
+  printf '%s\n' "${BLUE}Looking for an existing bash config...${NORMAL}"
   if [[ -f ~/.bashrc || -h ~/.bashrc ]]; then
-    local bashrc_backup=~/.bashrc.omb-backup-$(date +%Y%m%d%H%M%S)
-    printf "${YELLOW}Found ~/.bashrc.${NORMAL} ${GREEN}Backing up to $bashrc_backup${NORMAL}\n"
+    local bashrc_backup
+    bashrc_backup=~/.bashrc.omb-backup-$(date +%Y%m%d%H%M%S)
```
akinomyoga, line 139: Here again we do not use the exit status, so there is no reason to put them in separate lines.

akinomyoga (commented):
Thank you for your contributions! I actually wanted to process all at once, but incremental changes are also welcome! I was thinking of merging as many PRs as possible before applying the style changes because the global style changes would introduce unnecessary conflicts in the PRs, but I believe `tools/` directory is safe to introduce the style changes. Thanks!

I think about a half of the suggested changes are good ones, but there are still some noises. I left comments. Could you check them?

tools/git-prompt.sh
akinomyoga: I mean I'd like to keep this file untouched because this file is not managed by this project.

neuroevolutus: Thanks, Koichi. Yeah, that definitely makes sense.

akinomyoga: Thanks!

tools/install.sh
neuroevolutus: Thanks for catching that. I've added an ignore line for this hint.

tools/install.sh
neuroevolutus: Thanks. I've made the change and have updated the commit message accordingly.

akinomyoga: Thanks for updating!

I actually would like to make `disable=SC1087` a global setting for the shellcheck. The construct `$var[xxx]` is interpreted as an array expansion only in Zsh. It should never be an array expansion in POSIX sh and all the other POSIX-like shells. I naively think it's not so confusing to the users, so I'd like to turn `SC1087` off globally if possible. It is cleaner than putting `disable=SC1087` on every line. What do you think?

tools/install.sh
neuroevolutus: Thanks for teaching me that! At first, I was confused because in the [Bash manual](https://www.gnu.org/software/bash/manual/bash.html#index-return-status) it was stated that `return` was a synonym for `exit`, so I thought that creating this compound command with the `OR` conditional was redundant. However, I tried doing `return 1` in my Bash shell to see what would happen, and it indeed said `-bash: return: can only `return' from a function or sourced script`. I was still befuddled, so I made a script that simply called `return 1` and invoked it with `bash` only to get this same error. When I changed the script to run `return 1 || exit 1` and sourced it again, it produced nothing on standard error and finished with exit code `1`. It seems rather quirky that `exit` and `return` would have different semantics despite being called synonymous, but this is definitely something I'm glad to have learned.

In any case, there should now be ShellCheck ignore lines in all places where this change should be applicable.

tools/install.sh
neuroevolutus: Thanks so much for mentioning the point about `fork`. I would not have thought to considered that at all. I've updated the code to use pure Bash following the advice I found [here](https://serverfault.com/a/72511), [here](https://serverfault.com/questions/72476/clean-way-to-write-complex-multi-line-string-to-a-variable#comment497435_72511) and [here](https://serverfault.com/questions/72476/clean-way-to-write-complex-multi-line-string-to-a-variable#comment579993_72511). However, this is approaching code that might be a bit too "clever," so I'd love to hear what you think.

neuroevolutus:
In the function `_omb_install_main`, I've altered the code to have a single `local` declaration for all of the function's variables so I could avoid having far more occurrences of the `# shellcheck disable=SC2155` ignore line.

tools/install.sh
akinomyoga: This is my personal preference, but I would like to avoid also the use of here documents if it's not necessary because the here documents actually save the contents to temporary files. The recent versions of Bash (Bash 5.1 and higher) started to create a pipe instead of a temporary file for a here document whose content is not larger than 64 KiB (or some other values depending on the operating system). But, the creation of pipe seems still redundant to me compared to the arguments to `printf '%s\n'`. If the reason to switch is just because of the false positive of shellcheck but not because it is better than `printf '%s\n'`, I wouldn't like to switch honestly. Do you have better reasoning?

tools/install.sh
akinomyoga: That section of the Bash manual defines the general terms used in the text but not in the code. The *return status* and the *exit status* do not represent `return <status>` and `exit <status>` in the code, but the words in the textual explanation like "the *return status* of a coprocess is ...." and "the *exit status* of a pipe line is ...".

> It seems rather quirky that `exit` and `return` would have different semantics despite being called synonymous,

So, they are not synonyms. What are the synonyms are the words *return status* and *exit status*.

> In any case, there should now be ShellCheck ignore lines in all places where this change should be applicable.

Hmm, my `shellcheck` doesn't seem to complain about `return 2 &>/dev/null || exit 2`. Maybe this is related to the version of `shellcheck`. My version is 0.8.0, but the latest version seems to be 0.9.0. I have checked the `shellcheck` behavior of the online editor at https://www.shellcheck.net/ which seems to complain about it.

I thought it might be worth reporting the issue, but I found an existing report for this false positive: https://github.com/koalaman/shellcheck/issues/2620.

Anyway, it is good to keep `disable=SC2317` for a while even if it would be fixed at the shellcheck's side.

akinomyoga:
> In the function `_omb_install_main`, I've altered the code to have a single `local` declaration for all of the function's variables so I could avoid having far more occurrences of the `# shellcheck disable=SC2155` ignore line.

Thanks, but I actually prefer the previous version. The variable scope should be kept small, and I wouldn't like to change the code in a worse direction just to suppress the warnings. Two separate lines of `disable=SC2155` would be just fine. It is understandable to declare `RED`, ..., and `NORMAL` before the `if` statement rather than declaring them in each branch, so the declaration in the previous version [`4212951c`](https://github.com/ohmybash/oh-my-bash/commit/4212951c2a8a58b77a5d9fff2c8282baf8bfa44f#diff-5dc310979be51acc76eb748ef379630b65614b5a4132fc5a870019ade043935bR210) is 👍.

tools/install.sh
neuroevolutus: Sure thing. I think that would be much simpler as well!

tools/install.sh
neuroevolutus: For sure, after learning that, I can definitely agree that using a heredoc would be overkill. One thing I came across, however, is that `# disable shellcheck=SC1003,SC2016` above the call to `printf` does not seem to silence the error. Given that, would it be appropriate to split up this block into separate `printf` command lines in order to both maintain visual consistency and be able to add a single ShellCheck ignore comment above the offending line containing backticks?

akinomyoga: Thank you for updating the commit!

> however, is that even putting `# disable shellcheck=SC1003,SC2016` above the call to `printf` does not seem to silence the error.

Isn't that a typo of `# shellcheck disable=SC1003,SC2016` (I mean `disable` and `shellcheck` are swapped). As far as I try in https://www.shellcheck.net/, it does seem to suppress the errors.

neuroevolutus: Oh dear, I have no idea how I made that mistake. Thanks so much for catching that! Updating the PR now...

neuroevolutus: Done!

akinomyoga (dismissed):
Thank you for your patience and all the updates! I think I'll later merge this.

akinomyoga:
I added a newline bc4018b9f67e2abd41951dfe83f30a6977d0b514.

akinomyoga:
Ah, I actually think we should move the declaration of `local ncolors` outside the `if` statement because otherwise, the later code unintendedly references the global variable `ncolors` when the `if` condition is not satisfied.

akinomyoga (approved):

akinomyoga:
Thanks. I've merged it.

PR #418, opened 2023-03-21
sudo plugin

vicmunoz:
Simple sudo plugin to add sudo at the beginning of the line or with last command executed if line is empty.

### DIFF
--- a/plugins/sudo/sudo.plugin.sh
+++ b/plugins/sudo/sudo.plugin.sh
@@ -1,0 +1,11 @@
+#! bash oh-my-bash.module
+
+function _omb_plugin_sudo_add_sudo {
+  if [[ -z $READLINE_LINE ]]; then
+    READLINE_LINE="$(fc -ln -1 | command sed 's/^[[:space:]]\{1,\}//')"
+  fi
+  READLINE_LINE="sudo $READLINE_LINE"
+  ((READLINE_POINT += 5))
+}
+
+bind -x '"\e\e": _omb_plugin_sudo_add_sudo'
--- a/plugins/sudo/sudo.plugin.sh
+++ b/plugins/sudo/sudo.plugin.sh
@@ -8,4 +8,6 @@ function _omb_plugin_sudo_add_sudo {
   ((READLINE_POINT += 5))
 }
 
-bind -x '"\e\e": _omb_plugin_sudo_add_sudo'
+bind -m emacs -x '"\e\e": _omb_plugin_sudo_add_sudo'
+bind -m vi-insert -x '"\e\e": _omb_plugin_sudo_add_sudo'
+bind -m vi-command -x '"\e\e": _omb_plugin_sudo_add_sudo'
--- a/plugins/sudo/README.md
+++ b/plugins/sudo/README.md
@@ -1,0 +1,50 @@
+# sudo
+
+Easily prefix your current or previous commands with `sudo` by pressing <kbd>esc</kbd> twice.
+
+To use it, add `sudo` to the plugins array in your bashrc file:
+
+```bash
+plugins=(... sudo)
+```
+
+## Usage
+
+### Current typed commands
+
+Say you have typed a long command and forgot to add `sudo` in front:
+
+```console
+$ apt-get install build-essential
+```
+
+By pressing the <kbd>esc</kbd> key twice, you will have the same command with `sudo` prefixed without typing:
+
+```console
+$ sudo apt-get install build-essential
+```
+
+### Previous executed commands
+
+Say you want to delete a system file and denied:
+
+```console
+$ rm some-system-file.txt
+-su: some-system-file.txt: Permission denied
+$
+```
+
+By pressing the <kbd>esc</kbd> key twice, you will have the same command with `sudo` prefixed without typing:
+
+```console
+$ rm some-system-file.txt
+-su: some-system-file.txt: Permission denied
+$ sudo rm some-system-file.txt
+Password:
+$
+```
+
+## Credits
+
+Inspired by ohmyzsh sudo plugin.
+


plugins/sudo/sudo.plugin.sh
```diff
@@ -0,0 +1,14 @@
+#! bash oh-my-bash.module
+
+add_sudo() {
+  local prefix suffix
+  prefix="${READLINE_LINE:0:$READLINE_POINT}"
+  suffix="${READLINE_LINE:$READLINE_POINT}"
```
akinomyoga, line 6: By the way, do we need to split `READLINE_LINE` into `prefix` and `suffix`?

plugins/sudo/sudo.plugin.sh
```diff
@@ -0,0 +1,14 @@
+#! bash oh-my-bash.module
+
+add_sudo() {
```
akinomyoga, line 3:
```suggestion
function _omb_plugin_sudo_add_sudo {
```

- We are gradually switching to the naming convention with the prefix `_omb_plugin_{plugin_name}_` not to clutter the user's command/function space.
- We prefer `function fn` to `fn()` to avoid unexpected alias expansions.

Can you also update the function name in `bind -x`?

plugins/sudo/sudo.plugin.sh
```diff
@@ -0,0 +1,14 @@
...
+  local prefix suffix
+  prefix="${READLINE_LINE:0:$READLINE_POINT}"
+  suffix="${READLINE_LINE:$READLINE_POINT}"
+  if [[ -z $prefix ]] && [[ -z $sufix ]]; then
```
akinomyoga, line 7:
```suggestion
  if [[ -z $prefix && -z $suffix ]]; then
```
- If we keep `prefix` and `suffix`, `sufix` is a typo of `suffix`.
- `&&` can be used in the conditional command.

plugins/sudo/sudo.plugin.sh
```diff
@@ -0,0 +1,14 @@
...
+    prefix="$(fc -ln -1 | awk '{$1=$1};1')"
```
akinomyoga, line 8: What does `awk '{$1=$1};1'` do? If you would like to just strip the beginning whitespaces, could you use `sed 's/^[[:space:]]\{1,\}//'`?

plugins/sudo/sudo.plugin.sh
```diff
@@ -0,0 +1,14 @@
...
+  fi
+  READLINE_LINE="sudo $prefix$suffix"
+  ((READLINE_POINT += 5))
+}
+
+bind -x '"\e\es": add_sudo'
```
akinomyoga, line 14: The upstream ohmyzsh sudo seems to set up the keybindings for all of the three keymaps (corresponding to Bash's `emacs`, `vi-insert`, and `vi-command`). With the above `bind -x`, we only set the keybinding for the currently selected keymap. Can we bind all of the three as ohmyzsh sudo does? What do you think?

akinomyoga (changes requested):

plugins/sudo/sudo.plugin.sh
vicmunoz: fixed in last commit

plugins/sudo/sudo.plugin.sh
vicmunoz: fixed in last commit

plugins/sudo/sudo.plugin.sh
vicmunoz: replaced awk with sed

akinomyoga: Thanks!

plugins/sudo/sudo.plugin.sh
akinomyoga: What do you think?

vicmunoz: removed prefix and suffix in last commit

plugins/sudo/sudo.plugin.sh
vicmunoz: I'm not familiar with vim or emacs keymap. To be honest, I'm not a used user of oh-my-bash. I needed this functionality for bash and first look was this repo, due it was not present, I thinked would be a good idea to send you a first version. If it's suitable for the framework, happy to help

vicmunoz: I'm not familiar with vim or emacs keymap. To be honest, I'm not a used user of oh-my-bash. I needed this functionality for bash and first look was this repo, due it was not present, I thinked would be a good idea to send you a first version. If it's suitable for the framework, happy to help

akinomyoga: If the reason that you would request the feature is to use a Bash version of ohmyzsh/sudo, I would expect it would work in the same way as ohmyzsh/sudo. Anyway, if the plugin has the same name as in that of ohmyzsh, I believe we should make the behavior similar to the original version.

plugins/sudo/sudo.plugin.sh
```diff
@@ -0,0 +1,11 @@
...
+
+function _omb_plugin_sudo_add_sudo {
+  if [[ -z $READLINE_LINE ]]; then
+    READLINE_LINE="$(fc -ln -1 | command sed 's/^[[:space:]]\{1,\}//')"
+  fi
+  READLINE_LINE="sudo $READLINE_LINE"
+  ((READLINE_POINT += 5))
+}
+
+bind -x '"\e\e": _omb_plugin_sudo_add_sudo'
```
akinomyoga, line 11: What I discuss above is the three keybindings in [ohmyzsh sudo plugin L106-108](https://github.com/ohmyzsh/ohmyzsh/blob/d47e1d65f66f9bb2e7a96ba58797b33f0e91a623/plugins/sudo/sudo.plugin.zsh#L106-L108). What I suggest is the following change:

```suggestion
bind -m emacs -x '"\e\e": _omb_plugin_sudo_add_sudo'
bind -m vi-insert -x '"\e\e": _omb_plugin_sudo_add_sudo'
bind -m vi-command -x '"\e\e": _omb_plugin_sudo_add_sudo'
```

akinomyoga:
squashed

akinomyoga (dismissed):

akinomyoga (approved):

PR #423, opened 2023-03-30
Update README.md

biereluc:
Add a way to easy modify theme more quikly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,13 @@ OSH_THEME="agnoster" # (this is one of the fancy ones)
 # see https://github.com/ohmybash/oh-my-bash/wiki/Themes#agnoster
 ```
 
+To modify theme automatically, you can use this sed command to edit OSH_THEME.
+
+```shell
+theme="powerline"
+sed -i "s/\(^OSH_THEME=\"\)\([^\"\\n]*\)\(\".*$\)/\1$theme\3/g" ~/.bashrc
+```
+
 Open up a new terminal window and your prompt should look something like this:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +97,13 @@ OSH_THEME="agnoster" # (this is one of the fancy ones)
 # see https://github.com/ohmybash/oh-my-bash/wiki/Themes#agnoster
 ```
 
+To modify theme automatically, you can use this sed command to edit OSH_THEME.
+
+```shell
+theme="powerline"
+sed -i "s/\(^OSH_THEME=\"\)\([^\"\\n]*\)\(\".*$\)/\1$theme\3/g" ~/.bashrc
```
akinomyoga, line 104: - This usage of `-i` is only supported by GNU sed.
- `g` doesn't have an effect because we are matching using anchor `^` for the beginning of the line.
- This doesn't work when the right-hand side is not enclosed within double quotations.
- I see `\n` in the bracket expression, but newline characters do not appear with this usage because the matching is performed for each line.

akinomyoga (commented):
Thank you for opening the PR!

I'm not sure if it's worth including it in README. There are a few caveats raised below. Some of these can confuse users.

Also, the command line seems too complicated and can break the user's bashrc when the user makes a mistake. It is safer to use it only when the user understands what this does. But then, I guess the user can construct the user's own command line to achieve this.

For general users, it seems easier to directly edit the file.

akinomyoga:
Thank you anyway for opening the PR!

PR #427, opened 2023-04-03
Removing redundant HISTIGNORE patterns

The-Judge:
OMB currently sets in `lib/history.sh`:

```
# Avoid duplicate entries
HISTCONTROL="erasedups:ignoreboth"

# Don't record some commands
export HISTIGNORE="&:[ ]*:exit:ls:bg:fg:history:clear"
```

According to [official GNU bash manual](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-HISTIGNORE), having `&:[ ]*` in `HISTIGNORE` provides the functionality of `ignoreboth`. Thus, having both provides redundant results but makes the config harder to understand.

### DIFF
--- a/lib/history.sh
+++ b/lib/history.sh
@@ -37,7 +37,7 @@ export HISTFILESIZE=
 HISTCONTROL="erasedups:ignoreboth"
 
 # Don't record some commands
-export HISTIGNORE="&:[ ]*:exit:ls:bg:fg:history:clear"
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):

akinomyoga:
Thanks! All of the three PRs are good ones.

PR #432, opened 2023-04-07
fix: Quoting

### LINKED ISSUE #242: Issue with $HOME directory
Recently I got a fresh OMB installation on Git Bash. But every time I launch Terminal , I keep getting message like this.

`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected`
`bash: [: /c/Users/Ibne: binary operator expected` 

My `$HOME` directory is `/c/Users/Ibne Nahian` in which `Ibne Nahian` folder has a space in its name. But OMB detecting `/c/Users/Ibne` as `$HOME` which is not expected. Can you guys take care of this issue ?

hyperupcall:
Improve quoting in various areas.

I also had the chance to reproduce #242, but it seems like the issue has been fixed since then. So closes #242

### DIFF
--- a/completions/django.completion.sh
+++ b/completions/django.completion.sh
@@ -32,42 +32,43 @@
 #
 # To uninstall, just remove the line from your .bash_profile and .bashrc.
 
-_django_completion()
-{
-    COMPREPLY=( $( COMP_WORDS="${COMP_WORDS[*]}" \
-                   COMP_CWORD=$COMP_CWORD \
-	               DJANGO_AUTO_COMPLETE=1 $1 ) )
+function _omb_completion_django {
+  COMPREPLY=($(COMP_WORDS="${COMP_WORDS[*]}" \
+    COMP_CWORD=$COMP_CWORD \
+	  DJANGO_AUTO_COMPLETE=1 "$1"))
 }
-complete -F _django_completion -o default django-admin.py manage.py django-admin
+complete -F _omb_completion_django -o default django-admin.py manage.py django-admin
 
-_python_django_completion()
-{
-    if [[ ${COMP_CWORD} -ge 2 ]]; then
-        PYTHON_EXE=$( basename -- "${COMP_WORDS[0]}" )
-        echo $PYTHON_EXE | command grep -E "python([2-9]\.[0-9])?" >/dev/null 2>&1
-        if [[ $? == 0 ]]; then
-            PYTHON_SCRIPT=$( basename -- "${COMP_WORDS[1]}" )
-            echo $PYTHON_SCRIPT | command grep -E "manage\.py|django-admin(\.py)?" >/dev/null 2>&1
-            if [[ $? == 0 ]]; then
-                COMPREPLY=( $( COMP_WORDS="${COMP_WORDS[*]:1}" \
-                               COMP_CWORD=$(( COMP_CWORD-1 )) \
-                               DJANGO_AUTO_COMPLETE=1 ${COMP_WORDS[*]} ) )
-            fi
-        fi
+function _omb_completion_django_python {
+  if ((COMP_CWORD >= 2)); then
+    local PYTHON_EXE=$(basename -- "${COMP_WORDS[0]}")
+    if command grep -E "python([2-9]\.[0-9])?" >/dev/null 2>&1 <<< "$PYTHON_EXE"; then
+      local PYTHON_SCRIPT=$(basename -- "${COMP_WORDS[1]}")
+      if command grep -E "manage\.py|django-admin(\.py)?" >/dev/null 2>&1 <<< "$PYTHON_SCRIPT"; then
+        COMPREPLY=($(COMP_WORDS="${COMP_WORDS[*]:1}" \
+          COMP_CWORD=$((COMP_CWORD - 1)) \
+          DJANGO_AUTO_COMPLETE=1 "${COMP_WORDS[@]}"))
+      fi
     fi
+  fi
 }
 
-# Support for multiple interpreters.
-unset -v pythons
-if _omb_util_command_exists whereis; then
+function _omb_completion_django_init {
+  # Support for multiple interpreters.
+  local -a pythons=()
+  if _omb_util_command_exists whereis; then
+    local python_interpreters
     _omb_util_split python_interpreters "$(whereis python | cut -d " " -f 2-)"
-    pythons=()
+    local python
     for python in "${python_interpreters[@]}"; do
-        pythons+=("$(basename -- "$python")")
+      pythons+=("$(basename -- "$python")")
     done
-    pythons=($(printf '%s\n' "${pythons[@]}" | sort -u))
-else
+    _omb_util_split pythons "$(printf '%s\n' "${pythons[@]}" | sort -u)" $'\n'
+  else
     pythons=(python)
-fi
+  fi
 
-complete -F _python_django_completion -o default "${pythons[@]}"
+  complete -F _omb_completion_django_python -o default "${pythons[@]}"
+  unset -f "$FUNCNAME"
+}
+_omb_completion_django_init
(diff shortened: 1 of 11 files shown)


completions/django.completion.sh
```diff
@@ -58,16 +58,16 @@ _python_django_completion()
...
+        pythons="${pythons} $(basename -- "$python")"
     done
-    pythons=$(echo $pythons | tr " " "\n" | sort -u | tr "\n" " ")
+    pythons=$(echo "$pythons" | tr " " "\n" | sort -u | tr "\n" " ")
 else
     pythons=python
 fi
 
-complete -F _python_django_completion -o default $pythons
+complete -F _python_django_completion -o default "$pythons"
```
akinomyoga, line 72: I believe we originally expected word splitting here. Another approach is to use an array, but we need to anyway apply word splitting to the result of `whereis python` at some point at least once.

Also, for completeness, we actually need to care about `IFS` and possible pathname expansions. Maybe I could consider adding a utility function `_omb_util_split ARRAY_NAME TEXT` in `lib/utils.sh` to perform the safe splitting.

lib/omb-prompt-base.sh
```diff
@@ -341,15 +341,15 @@ function svn_prompt_vars {
 # - .hg is located in ~/Projects/Foo/.hg
 # - get_hg_root starts at ~/Projects/Foo/Bar and sees that there is no .hg directory, so then it goes into ~/Projects/Foo
 function get_hg_root {
-    local CURRENT_DIR=$(pwd)
+    local CURRENT_DIR="$PWD"
```
akinomyoga, line 344: Quoting the parameter expansions in the right-hand sides of variable assignments are actually not needed. I'd like to normalize this in the codebase at some point, but maybe it is good to remove the unnecessary quoting in this chance.
```suggestion
    local CURRENT_DIR=$PWD
```

themes/agnoster/agnoster.theme.sh
```diff
@@ -104,10 +104,10 @@
 
 # then add the following to your .bashrc:
 
-# export THEME=$HOME/.bash/themes/agnoster-bash/agnoster.bash
-# if [[ -f $THEME ]]; then
+# export THEME="$HOME/.bash/themes/agnoster-bash/agnoster.bash"
+# if [[ -f "$THEME" ]]; then
```
akinomyoga, line 108: One of the ideas to use the conditional commands `[[ ... ]]` is to make use of the fact that the arguments doesn't undergo word splitting and pathname expansions. So I would like to normalize the codebase so that we avoid the quoting of the arguments in the conditional commands (except for the right-hand side of `==`, `!=`, and `=~`, where the quoting takes effect).

```suggestion
# if [[ -f $THEME ]]; then
```

tools/git-prompt.sh
```diff
@@ -255,7 +255,7 @@ __git_ps1_colorize_gitstring ()
 	local flags_color="$c_lblue"
 
 	local branch_color=""
```
akinomyoga, line 257: Actually, this file `tools/git-prompt.sh` is just a copy of the upstream [git/git `contrib/completion/git-prompt.sh`](https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh), so I would like to avoid edit this file independently from the upstream version.

Hmm, these seem not to have yet been fixed in the upstream either.

- https://github.com/git/git/blob/0607f793cbe0af16aee6d2480056d891835884bd/contrib/completion/git-prompt.sh#L272
- https://github.com/git/git/blob/0607f793cbe0af16aee6d2480056d891835884bd/contrib/completion/git-prompt.sh#L301
- https://github.com/git/git/blob/0607f793cbe0af16aee6d2480056d891835884bd/contrib/completion/git-prompt.sh#L499

akinomyoga (commented):
Thank you! I have comments as follows.

akinomyoga:
I was also searching the upstreams

- https://github.com/django/django/blob/main/extras/django_bash_completion
- [archived] https://github.com/iljaweis/vault-bash-completion
- [archived] https://github.com/kura/vagrant-bash-completion
- https://github.com/github/hub/blob/master/etc/hub.bash_completion.sh
- https://github.com/speedenator/agnoster-bash/blob/master/agnoster.bash#L40

akinomyoga:
Updated commit messages for Conventional Commits.

lib/omb-prompt-base.sh
hyperupcall: Thank you for the correction - I prefer the quote-less approach as well

tools/git-prompt.sh
hyperupcall: Gotchia :+1: Yeah it would be better to update upstream

completions/django.completion.sh
hyperupcall: That does appear to fix the word splitting issue

hyperupcall:
I'm not sure what other improvements I could add, but it looks good to me. It's probably not as productive to fix the completions because it is better to do so upstream - it would be nice if these things could be autofixed, but I don't think there currently exists a tool to do that

tools/git-prompt.sh
akinomyoga: I have created patches for the upstream Git.

https://github.com/akinomyoga/git/commit/ce955e4f0293214aa8171a789bec9f073f7913e2
https://github.com/akinomyoga/git/commit/a7799fb5a3d65cb48db345e37c8f87a62e2c7fba

I haven't included the change `(* | default)` -> `(default | *)` in the case statement.

To submit patches to git, we need to include the `Signed-off-by` line for the contributed people in the commit message.

```
Signed-off-by: Edwin Kofler <edwin@kofler.dev>
```

@hyperupcall Can you agree on submitting the patches to the upstream git and adding the above line in the commit messages of the patches? The comments to the patches are also welcome.

hyperupcall: Yes! I will submit these patches to upstream Git and add the `Signed-off-by: Edwin Kofler <edwin@kofler.dev>` line

The `*|default` thing only fixes a ShellCheck issue, so I don't see any strong reason to include it in the patches

akinomyoga: Thank you for your reply! I have updated the patches.

https://github.com/akinomyoga/git/commit/e80559019501ba8f4b942e92257fea84baa69f8c
https://github.com/akinomyoga/git/commit/2da42103fbe55aed548a543d1bf4aeace8179b71

Ah, I intended to submit the patches to `git@vger.kernel.org` but will you submit the patches?

Anyway, I first submit them to myself (with Cc to you) for testing purposes.

hyperupcall: Ohh, you meant that you were going to submit the patches, I see what you mean now :sweat_smile: 

It might be better for you to submit them since it seems you were already going to do so? I don't mind either way, but it seems more right to go with what you originally planned

hyperupcall: Also I got your emails, I'll check the patch contents

hyperupcall: The patches look good to me, and I applied them fine

akinomyoga: Thanks!

akinomyoga: I have submitted them.

akinomyoga (approved):

PR #434, opened 2023-04-09
plugins/sudo: Added toggle support

dheisom:
Hi!
In this PR I've created the functionality of toggling `sudo` and fixed a bug that when you press `<ESC>` four times or more it was just adding more and more `sudo` commands to the begin.

### DIFF
--- a/plugins/sudo/sudo.plugin.sh
+++ b/plugins/sudo/sudo.plugin.sh
@@ -1,21 +1,23 @@
 #! bash oh-my-bash.module
 
+function _omb_plugin_sudo__is_inserted {
+  [[ "${READLINE_LINE::5}" = "sudo " ]]
+}
+
 function _omb_plugin_sudo_add_sudo {
-  function __omb_plugin_sudo_is_inserted {
-    [[ "${READLINE_LINE::4}" = "sudo" ]]; return $?
-  }
   if [[ -z $READLINE_LINE ]]; then
     READLINE_LINE="$(fc -ln -1 | command sed 's/^[[:space:]]\{1,\}//')"
-    ((READLINE_POINT += ${#READLINE_LINE}))
-    __omb_plugin_sudo_is_inserted && return
+    READLINE_POINT=${#READLINE_LINE}
+    _omb_plugin_sudo__is_inserted && return
   fi
-  if __omb_plugin_sudo_is_inserted; then
-    READLINE_LINE="${READLINE_LINE//sudo /}"
+  if _omb_plugin_sudo__is_inserted; then
+    READLINE_LINE="${READLINE_LINE#sudo }"
     ((READLINE_POINT -= 5))
   else
     READLINE_LINE="sudo $READLINE_LINE"
     ((READLINE_POINT += 5))
   fi
+  ((READLINE_POINT < 0)) && READLINE_POINT=0
 }
 
 bind -m emacs -x '"\e\e": _omb_plugin_sudo_add_sudo'
--- a/plugins/sudo/sudo.plugin.sh
+++ b/plugins/sudo/sudo.plugin.sh
@@ -1,7 +1,7 @@
 #! bash oh-my-bash.module
 
 function _omb_plugin_sudo__is_inserted {
-  [[ "${READLINE_LINE::5}" = "sudo " ]]
+  [[ "$READLINE_LINE" = 'sudo '* ]]
 }
 
 function _omb_plugin_sudo_add_sudo {
--- a/plugins/sudo/sudo.plugin.sh
+++ b/plugins/sudo/sudo.plugin.sh
@@ -1,11 +1,21 @@
 #! bash oh-my-bash.module
 
 function _omb_plugin_sudo_add_sudo {
+  function __omb_plugin_sudo_is_inserted {
+    [[ "${READLINE_LINE::4}" = "sudo" ]]; return $?
+  }
   if [[ -z $READLINE_LINE ]]; then
     READLINE_LINE="$(fc -ln -1 | command sed 's/^[[:space:]]\{1,\}//')"
+    ((READLINE_POINT += ${#READLINE_LINE}))
+    __omb_plugin_sudo_is_inserted && return
   fi
-  READLINE_LINE="sudo $READLINE_LINE"
-  ((READLINE_POINT += 5))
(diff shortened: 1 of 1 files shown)


plugins/sudo/sudo.plugin.sh
```diff
@@ -1,11 +1,21 @@
 #! bash oh-my-bash.module
 
 function _omb_plugin_sudo_add_sudo {
+  function __omb_plugin_sudo_is_inserted {
```
akinomyoga, line 4: Can you put the function outside? Also, can you rename it as `_omb_plugin_sudo_is_inserted`, or `_omb_plugin_sudo__is_inserted` if you would like to explicitly make it *private*. cf #280

plugins/sudo/sudo.plugin.sh
```diff
@@ -1,11 +1,21 @@
...
+    [[ "${READLINE_LINE::4}" = "sudo" ]]; return $?
```
akinomyoga, line 5:
```suggestion
    [[ $READLINE_LINE == sudo* ]]
```

- There is no need to explicitly write `return $?`. The exit status of the function is that of the last command executed in the function.
- Also, we can actually use the glob pattern matching here.
- Another thing I thought is that maybe there is a command name like `sudoxxxx` in which case we might still prefix `sudo ` when <kbd>ESC ESC</kbd> is pressed. But probably we do not need to prefix `sudo ` for such a command because such a command would probably call `sudo` inside like `sudoedit`. But I'm not sure if we should switch the behavior depending on the detailed context or knowledge about the commands because it will make it cognitively harder for humans to predict what happens next by pressing each key.

plugins/sudo/sudo.plugin.sh
```diff
@@ -1,11 +1,21 @@
...
+  }
   if [[ -z $READLINE_LINE ]]; then
     READLINE_LINE="$(fc -ln -1 | command sed 's/^[[:space:]]\{1,\}//')"
+    ((READLINE_POINT += ${#READLINE_LINE}))
```
akinomyoga, line 9: I believe we can just assign the length without performing the arithmetic evaluation.

```suggestion
    READLINE_POINT=${#READLINE_LINE}
```

plugins/sudo/sudo.plugin.sh
```diff
@@ -1,11 +1,21 @@
...
+    __omb_plugin_sudo_is_inserted && return
+  fi
+  if __omb_plugin_sudo_is_inserted; then
+    READLINE_LINE="${READLINE_LINE//sudo /}"
+    ((READLINE_POINT -= 5))
```
akinomyoga, line 14: This can make the cursor position negative. I think Bash handles it properly, but it is good to ensure it is non-negative.

```suggestion
    ((READLINE_POINT -= 5))
    ((READLINE_POINT >= 0)) || READLINE_POINT=0
```

plugins/sudo/sudo.plugin.sh
```diff
@@ -1,11 +1,21 @@
...
+  function __omb_plugin_sudo_is_inserted {
+    [[ "${READLINE_LINE::4}" = "sudo" ]]; return $?
+  }
   if [[ -z $READLINE_LINE ]]; then
     READLINE_LINE="$(fc -ln -1 | command sed 's/^[[:space:]]\{1,\}//')"
+    ((READLINE_POINT += ${#READLINE_LINE}))
+    __omb_plugin_sudo_is_inserted && return
+  fi
+  if __omb_plugin_sudo_is_inserted; then
+    READLINE_LINE="${READLINE_LINE//sudo /}"
```
akinomyoga, line 13: This will remove every substring `sudo ` in any part of the command line.

```suggestion
    READLINE_LINE="${READLINE_LINE/#sudo }"
```

Also, when the command line looks e.g. `sudoedit file` and when `_is_inserted` succeeds, this fails to remove the prefix `sudo ` (because the space is not present) and just moves the cursor backward by 5 characters.

There are two options. One is to make `_is_inserted` to detect `sudo ` exactly (including the space). Another is to add an additional test to determine whether to remove `sudo `.

akinomyoga (commented):
Thank you for opening the PR! I quickly checked the suggested change and left comments. Could you see these comments and, if you agree, apply changes?

plugins/sudo/sudo.plugin.sh
dheisom: Ok, I'll rename, I think that if it keep there it can decrease the speed too

plugins/sudo/sudo.plugin.sh
dheisom: I don't like to use glob pattern matching, to fix the problem presented I can get the 5th character and check if it's a space.

Would I use glob pattern matching anyway? If yes, it will still with that problem

plugins/sudo/sudo.plugin.sh
dheisom: Ok, I was thinking that it could have some value from the shell character

plugins/sudo/sudo.plugin.sh
akinomyoga: Is there a reason to put the function inside the function?

Bash doesn't support local functions. If one defines the function inside the function, the function will be defined dynamically in the global space and remains after the outer function is returned. With the current implementation, the function `__omb_plugin_sudo_is_inserted ` is newly defined in the global space overwriting the previous one every time the function `_omb_plugin_sudo_add_sudo` is called.

plugins/sudo/sudo.plugin.sh
akinomyoga: > Would I use glob pattern matching anyway? If yes, it will still with that problem

The glob pattern can include a space.

```suggestion
    [[ $READLINE_LINE == 'sudo '* ]]
```

plugins/sudo/sudo.plugin.sh
dheisom: It was just to looks like a local function, just trash of my head

plugins/sudo/sudo.plugin.sh
dheisom: Ok, I'll make use of glob pattern

dheisom:
I think it is read

akinomyoga (approved):

akinomyoga:
Thanks

PR #436, opened 2023-04-15
Add new Powerline Light theme

imran-uk:
The regular Powerline theme is good... if you have a dark background in your Terminal ;)

If you use a light background such as "black on light yellow" or Solarized Light, then this theme is for you.

Tweak are done in themes/powerline-light/powerline-light.theme.sh used this colour chart as a reference: https://github.com/gawin/bash-colors-256


akinomyoga:

Could you make the image smaller? The image will be used in [themes/THEMES.md](https://github.com/ohmybash/oh-my-bash/blob/master/themes/THEMES.md) and [the wiki page](https://github.com/ohmybash/oh-my-bash/wiki/Themes), where images typically have a size of around 700x290. Also, please update `themes/THEMES.md`.

imran-uk:
> Thank you for your contribution!
> 
> Could you make the image smaller? The image will be used in [themes/THEMES.md](https://github.com/ohmybash/oh-my-bash/blob/master/themes/THEMES.md) and [the wiki page](https://github.com/ohmybash/oh-my-bash/wiki/Themes), where images typically have a size of around 700x290. Also, please update `themes/THEMES.md`.

Thanks for quick reply - done. Please let me know if OK.

akinomyoga:
OK

akinomyoga (approved):

akinomyoga (changes requested):
A powerline-based theme is supposed to source `powerline/powerline.base.sh` and then define its own functions. The base.sh of the new theme seems to be the same as `powerline/powerline.base.sh`, so I have removed the file and changed `powerline-light.theme.sh` to directly source `powerline/powerline.base.sh` (f6b1ea5b09cbb87222b171ae7495b7d0a630861d).

I now tried the theme. It seems the color of error exit statuses seems to be now dark cyan. However, we usually expect warning colors for failures. Could you update the color of the error exit statuses?

imran-uk:
Error exit status now a shade of red/orange (which is much more logical).

Screenshot now a bit larger, previously a bit too small.

akinomyoga:
> Error exit status now a shade of red/orange (which is much more logical).

Thank you!

> Screenshot now a bit larger, previously a bit too small.

Now the image is too large; the height is 373px. Could you prepare a smaller image of the height around 290px? You can crop the image instead of reducing the resolution. The display background and the terminal frame are irrelevant to the OMB theme.

imran-uk:
Okay, I think it should be ready now.

akinomyoga (approved):

PR #439, opened 2023-05-08
feat(git.plugin): adding alias to open predefined difftool

tal-lederman-dy:
Useful for visual side by side diff VIA external difftool. Here is an example for meld: https://stackoverflow.com/questions/2006032/view-differences-of-branches-with-meld


akinomyoga (approved):

akinomyoga:
Thank you for your contribution! I have merged it.

PR #443, opened 2023-05-23
omb-prompt-base: do not attempt to process git vars outside a git tree

jjmcdn:
I noticed this when checking out the n0qorg, being outside a git repository would have an undecorated `detached:` (from `SCM_THEME_DETACHED_PREFIX`) in the prompt.  It's easy enough to correct this in the theme but since the problem also may be observed in hawaii50, iterate, pro, standard and zitron as well, and `git rev-parse` is fast to execute anyway, this seems to be the correct place to insert this quick check.

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -237,6 +237,7 @@ function git_status_summary {
 function git_prompt_vars {
   local details=''
   local git_status_flags=''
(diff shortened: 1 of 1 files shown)


akinomyoga (approved):

akinomyoga:
Thanks!

PR #444, opened 2023-05-24
Setting to disable consideration repos with untracked files as dirty

awerebea:
Add `SCM_GIT_DISABLE_UNTRACKED_DIRTY="true"` setting for `lib/omb-prompt-base.sh` to disable consideration repos with untracked files as dirty.

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -30,6 +30,7 @@ THEME_BATTERY_PERCENTAGE_CHECK=${THEME_BATTERY_PERCENTAGE_CHECK:=true}
 
 SCM_GIT_SHOW_DETAILS=${SCM_GIT_SHOW_DETAILS:=true}
 SCM_GIT_SHOW_REMOTE_INFO=${SCM_GIT_SHOW_REMOTE_INFO:=auto}
+SCM_GIT_DISABLE_UNTRACKED_DIRTY=${SCM_GIT_DISABLE_UNTRACKED_DIRTY:=false}
 SCM_GIT_IGNORE_UNTRACKED=${SCM_GIT_IGNORE_UNTRACKED:=false}
 SCM_GIT_SHOW_CURRENT_USER=${SCM_GIT_SHOW_CURRENT_USER:=false}
 SCM_GIT_SHOW_MINIMAL_INFO=${SCM_GIT_SHOW_MINIMAL_INFO:=false}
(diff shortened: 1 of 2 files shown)


akinomyoga:
Thank you for your contribution! But in my understanding, `DISABLE_UNTRACKED_FILES_DIRTY` is the configuration for `lib/git.sh`. The behavior of `git_prompt_vars` in `lib/omb-prompt-base.sh` is controlled by the variable `SCM_GIT_IGNORE_UNTRACKED`. You can set `SCM_GIT_IGNORE_UNTRACKED=true`.

awerebea:
I see your point, but I don't want to ignore untracked files completely. I want to see the count of untracked files. I just want to disable considering the presence of untracked files as the dirty state of the repo. As I understand from the comment above this option in the default .bashrc example, this option should be responsible exactly for this behavior. Isn't it?

akinomyoga:
OK, so you specifically want to show the count of the untracked files yet don't want to mark it the dirty state under the presence of the untracked files. I now checked the comment above the option `DISABLE_UNTRACKED_FILES_DIRTY`, but it doesn't read that specific. Rather, the purpose seems to be to speed up the check. Checking the number of untracked files takes as long time (or more) as the dirty check for the untracked files, which contradicts the speedup.

https://github.com/ohmybash/oh-my-bash/blob/e91797d914775853deab2eb0c6ba59203cfad78a/templates/bashrc.osh-template#L38-L43

awerebea:
> Rather, the purpose seems to be to speed up the check.

Well, I don't know what the real purpose of this option is. But according to my observations, it does nothing. It does not change in any way the display of the "purity" of the repository in the presence of untracked files. It's marked as dirty, whether the value is true or whatever.
I'm interested in the first part of the statement in the comment:
> Uncomment the following line if you want to disable marking untracked files under VCS as dirty.

I personally don't consider a repository dirty if I have some extra files in there and I don't want to add those files to .gitignore.
Since I didn't notice any effect from changing this option, I assumed it was a bug and tried to fix it.
If you are against this PR, can you suggest how to achieve this behavior in another way?

akinomyoga:
I expect you to **modify the PR** to solve the problems that I already raised. The problems are

- The variable `DISABLE_UNTRACKED_FILES_DIRTY` is a setting for `lib/git.sh`. There is already a corresponding setting `SCM_GIT_IGNORE_UNTRACKED` for `lib/omb-prompt-base.sh`.
- The assumption that "*`DISABLE_UNTRACKED_FILES_DIRTY` does nothing*" is false. It does change the behavior of `lib/git.sh` used by a few themes such as `sexy`. You need to discard this false assumption for the reasoning of the PR and make a new reason. As long as it's based on a wrong assumption, I wouldn't be able to merge it even if I would be satisfied with the actual change.
- The change conflicts with the latter half of the code comment of the setting, even though it is compatible with the first half. Maybe you are only interested in the first part of the statement in the comment, but that cannot be a reason for us to ignore the second part of the comment.

Also, you are the one who is supposed to *suggest another way* that solves the above problems. There are an infinite number of ways to solve these problems. For example, you could create another configuration variable for `lib/omb-prompt-base.sh` rather than trying to introduce conflicts with the existing setting intended for another purpose. For another example, you could suggest a global refactoring of the configuration variables including the unification of the existing variables split for `lib/git.sh` and `lib/omb-prompt-base.sh` with consistent naming. At the same time, you need to provide a reason why your personal consideration of a dirty repository is worth sharing with others.

awerebea:
Changed the PR's title and initial comment.
Added new `DISABLE_CONSIDER_UNTRACKED_FILES_DIRTY="true"` setting for `lib/omb-prompt-base.sh` instead of current `DISABLE_UNTRACKED_FILES_DIRTY`.

> At the same time, you need to provide a (formal) reason why your personal consideration of a dirty repository is worth sharing with others.

Currently, I'm using `oh-my-zsh` framework with enabled `git` plugin and `powerlevel10k` theme in `zsh` shell.
And I'm used to the fact that the repository is not displayed as dirty if there are untracked files in it. I didn't find a way to achieve the same behavior with `oh-my-bash`, so I implemented this change. In case anyone else is using the mentioned theme, or just prefers to treat their repositories in the same way, I thought I'd share this change via PR.

akinomyoga:
Thank you for deciding on the interface and updating the PR. I think I'll consider merging it, but I think there are still things to update. Could you consider choosing a variable name consistent with the other configurations for `_prompt_git_info` in `lib/omb-promp-base`? Such as `SCM_GIT_IGNORE_UNTRACKED_DIRTY`. If you would like to add the variable to the bashrc template, could you also add the similar setting `SCM_GIT_IGNORE_UNTRACKED` for consistency?

awerebea:
Done!

akinomyoga (approved):
Thanks! I'll merge it after waiting for a few days.

akinomyoga (approved):

PR #459, opened 2023-07-25
aliases/package-manager: Add Fedora Commands to package-manager.alliases.bash

trb143:
Add commands so the DNF is supported for alliases.

PR #460, opened 2023-07-28
aliases/package-manager: Add nala aliases

GR3YH4TT3R93:


akinomyoga (commented):
Thank you for your contribution. Could you also consider updating the list [*alias:package-manager* in `aliases/README.md`](https://github.com/ohmybash/oh-my-bash/blob/master/aliases/README.md#aliaspackage-manager)?

Although I forgot to request it for `dnf` in #459, we need to update it. For `dnf`, I'll update it later.

Also, is there a reason that the aliases for `nala` are added in this position instead of the bottom of the file?

GR3YH4TT3R93:
> Thank you for your contribution. Could you also consider updating the list [_alias:package-manager_ in `aliases/README.md`](https://github.com/ohmybash/oh-my-bash/blob/master/aliases/README.md#aliaspackage-manager)?
> 
> Although I forgot to request it for `dnf` in #459, we need to update it. For `dnf`, I'll update it later.
> 
> Also, is there a reason that the aliases for `nala` are added in this position instead of the bottom of the file?

Done and done!

As for the position of nala, I added it there because Nala is a wrapper for libapt so it felt "right" to have code for associated packages in close proximity.

I can change it if you'd like?

aliases/README.md
```diff
@@ -115,3 +116,19 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=
 | `tplan`  | `terraform plan`             |
 | `tapply` | `terraform apply`            |
 | `tfmt`   | `terraform fmt`              |
+
+## alias:nala
```
akinomyoga, line 120: Thank you for writing a detailed description for each alias. However, I'm not sure what to do with it.

First of all, all of the section titles `alias:*` are actually the module names, which can be loaded by `_omb_module_require`. However, we don't have a module named `alias:nala`. The `nala` aliases are a part of `alias:package-manager` module. At least, we need to give it a different section name and make it a subsection of `## alias:package-manager`. Then, another problem is that we have detailed descriptions only for the `nala` aliases, which is an imbalance. Although it may require some investigations, could you also add similar descriptions to all the other package-manager aliases?

akinomyoga (changes requested):
> As for the position of nala, I added it there because Nala is a wrapper for libapt so it felt "right" to have code for associated packages in close proximity.

That makes sense. Then, let's keep the current position in this PR.

aliases/README.md
GR3YH4TT3R93: My bad, I'm quite new to all this (git, GitHub, contributing to projects, etc) so I appreciate the patience 😊

As for fixing the documentation, what would be the recommended naming scheme/ template for making it a subsection? Same as for the Debian documentation? (e.g. "# Nala Package Manager" w/ descriptions below)

Should I put it at the bottom below the info about sudo? 

And last but not least, in reference to adding adtl descriptions, I can definitely do that!

akinomyoga: > Same as for the Debian documentation? (e.g. "# Nala Package Manager" w/ descriptions below)

That is a good one. One thing to add is that the section level should be`###` to make it a subsection of `## alias:package-manager`.

> Should I put it at the bottom below the info about sudo?

No, "making a subsection of `alias:package-manager`" means that make the section of nala aliases *a part* of the section `alias:package-manager`.

GR3YH4TT3R93: Sorry about that, I think I've addressed it in my latest push. 😅

If I haven't, could you please eli5 where you'd like it to be? Sorry for the inconveniences.

aliases/README.md
```diff
@@ -106,6 +107,63 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=sudo
...
+| `eb`    | `sudo ebuild`                                    | An ebuild must be, at a minimum, a valid Portage package directory name without a version or category, such as portage or python. |
+| `er`    | `sudo emerge -c`                                 | Cleans the system by removing packages that are not associated with explicitly merged packages.                                   |
+| `ers`   | `sudo emerge -c`                                 | Cleans the system by removing packages that are not associated with explicitly merged packages.                                   |
+| `emfu`  | `sudo emerge --sync && sudo emerge -uDUj @world` | emerge Update & Upgrade.                                                                                                          |
+| `elip`  | `sudo eix-installed -a`                          | list all installed programs.                                                                                                      |
+
+### Paludis Package Manager
```
akinomyoga, line 123:
```suggestion
### Paludis Package Manager (`cave`)
```

aliases/README.md
```diff
@@ -106,6 +107,63 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=sudo
...
+| `aptr`  | `sudo apt remove -y`                                                                                               | Performs the requested action on one or more packages specified via regex(7), glob(7) or exact match.                                                                                                                                |
+| `aptar` | `sudo apt autoremove -y`                                                                                           | Autoremove is used to remove packages that were automatically installed to satisfy dependencies for other packages and are now no longer needed as dependencies changed or the package(s) needing them were removed in the meantime. |
+| `aptli` | `sudo apt list`                                                                                                    | List is somewhat similar to dpkg-query --list in that it can display a list of packages satisfying certain criteria.                                                                                                                 |
+
+###  Debian Package Manager
```
akinomyoga, line 145:
```suggestion
### Debian Package Manager (`dpkg`)
```

akinomyoga (approved):
Great work!

Some of the package managers have command names that are different from the names of the package managers. For such package managers, it is good to include the command names in the section headings.

aliases/README.md
akinomyoga: > > Should I put it at the bottom below the info about sudo?
> 
> No, "making a subsection of `alias:package-manager`" means that make the section of nala aliases _a part_ of the section `alias:package-manager`.

Ah, OK. I misunderstood what you mean by the *bottom*. Please forget the above part of my comments.

akinomyoga (dismissed):

akinomyoga (dismissed):

akinomyoga:
Thank you! I'll later merge it if there are no additional changes.

GR3YH4TT3R93:
Sounds good! I'm doing another once over to make absolutely sure everything is perfect right now

GR3YH4TT3R93:
That was the last one! There should be zero issues with spelling or punctuation now.

I would recommend the contributors of Emerge and Paludis to check over the docs for their package managers to make sure that the info is correct. Especially Paludis was hard to source.

akinomyoga (approved):

akinomyoga (dismissed):

GR3YH4TT3R93:
I'm so sorry... I just checked all commands and they now work correctly.

GR3YH4TT3R93:
I'm an idiot who doesn't know how to use got/ github

PR #481, opened 2023-10-02
themes/new-theme

edsonarios:
My reasons for this new topic.
I usually always install oh-my-bash to start with a new machine or virtual or cloud environment, so I like to have all its functionalities quickly, principally git plugin
My customizations:
- I like to see what time I execute my last command
- I like to see only a (*) when there is a change, I only need that, because if I want to know the details or how many files are pending, I just run "gst"
- Sometimes I work with python, and it is good to know if I have the environment activated so as not to make mistakes
- If a command is not executed properly, it is good to see it in a different color
- My main reason is to be able to see the exact route of where I am, because sometimes I work on multiple versions of the same project and not seeing the complete route usually confuses me
- And finally if I am in my HOME directory, I like to see only "~/", this is useful in windows, because if you use the traditional way it shows something like "/c/Users/name" and it is unnecessary information

image

I hope that the next time I install oh-my-bash I only need to change the theme name, to have my customizations, thank you very much 😊


themes/edsonarios/edsonarios.theme.sh
```diff
@@ -0,0 +1,19 @@
...
+OMB_PROMPT_CONDAENV_FORMAT=' [%s]'
+
+STATUS_THEME_PROMPT_BAD="${_omb_prompt_bold_brown}❯_${_omb_prompt_normal} "
+STATUS_THEME_PROMPT_OK="${_omb_prompt_bold_green}❯_${_omb_prompt_normal} "
+
+function _omb_theme_PROMPT_COMMAND {
+  local ret_status="$( [ $? -eq 0 ] && echo -e "$STATUS_THEME_PROMPT_OK" || echo -e "$STATUS_THEME_PROMPT_BAD")"
+  local directory="$(if [[ ${PWD} == $HOME ]]; then echo '\W'; else echo ${PWD}; fi)/"
+
+  _omb_prompt_get_python_venv
```
akinomyoga, line 15:
```suggestion
  local python_venv
  _omb_prompt_get_python_venv
```

themes/edsonarios/edsonarios.theme.sh
```diff
@@ -0,0 +1,19 @@
+#! bash oh-my-bash.module
+# Background: #333333
+# For unstaged(*) and staged(+) values next to branch name in __git_ps1
+GIT_PS1_SHOWDIRTYSTATE="enabled"
+OMB_PROMPT_VIRTUALENV_FORMAT=' [%s]'
+OMB_PROMPT_CONDAENV_FORMAT=' [%s]'
+
+STATUS_THEME_PROMPT_BAD="${_omb_prompt_bold_brown}❯_${_omb_prompt_normal} "
+STATUS_THEME_PROMPT_OK="${_omb_prompt_bold_green}❯_${_omb_prompt_normal} "
```
akinomyoga, line 9: The variable names specific to this theme should start with `OMB_THEME_` such as `OMB_THEME_EDSONARIOS_STATUS_BAD`. We have these variable names in the `purity` theme, but we want to shift to organized variable names (#280). Not to cause unnecessary compatibility issues in the future, we would like to apply the naming convention from the beginning for the new themes.

themes/edsonarios/edsonarios.theme.sh
```diff
@@ -0,0 +1,19 @@
...
+
+function _omb_theme_PROMPT_COMMAND {
+  local ret_status="$( [ $? -eq 0 ] && echo -e "$STATUS_THEME_PROMPT_OK" || echo -e "$STATUS_THEME_PROMPT_BAD")"
+  local directory="$(if [[ ${PWD} == $HOME ]]; then echo '\W'; else echo ${PWD}; fi)/"
```
akinomyoga, line 13: Can you avoid using command substitutions, which involve forks? You can assign values inside the `if` statements. I'd appreciate it if you could also update the corresponding codes in the `purity` theme.

themes/edsonarios/edsonarios.theme.sh
```diff
@@ -0,0 +1,19 @@
+#! bash oh-my-bash.module
+# Background: #333333
```
akinomyoga, line 2: Could you elaborate on this line? Maybe it means that this theme assumes the terminal background color to be `#333333`, but this is merely my guess. It is better to explicitly say what it actually means rather than expecting extra guess work from the readers.

akinomyoga (changes requested):

Could you prepare the image in the size of about 290x800? You should also update `themes/THEMES.md` It's described in [CONTRIBUTING.md](https://github.com/ohmybash/oh-my-bash/blob/master/CONTRIBUTING.md#you-have-an-addition).

> When you add new themes, please also update [themes/THEMES.md](https://github.com/ohmybash/oh-my-bash/blob/master/themes/THEMES.md). The theme needs to have exactly one image of height ~290px and width 600--800px. After your new themes are merged, the list in [Themes](https://github.com/ohmybash/oh-my-bash/wiki/Themes) in the Wiki also needs to be updated.

akinomyoga:
> * And finally if I am in my HOME directory, I like to see only "~/", this is useful in windows, because if you use the traditional way it shows something like "/c/Users/name" and it is unnecessary information

This background information is non-trivial, so it is better to also include the information in a code comment on the corresponding code.

edsonarios:
Thanks for your comments, I updated my PR with the observations, I hope I have resolved them to your satisfaction. Any other comments, I will be happy to resolve them, thank you very much.

themes/edsonarios/edsonarios.theme.sh
edsonarios: Done

themes/edsonarios/edsonarios.theme.sh
edsonarios: Done

themes/edsonarios/edsonarios.theme.sh
edsonarios: Done

themes/edsonarios/edsonarios.theme.sh
edsonarios: Done and comment on its use

akinomyoga:
Thank you for updating. I've pushed additional fixes. Could you check these?

akinomyoga (dismissed):

edsonarios:
All good, thanks so much

akinomyoga (approved):
Thanks for checking!

akinomyoga:
I've merged. Could you also edit [the wiki page of themes](https://github.com/ohmybash/oh-my-bash/wiki/Themes)?

edsonarios:
Sure

edsonarios:
Done, thanks

akinomyoga:
Thank you

PR #484, opened 2023-10-12
lib/git: Fix problems caused by uses of Zsh features

akinomyoga:
This also includes refactoring.

- 779d5fada73e401357a63d727bb7291e42bad495 The main fix for the uses of Zsh constructs.
- e2a9fae9274abde0be6ae8889c44dc4f22ae9cd0 This is to avoid a conflict with the same name of a function defined in another file.
- 960ac2596f68145f48feb0b035d9a226da52ddac This makes the invocation of `command git` for prompts and completions indirect through the function `_omb_prompt_git`. This will be a base for #428 and #173. As far as we directly call `command git`, we need to specify the command modifications for timeouts and git settings for every invocation of `command git`, which is not useful.
- ad07989f4abfa3c6c5fc311e917bf5f78de2470d A utility function to split a string safely.
- a4e0ca1ec9f46f07480a4c6361973c6ff1918fdd and 4d541cf1085dcfb0be33f4724e62c3e7b4e2ff6c Style changes.

### DIFF
--- a/completions/gh.completion.sh
+++ b/completions/gh.completion.sh
@@ -322,7 +322,7 @@ EOF
       format=${format//\p/\2}
       format=${format//\o/\3}
     fi
-    command git config --get-regexp 'remote\.[^.]*\.url' |
+    _omb_prompt_git config --get-regexp 'remote\.[^.]*\.url' |
     grep -E ' ((https?|git)://|git@)github\.com[:/][^:/]+/[^/]+$' |
     sed -E 's#^remote\.([^.]+)\.url +.+[:/](([^/]+)/[^.]+)(\.git)?$#'"$format"'#'
   }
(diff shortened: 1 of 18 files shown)


akinomyoga:
@hyperupcall Hi, how are you doing? The GitHub Action `lint` (*Check EditorConfig Format*), which you introduced in #433, has been very helpful so far. Thank you! I'm reaching out to you because a few days ago, the action seems to have started to always produce an error. Could you take the time to look at it when you have time? It seems to be related to the Go version of the environment, but I haven't carefully examined it.

hyperupcall:
Hey! I'll take a look at it tomorrow. My best guess right now is that an invalid assumption about modules was made, like in relation to `GO111MODULE`s.

akinomyoga:
Thank you!

PR #491, opened 2023-10-26
Fix bashmarks completion

johnf:
Refactor broke completion for bm without any arguments


akinomyoga (approved):
Thanks for the contribution! I've modified the changes and updated the commit message.

PR #492, opened 2023-10-28
Aws completion

johnf:
Add completion to the aws profile plugin.
Also did a bit of cleanup for deprecations etc

### DIFF
--- a/plugins/aws/aws.plugin.sh
+++ b/plugins/aws/aws.plugin.sh
@@ -12,8 +12,6 @@ function agp {
 }
 
 function asp {
-  local rprompt=${RPROMPT/<aws:$(agp)>/}
-
   export AWS_PROFILE=$1
 }
 
--- a/plugins/aws/aws.plugin.sh
+++ b/plugins/aws/aws.plugin.sh
@@ -7,8 +7,6 @@
 # command 'asp' sets the AWS profile to use (aws set profile)
 #
 
-export AWS_HOME=~/.aws
-
 function agp {
   echo $AWS_DEFAULT_PROFILE
 }
--- a/plugins/aws/aws.plugin.sh
+++ b/plugins/aws/aws.plugin.sh
@@ -16,3 +16,12 @@ function asp {
 
   export AWS_PROFILE=$1
 }
+
+function _aws_profiles {
+  grep '\[profile' ~/.aws/config \
+    | grep -v '^[[:space:]]*#' \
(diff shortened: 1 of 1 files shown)


akinomyoga:
@johnf Thank you for your contribution, but I'm not familiar with the history and changes of AWS settings. Could you provide the contexts for each commit?

johnf:
@akinomyoga Let me know if you want me to expand on the messages of the commits themselves but the context is

* remove unneeded AWS_HOME variable - AWS_HOME is not used by the SDKs. I think this was being used as a shortcut in the zsh scripts this was borrowed from
* remove deprecated AWS_DEFAULT_PROFILE - This env variable has not been used by any of the SDKs in years, and is unneeded. AWS_PROFILE has been the convention for a long time
* Add completion - This adds completion for the asp command. The aws config file has sections delimited by `[foo]` for each profile name
* Remove unused rprompt (left over from zsh?) - rprompt is a zshism I believe, again this if from the old zsh code

akinomyoga:
Thank you for the explanation.

> * remove unneeded AWS_HOME variable - AWS_HOME is not used by the SDKs. I think this was being used as a shortcut in the zsh scripts this was borrowed from

I would like to keep the backward compatibility. There is no insurance that no OMB user uses the variable `AWS_HOME` in their `~/.bashrc` [1]. Is there a strong reason to remove the variable, e.g., are any problems caused by the variable? Or another way is to mark `AWS_HOME` as deprecated by `_omb_deprecate_declare 20000 AWS_HOME` while keeping the variable.

- [1] [Code search results- "$AWS_HOME/"](https://github.com/search?q=%22%24AWS_HOME%2F%22&amp;type=code)

By the way, I find https://github.com/ohmyzsh/ohmyzsh/commit/543044efe3940b623069996cf6b2f5127635815f where `AWS_CONFIG_FILE` is used for `~/.aws/config`.

> * remove deprecated AWS_DEFAULT_PROFILE - This env variable has not been used by any of the SDKs in years, and is unneeded. AWS_PROFILE has been the convention for a long time

For the same reason, do we need to remove the variable? OMZ still seems to keep [the legacy name](https://github.com/ohmyzsh/ohmyzsh/blob/38e80e9f21ee981b5f616003c788dcbb9267466c/plugins/aws/aws.plugin.zsh#L25). This can be marked deprecated by `_omb_deprecate_declare 20000 AWS_DEFAULT_PROFILE AWS_PROFILE`.

PR #506, opened 2023-12-12
lib/nvm: update autoloader to keep up with the upstream

### LINKED ISSUE #504: nvm plugin always add "v" for version prefix although it's not a semantic version
According to the nvm's [docs](nwinkler) we can put not only semantic version but also `lts/*` and `node`. In the nvm plugin, it will add prefix "v" from `.nvmrc` file if the first letter is not "v" which will fail to install if the `.nvmrc` is not semantic version like `lts/hydrogen` or `node`

image

akinomyoga:
Fixes #504

This PR applies the changes introduced in https://github.com/nvm-sh/nvm/pull/2017. I skip the upstream changes introduced in https://github.com/nvm-sh/nvm/pull/2313 and in https://github.com/nvm-sh/nvm/pull/3082.

@ffd114 Could you check this branch (`nvm-update` branch at `akinomyoga/oh-my-bash`)?

### DIFF
--- a/plugins/nvm/nvm.plugin.sh
+++ b/plugins/nvm/nvm.plugin.sh
@@ -79,17 +79,17 @@ if _omb_util_command_exists nvm && [[ ${OMB_PLUGIN_NVM_AUTO_USE-} == true ]]; th
       local nvm_version
       nvm_version=$(< "$nvm_path"/.nvmrc)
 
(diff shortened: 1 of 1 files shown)


ffd114:
It works great now. Many thanks.

image

akinomyoga:
Thank you for checking!

PR #509, opened 2023-12-21
New theme for oh-my-bash

TheWatcherMultiversal:
- Theme based on powerline with icon usage | powerline-icon


themes/powerline-icon/powerline-icon.base.sh
```diff
@@ -0,0 +1,171 @@
...
+    "sudo")
+      if [[ "${color}" == "${USER_INFO_THEME_PROMPT_COLOR_SUDO}" ]]; then
+        user_info="!"
+      fi
+      ;;
+    *)
+      if [[ -n "${SSH_CLIENT}" ]]; then
+        user_info="${USER_INFO_SSH_CHAR}${USER}@${HOSTNAME}"
+      else
+        user_info="🐧 Linux `date +%X\ %D`"
```
akinomyoga, line 37: This is not a user info. As far as you modify the function content, the code needs to be consistent with the function name. This function returns the user information except for the part that you modified, so the function name should be kept as is. Your modification broke the consistency of the function name and the actual processing. Instead of modifying an unrelated function, you need to prepare a new function if you would like to show the date with a Linux icon.

themes/powerline-icon/powerline-icon.base.sh
```diff
@@ -0,0 +1,171 @@
...
+  # the space character ' ' whence we need to make sure that the space separated list
+  # stored in POWERLINE_PROMPT is converted into an array correctly.
+  IFS=' ' read -r -a POWERLINE_PROMPT_ARRAY <<< "${POWERLINE_PROMPT}"
+
+  ## left prompt ##
+  for segment in ${POWERLINE_PROMPT_ARRAY[@]}; do
+    local info="$(__powerline_${segment}_prompt)"
+    [[ -n "${info}" ]] && __powerline_left_segment "${info}"
+  done
+  [[ "${last_status}" -ne 0 ]] && __powerline_left_segment $(__powerline_last_status_prompt ${last_status}) || __powerline_left_segment $(__powerline_last_status_prompt ${last_status})
```
akinomyoga, line 162:
```suggestion
  __powerline_left_segment $(__powerline_last_status_prompt ${last_status})
```
The command executed when the condition `[[ "${last_status}" -ne 0 ]]` fails is exactly the same as the one the condition succeeds. These command shouldn't be repeated, but the command should be run without the condition.

akinomyoga (changes requested):
Thanks for your contribution.

First of all, the powerline-based themes should basically inherit the functions defined in `themes/powerline/powerline.base.sh` by sourcing it. Only the functions that would be modified by the `powerline-icon` theme should be defined in `powerline-icon.base.sh`. Please see `powerline-*.base.sh` of the other powerline-based themes:

- [oh-my-bash/themes/powerline-naked/powerline-naked.base.sh at master · ohmybash/oh-my-bash](https://github.com/ohmybash/oh-my-bash/blob/master/themes/powerline-naked/powerline-naked.base.sh)
- [oh-my-bash/themes/powerline-plain/powerline-plain.base.sh at master · ohmybash/oh-my-bash](https://github.com/ohmybash/oh-my-bash/blob/master/themes/powerline-plain/powerline-plain.base.sh)
- [oh-my-bash/themes/powerline-multiline/powerline-multiline.base.sh at master · ohmybash/oh-my-bash](https://github.com/ohmybash/oh-my-bash/blob/master/themes/powerline-multiline/powerline-multiline.base.sh)

TheWatcherMultiversal:
Hello akinomyoga, thank you very much for reviewing my code. I hope the code correction is now correct. I will be attentive to any changes.

akinomyoga:
Thank you for updating! Looks good now! I'll later look at it in detail.

akinomyoga (approved):

akinomyoga:
I applied changes that were introduced to powerline themes in #510.

themes/powerline-icon/powerline-icon.base.sh
```diff
@@ -0,0 +1,39 @@
...
+  esac
+  [[ -n "${user_info}" ]] && echo "🐧 ${user_info} `date +%X\ %D`|${color}|${secondary_color}"
+}
+
+function __powerline_cwd_prompt {
+  echo "$(pwd | sed "s|^${HOME}|🏠|")|${CWD_THEME_PROMPT_COLOR}"
+}
+
+function __powerline_last_status_prompt {
+  [[ "$1" -ne 0 ]] && echo "❌${1}|${LAST_STATUS_THEME_PROMPT_COLOR}" || echo "✅|${LAST_STATUS_THEME_PROMPT_COLOR_2}"
```
akinomyoga, line 38: A new variable `LAST_STATUS_THEME_PROMPT_COLOR_2` seems to be introduced, but could you name it better rather than having a suffix like `_2`? For example, like `LAST_STATUS_THEME_PROMPT_COLOR_SUCCESS`. Also, please update this line to use `if then else fi`.

akinomyoga (commented):
I've checked it again. I have a comment on one line. Could you check it?

themes/powerline-icon/powerline-icon.base.sh
TheWatcherMultiversal: Of course, I'll apply the changes shortly. Thank you for reviewing.

TheWatcherMultiversal:
Done, the changes have been made. I remain attentive to any further modifications. Have a good day!

akinomyoga (approved):

akinomyoga:
Thank you for the update! I'll merge it soon.

PR #510, opened 2023-12-25
themes/powerline: Added hex color capacity and user info text color

GiulianoWF:
- Updated the set_color function to support parsing RGB and Hex color formats.
- Introduced the hex_to_rgb function for converting Hex to RGB color values.
- Modified the __powerline_left_segment function to accept the text color as third argument of each segment.
- Enhanced the __powerline_user_info_prompt function to pass the text color as its third argument.


themes/powerline/powerline.base.sh
```diff
@@ -3,20 +3,60 @@
...
+  if [[ -n "$fg" && -n "$bg" ]]; then
+    # Both foreground and background are provided
+    echo -e "\[\033[${fg};${bg}m\]"
+  elif [[ -n "$fg" ]]; then
+    # Only foreground is provided
+    echo -e "\[\033[${fg}m\]"
+  elif [[ -n "$bg" ]]; then
+    # Only background is provided
+    echo -e "\[\033[${bg}m\]"
   fi
```
akinomyoga, line 53: Couldn't this be realized by just `${var:+...}`?

```suggestion
  echo -e "\[\033[${fg}${fg:+${bg:+;}}${bg}m\]"
```

themes/powerline/powerline.theme.sh
```diff
@@ -7,6 +7,7 @@ POWERLINE_LEFT_SEPARATOR=${POWERLINE_LEFT_SEPARATOR:=""}
 
 USER_INFO_SSH_CHAR=${POWERLINE_USER_INFO_SSH_CHAR:=" "}
 USER_INFO_THEME_PROMPT_COLOR=32
+USER_INFO_THEME_PROMPT_TEXT_COLOR="#4c566a"
```
akinomyoga, line 10: - When you add a new configuration variable of the powerline theme, could you also define it in the other powerline themes (`powerline-*`)?
- Also, could you make the default value within the 256 index colors? Not all the terminals support 24-bit color specifications for SGR. The terminals without the support for 24-bit color will misinterpret the R/G/B values as independent SGR codes, whose results could be disastrous. The 24-bit colors should only be enabled when the user explicitly requests them. More specifically, you can override the variable value after sourcing `oh-my-bash.sh` in `~/.bashrc`.

themes/powerline/powerline.base.sh
```diff
@@ -128,11 +168,16 @@ function __powerline_left_segment {
   IFS="${OLD_IFS}"
   local separator_char="${POWERLINE_LEFT_SEPARATOR}"
   local separator=""
+  local text_color="-"
+  if [[ -n "${params[2]}" ]]; then
+    text_color="${params[2]}"
+  fi
```
akinomyoga, line 174: Is it different from the following?

```suggestion
  local text_color=${params[2]:-"-"}
```

akinomyoga (changes requested):
Thanks for your contribution! Could you check the following comments and update the PR?

GiulianoWF:
Thanks for the feedback! I've updated the PR with the suggested changes.

Regarding the powerline-naked theme, I've added a new variable as per the discussion. It's currently not active due to the way the background color parameter is utilized.

themes/powerline-naked/powerline-naked.theme.sh
```diff
@@ -7,6 +7,7 @@ POWERLINE_LEFT_SEPARATOR=${POWERLINE_LEFT_SEPARATOR:=""}
 
 USER_INFO_SSH_CHAR=${POWERLINE_USER_INFO_SSH_CHAR:=" "}
```
akinomyoga, line 8: > Regarding the powerline-naked theme, I've added a new variable as per the discussion. It's currently not active due to the way the background color parameter is utilized.

Hmm, does it mean `USER_INFO_THEME_PROMPT_COLOR` has been actually used as the text color, and the background color is not specified?

What do you think of describing the situation in the code comment above here? For example, we describe that in the `powerline-naked` theme, the specified background color is used as the text color, that the actual background color is not specified, and that `USER_INFO_THEME_PROMPT_COLOR` is used for the text color. Maybe this description might be wrong since I only quickly checked the code, so please correct it properly in such a case.

Another possibility would be to reverse the background and text colors in the `powerline-naked` theme. Currently, only `${params[1]}` is used for the text color, and `${params[2]}` is ignored by `powerline-naked`, but one could use `${params[2]}` as the background color, and explain to the users in the code comment that the background and text colors are reversed in this theme.

akinomyoga (changes requested):
Thank you for updating! These are remaining points and the discussion about `powerline-neaked`.

themes/powerline-naked/powerline-naked.theme.sh
GiulianoWF: Okay, I've been thinking about it.
The whole confusion arises from the fact that I wanted to change the text color in the user's information for the Powerline theme. I accomplished this by using a TEXT_COLOR variable and passing it around as the second argument. This approach was true for most Powerline themes but not for the naked theme. In the naked theme, the first argument already represented the text color. Therefore, a more accurate description of the variable's responsibility, in all cases and not just some, is that it is a secondary color. Consequently, it should be named as such. At least, that's my opinion.

USER_INFO_THEME_PROMPT_SECONDARY_COLOR

What do you think about it?

akinomyoga: Ah, that is a good idea. +1 for `USER_INFO_THEME_PROMPT_SECONDARY_COLOR`.

GiulianoWF:
I wanted to let you know that I've amended my last commit because I forgot to update a variable name. I wasn't entirely sure about the process and ended up force-pushing the change. I now understand this might not be the best practice, and I apologize for any mix-up it has caused.

If this is an issue, I'm more than willing to close this PR and open a new one with a fresh branch to keep things clean. Just let me know what works best for the project.

Thank you for your understanding.

akinomyoga:
Don't worry about it.

Actually, force-pushing of organized commits is more welcome than trivial commits going back and forth. PRs with good quality would be developed by the discussions, where additional changes are typical. Meanwhile, for later investigations into the background of implementation details, it's good to have an organized history with each commit containing a related set of changes. It might depend on the project, but I prefer to finally have reorganized commits with rebasing/force-pushing/etc in PRs.

Anyway, this time, I'll probably squash all of the commits finally since this introduces a new theme. It's different from a PR containing multiple *independent* fixes to an existing one.

akinomyoga (dismissed):

akinomyoga:
> Anyway, this time, I'll probably squash all of the commits finally since this introduces a new theme. It's different from a PR containing multiple *independent* fixes to an existing one.

Ah, sorry, this is not a new theme. I mistaken it with another PR, #509, which tries to introduce the `powerline-icon` theme. This PR actually contains two independent features, the RGB color support and the specification of the user-info text color. However, these two changes are already mixed in the initial commit, so I wouldn't try to separate it into two commits, i.e., I'll squash them.

(If you would like, you can squash the commits and again split it into two for the RGB color and the user-info text color. In such a case, please force-push them to this PR rather than creating another PR, for the continuation of the discussion.)

GiulianoWF:
Ok, done the changes.

akinomyoga (approved):
Great! Finally, can you adjust the commit messages as `Added` -> `Add`, and `User-info ...` -> `Add user-info ...`.

GiulianoWF:
And done

akinomyoga:
I've merged the PR. Thank you!

PR #513, opened 2023-12-31
Improve aliases of portage

blackteahamburger:
Removed 'ers', improved 'emfu', added 'ec'.

### DIFF
--- a/aliases/README.md
+++ b/aliases/README.md
@@ -118,8 +118,9 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=
 | `esync` | `sudo emerge --sync`                             | Updates repositories, for which auto-sync, sync-type and sync-uri attributes are set in repos.conf.                               |
 | `eb`    | `sudo ebuild`                                    | An ebuild must be, at a minimum, a valid Portage package directory name without a version or category, such as portage or python. |
 | `er`    | `sudo emerge -c`                                 | Cleans the system by removing packages that are not associated with explicitly merged packages.                                   |
-| `ers`   | `sudo emerge -c`                                 | Cleans the system by removing packages that are not associated with explicitly merged packages.                                   |
-| `emfu`  | `sudo emerge --sync && sudo emerge -uDUj @world` | Emerge Update & Upgrade.                                                                                                          |
+| `emfu`  | `sudo emerge --sync && sudo emerge -uDN @world`  | Emerge Update & Upgrade.                                                                                                          |
+| `ecd`   | `sudo eclean-dist -d`                            | Cleans repository source files                                                                                                     |
+| `ecp`   | `sudo eclean-pkg -d`                             | Cleans binary packages                                                                                                            |
 | `elip`  | `sudo eix-installed -a`                          | List all installed programs.                                                                                                      |
 
 ### Paludis Package Manager (`cave`)
--- a/aliases/README.md
+++ b/aliases/README.md
@@ -93,7 +93,7 @@ To activate it, add `debian` to `plugins(...)` in your `.bashrc` file:
 
 This plugin provides the set of aliases that can be used to control package managers.  Here is the list of the supported aliases for each package manager.  You can find the details of each alias in the source [`package-manager.aliases.bash`](package-manager.aliases.bash).
 
-- `emerge` (Portage Enoch Merge) ... `em`, `es`, `esync`, `eb`, `er`, `ers`, `emfu`, `elip`
+- `emerge` (Portage Enoch Merge) ... `em`, `es`, `esync`, `eb`, `er`, `emfu`, `ecd`, `ecp`, `elip`
 - `cave` (Paludis Cave) ... `cave`, `cr`, `cui`, `cs`, `cli`
 - `apt` (Advanced Packaging Tool) ... `apt`, `aptfu`, `apti`, `apts`, `aptr`, `aptar`, `aptli`
 - `dpkg` (Debian Package) ... `dpkg`
@@ -111,17 +111,18 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=
 
 ### Emerge Package Manager
 
-| Alias   | Command                                          | Description                                                                                                                       |
-| ------- | ------------------------------------------------ | --------------------------------------------------------------------------------------------------------------------------------- |
-| `em`    | `sudo emerge`                                    | Emerge is the definitive command-line interface to the Portage system.                                                            |
-| `es`    | `sudo emerge --search`                           | Searches for matches of the supplied string in the ebuild repository.                                                             |
-| `esync` | `sudo emerge --sync`                             | Updates repositories, for which auto-sync, sync-type and sync-uri attributes are set in repos.conf.                               |
-| `eb`    | `sudo ebuild`                                    | An ebuild must be, at a minimum, a valid Portage package directory name without a version or category, such as portage or python. |
-| `er`    | `sudo emerge -c`                                 | Cleans the system by removing packages that are not associated with explicitly merged packages.                                   |
-| `emfu`  | `sudo emerge --sync && sudo emerge -uDN @world`  | Emerge Update & Upgrade.                                                                                                          |
-| `ecd`   | `sudo eclean-dist -d`                            | Cleans repository source files                                                                                                     |
-| `ecp`   | `sudo eclean-pkg -d`                             | Cleans binary packages                                                                                                            |
-| `elip`  | `sudo eix-installed -a`                          | List all installed programs.                                                                                                      |
+| Alias   | Command                                          | Description                                                                                                                                       |
+| ------- | ------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `em`    | `sudo emerge`                                    | Emerge is the definitive command-line interface to the Portage system.                                                                            |
+| `es`    | `sudo emerge --search`                           | Searches for matches of the supplied string in the ebuild repository.                                                                             |
+| `esync` | `sudo emerge --sync`                             | Updates repositories, for which auto-sync, sync-type and sync-uri attributes are set in repos.conf.                                               |
+| `eb`    | `sudo ebuild`                                    | An ebuild must be, at a minimum, a valid Portage package directory name without a version or category, such as portage or python.                 |
+| `er`    | `sudo emerge -c`                                 | Cleans the system by removing packages that are not associated with explicitly merged packages.                                                   |
+| `emfu`  | `sudo emerge --sync && sudo emerge -uDN @world`  | Emerge Update & Upgrade.                                                                                                                          |
+| `ecd`   | `sudo eclean-dist -d`                            | Cleans repository source files                                                                                                                    |
+| `ecp`   | `sudo eclean-pkg -d`                             | Cleans binary packages                                                                                                                            |
+| `elip`  | `sudo eix-installed -a`                          | List all installed programs.                                                                                                                      |
+| `ers`   | `sudo emerge -c`                                 | Deprecated, retained for backwards compatibility. Cleans the system by removing packages that are not associated with explicitly merged packages. |
 
 ### Paludis Package Manager (`cave`)
 
--- a/aliases/README.md
+++ b/aliases/README.md
@@ -122,7 +122,7 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=
 | `ecd`   | `sudo eclean-dist -d`                            | Cleans repository source files                                                                                                                    |
 | `ecp`   | `sudo eclean-pkg -d`                             | Cleans binary packages                                                                                                                            |
 | `elip`  | `sudo eix-installed -a`                          | List all installed programs.                                                                                                                      |
-| `ers`   | `sudo emerge -c`                                 | Deprecated, retained for backwards compatibility. Cleans the system by removing packages that are not associated with explicitly merged packages. |
+| `ers`   | `sudo emerge -c`                                 | (Deprecated, retained for backward compatibility. Use `er` instead.)                                                                              |
 
 ### Paludis Package Manager (`cave`)
 
--- a/aliases/package-manager.aliases.bash
+++ b/aliases/package-manager.aliases.bash
@@ -22,8 +22,8 @@ if _omb_util_binary_exists emerge; then
   alias eb="${_omb_tmp_sudo}ebuild" # Enoch Build
   alias er="${_omb_tmp_sudo}emerge -c" # Enoch Remove
   alias emfu="${_omb_tmp_sudo}emerge --sync && ${_omb_tmp_sudo}emerge -uDN @world" # Enoch Upgrade System
-  # Deprecated, retained for backwards compatibility
-  alias ers="${_omb_tmp_sudo}emerge -c" # Enoch Remove Systempackage
+  # Deprecated, retained for backward compatibility. Use `er` instead.
+  alias ers="${_omb_tmp_sudo}emerge -c"
 fi
 if _omb_util_binary_exists eclean; then
   alias ecd="${_omb_tmp_sudo}eclean-dist -d" # Enoch Clean Repository Source Files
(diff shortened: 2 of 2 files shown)


akinomyoga:

> Removed 'ers',

I'm not familiar with emerge, but as far as I see [the man page](https://dev.gentoo.org/~zmedico/portage/doc/man/emerge.1.html) found online, `er` appears to be the one that should be modified. I.e., `er` should be `emerge --clean` and `ers` should be `emerge --depclean`. The command `emerge -c` seems to be the synonym of `emerge --depclean`, so `er` seems wrong while `ers` seems correct.

> improved 'emfu',

Could you explain the background? What would be the flags `-Uj` and `-N`? As far as I read the man pages `-U` seems cleverer than `-N` but maybe that causes an issue? Also, specifying `-j` seems faster. Anyway, I'm not familiar with it. Could you explain those changes?

blackteahamburger:
> Thank you for your contribution.
> 
> > Removed 'ers',
> 
> I'm not familiar with emerge, but as far as I see [the man page](https://dev.gentoo.org/~zmedico/portage/doc/man/emerge.1.html) found online, `er` appears to be the one that should be modified. I.e., `er` should be `emerge --clean` and `ers` should be `emerge --depclean`. The command `emerge -c` seems to be the synonym of `emerge --depclean`, so `er` seems wrong while `ers` seems correct.
> 

> --clean
Cleans up the system by examining the installed packages and removing older packages. This is accomplished by looking at each installed package and separating the installed versions by slot. Clean will remove all but the most recently installed version in each slot. Clean should not remove unslotted packages. Note: Most recently installed means most recent, not highest version.

`--clean` only removes old slots so its functionality is limited, I don't believe it is widely used.

> > improved 'emfu',
> 
> Could you explain the background? What would be the flags `-Uj` and `-N`? As far as I read the man pages `-U` seems cleverer than `-N` but maybe that causes an issue? Also, specifying `-j` seems faster. Anyway, I'm not familiar with it. Could you explain those changes?

[Gentoo wiki](https://wiki.gentoo.org/wiki/Binary_package_guide#Updating_packages_on_the_binary_package_host) says:

> Do not use --changed-use(-U) when updating packages on the binary package host, doing so will cause packages with added or removed USE flags to be skipped, which will cause their installation from binary package on the client to fail due to non-matching USE between the source ebuild and binary package (if the client's --binpkg-respect-use=y, the default). Use --newuse(-N), which will always rebuild packages even for added or removed USE flags, ensuring the binary package stays in sync with the source ebuild.

With this in mind I replaced `-U` with `-N`.

Also, I think `-j` should be added by the user themselves because `emfu` should focus on completing its own functions.

akinomyoga:
Thank you for your clarifications!

I first need to clarify that my requests are related to *backward compatibility* but not to the *ideal design*. If this PR would be the one that introduces new aliases, I would accept it without any requests. However, I need to be careful about the removal and changes in the behavior because this PR touches existing aliases, which might have already users. Even if some existing aliases are something that not so many people are using, we *do not remove* them unless there is a clear problem.

akinomyoga:
> `--clean` only removes old slots so its functionality is limited, I don't believe it is widely used.

OK. But I still don't see a reason to remove `ers`. You just don't have to use it.

The code comments of aliases `er` and `ers` seem to tell that the difference between the two implies that the former is `--clean` and the latter is `--depclean`, so if anything would be changed, `er` could be changed to use `--clean` because we seem to have a discrepancy between the code comments and the actual aliases. (Or does running emerge with `--clean` break the system?)

> With this in mind I replaced -U with -N.

Thank you for the description. I'm now convinced that we switch to `-N`.

> Also, I think `-j` should be added by the user themselves because `emfu` should focus on completing its own functions.

These are aliases. It's different from an independent command-line tool. It's subtle whether we require the aliases to be focusing on a specific function with a plain setting. Aliases are intended to encapsulate the frequently used set of flags. There is no clear boundary on whether we accept or reject specific additional flags, but many aliases include *additional* flags that are just useful, such as `-i` for `rm`, etc.

For example, I define my own alias that essentially behaves like `alias m='make -j'` because I always specify `-j` to `make`. If anything is broken by specifying `-j`, it means that `Makefile` is broken in the sense that it doesn't correctly specify the dependencies. If the idea of the `-j` option of `emerge` is the same, I'd guess always specifying `-j` is not supposed to cause issues. But I'm not familiar with it, so I need an explanation about it.

Seeing the explanation so far, I still suggest keeping `-j`. It's been there, I didn't receive an issue report caused by it, and the reason for your change seems a philosophical one but not the one to solve an actual issue.

blackteahamburger:
> The code comments of aliases `er` and `ers` seem to tell that the difference between the two implies that the former is `--clean` and the latter is `--depclean`, so if anything would be changed, `er` could be changed to use `--clean` because we seem to have a discrepancy between the code comments and the actual aliases. (Or does running emerge with `--clean` break the system?)

I don't believe the diffirence between `er` and `ers` is `--clean` and `--depclean`. They both mean removing packages, while `--clean` removes old slots (which is already included in `--depclean`), so changing `er` to `--clean` is not appropriate. And this won't maintain backwards compatibility because `er` won't be able to remove packages, which will be confusing. Their  difference is that `ers` removes "system packages", which I think may be packages in `@system`, which is strange since `--depclean` can be used to remove all packages. So I recommend removing `ers`. If you want to maintain backwards compatibility, I think `ers` can be marked as "deprecated" (e.g. emit a message when `ers` is executed?) and it can be removed after a while.

> Seeing the explanation so far, I still suggest keeping `-j`. It's been there, I didn't receive an issue report caused by it, and the reason for your change seems a philosophical one but not the one to solve an actual issue.

What you said makes sense. But there is another variable `MAKEOPT` for specifying parallel `make` jobs. [Gentoo wiki](https://wiki.gentoo.org/wiki/Handbook:AMD64/Installation/Stage#MAKEOPTS) says:

> When using parallel emerges (--jobs), the effective number of jobs run can grow exponentially (up to make jobs multiplied by emerge jobs). This can be worked around by running a localhost-only distcc configuration that will limit the number of compiler instances per host.

With this in mind suggest `-j` should be added by the user themselves.

akinomyoga:
> I don't believe the diffirence between `er` and `ers` is `--clean` and `--depclean`. [...]

Thanks for your opinion. Then, let's not change `er` to use `--clean`. The suggestion was for the case when *"if anything would be changed"*.

> So I recommend removing `ers`. If you want to maintain backwards compatibility, I think `ers` can be marked as "deprecated" (e.g. emit a message when `ers` is executed?) and it can be removed after a while.

I still don't see the reason for removing it. It doesn't seem to cause problems based on the discussion so far. 

> > When using parallel emerges (--jobs), the effective number of jobs run can grow exponentially (up to make jobs multiplied by emerge jobs). This can be worked around by running a localhost-only distcc configuration that will limit the number of compiler instances per host.
> 
> With this in mind I suggest `-j` should be added by the user themselves.

Thank you for those explanations. Does that mean specifying `-j` causes problems in general?

blackteahamburger:
> I still don't see the reason for removing it. It doesn't seem to cause problems based on the discussion so far.

It really doesn't cause problems, but it's a duplicate of `er` and is weird (What is a system package? If it's a package in system then it's exactly the same as `er`. If it's a package in `@system`, it's still redundant). So I think there's a strong case for removing it.

> Thank you for those explanations. Does that mean specifying `-j` causes problems in general?

Not really. [Gentoo wiki](https://wiki.gentoo.org/wiki/EMERGE_DEFAULT_OPTS#Parallel_builds) says:

> When MAKEOPTS="-jN" is used with EMERGE_DEFAULT_OPTS="--jobs K --load-average X.Y" the number of possible tasks created would be up to N*K. Therefore, both variables need to be set with each other in mind as they create up to K jobs each with up to N tasks.

They just need the right combination.

By the way, there's also a varible `EMERGE_DEFAULT_OPTS`, which is another reason why `-j` shouldn't be added to `emfu`.

akinomyoga:
> It really doesn't cause problems, but it's a duplicate of `er` and is weird (What is a system package? If it's a package in system then it's exactly the same as `er`. If it's a package in `@system`, it's still redundant). So I think there's a strong case for removing it.

These seem just the reason for the *ideal design*. I wouldn't accept a change that breaks the *backward compatibility* just for the ideal design. The only reason that would allow to break the backward compatibility would be the situation where the original behavior would be a bug or cause some problems or conflicts.

> Not really. [Gentoo wiki](https://wiki.gentoo.org/wiki/EMERGE_DEFAULT_OPTS#Parallel_builds) says:
> 
> > [...]
> 
> They just need the right combination.

That seems to imply that it would cause problems when used with a *wrong* combination? In particular, when the user doesn't set these variables, does specifying `-j` imply unlimited `K` and cause problems such as OOM? If that is the case, I would accept to drop the option `-j`. Or is portage clever enough to limit the number of the `emerge` jobs depending on the remaining memory? Anyway, I'm now becoming convinced that having `-j` by default would be tricky and wouldn't be so useful.

blackteahamburger:
> These seem just the reason for the _ideal design_. I wouldn't accept a change that breaks the _backward compatibility_ just for the ideal design. The only reason that would allow to break the backward compatibility would be the situation where the original behavior would be a bug or cause some problems or conflicts.

Is backwards compatibility really **that important**? Even the C++ standard removes old facilities. And `ers` is not only weird, but also longer than `er`, so I think the impact of removing it is very low. As mentioned above, it could be deprecated first and then removed after enough time to minimize the impact.

> That seems to imply that it would cause problems when used with a _wrong_ combination? In particular, when the user doesn't set these variables, does specifying `-j` imply unlimited `K` and cause problems such as OOM? If that is the case, I would accept to drop the option `-j`. Or is portage clever enough to limit the number of the `emerge` jobs depending on the remaining memory? Anyway, I'm now becoming convinced that having `-j` by default would be tricky and wouldn't be so useful.

To be honest, I don't know which combination is the most correct.

Gentoo wiki says:

> without an argument, -j will flood the processor with as many jobs as possible, and is not recommended

I just saw it. I used to think that `-j` without parameters was equivalent to `-j$(nproc)`. It seems that it is scarier than I imagined.

There is a `--load-average` option for `emerge` to keep the load average of the system less, it "specifies  that  no  new builds should be started if there are other builds running and the load average is at least LOAD (a floating-point number)". And also for `make`. But I just experienced oom-killer being triggered while compiling (with `--load-average`). Perhaps these options need to be adjusted flexibly, [this article](https://wiki.gentoo.org/wiki/Knowledge_Base:Emerge_out_of_memory) may be useful.

akinomyoga:
> > These seem just the reason for the _ideal design_. I wouldn't accept a change that breaks the _backward compatibility_ just for the ideal design. The only reason that would allow to break the backward compatibility would be the situation where the original behavior would be a bug or cause some problems or conflicts.
> 
> Is backwards compatibility really **that important**?

Yes.

> Even the C++ standard removes old facilities.

This repository is not a *standard* but an *implementation*, meaning that this repository doesn't require other frameworks to copy everything including legacies. The C++ standard dropped some of the old facilities from the *requirements* on the conforming implementations, but that does *not* mean that the conforming implementations *need to drop* the feature. The implementations are always allowed to have extensions where the standard doesn't specify. It's the implementation's discretion whether to become pedantic on strict conformance or to be permissive.

The policy of this repository is to be careful in removing features. In the discussions so far, you didn't seem to care about the backward compatibility at all. However, even the C++ standard you mention cares about backward compatibility, so they survey the existing code bases before removing any features. Only when they confirm the feature is not widely used (or any real problems are found), the features would be dropped.

I should note that it's also related to the availability of human resources. We may follow a similar strategy as the C++ standard in principle, but unfortunately, there is no firm way to perform the survey for Oh My Bash (OMB). There is no way to know the exact size of the user base of OMB, but I estimate it to be the order of 100k users. Even though, it is hard to know the frequency of the uses of each alias under this large user base. Someone would need to perform a real survey on the users, but I doubt it's worth it. This project is not as valuable as you think.

Also, I'm not a user of OMB but merely a passive *maintainer* of OMB as discussed in #268. I started to be involved in OMB to handle the issues in ble.sh caused by OMB and took over the maintenance after a while. The users of OMB are not supposed to be as well-skilled as you think. OMB primarily targets users who are not familiar with Bash enough they configure the Bash settings by themselves. I'm really passive for this specific repository, so I wouldn't like to cause any unnecessary trouble that may puzzle the users. I have to say that this repository already contains tons of unreasonable legacies, but I wouldn't have time and wouldn't be willing to clean them up while handling the questions and reports from the users.

> > without an argument, -j will flood the processor with as many jobs as possible, and is not recommended
> 
> I just saw it. I used to think that `-j` without parameters was equivalent to `-j$(nproc)`. It seems that it is scarier than I imagined.

That's good to know. Then, let's drop the option `-j` as it is in the PR.

blackteahamburger:
In this case I would suggest adding a relevant reminder in the comments rather than leaving it as is to remove confusion as to why `ers` is retained since `ers` itself is weird.

akinomyoga:
I agree with that. Could you add the commentary, and also you can comment that in `aliases/README.md`.

aliases/package-manager.aliases.bash
```diff
@@ -21,8 +21,15 @@ if _omb_util_binary_exists emerge; then
   alias esync="${_omb_tmp_sudo}emerge --sync" # Enoch SYNC
   alias eb="${_omb_tmp_sudo}ebuild" # Enoch Build
   alias er="${_omb_tmp_sudo}emerge -c" # Enoch Remove
+  alias emfu="${_omb_tmp_sudo}emerge --sync && ${_omb_tmp_sudo}emerge -uDN @world" # Enoch Upgrade System
+  # Deprecated, retained for backwards compatibility
```
akinomyoga, line 25:
```suggestion
  # Deprecated, retained for backward compatibility
```

akinomyoga: I sometimes see "backwards compatibility" online, but "backwards" is technically an adverb but not an adjective.

akinomyoga: Can we also add a sentence "Use `cr` instead."?

aliases/README.md
```diff
@@ -111,16 +111,18 @@ OMB_ALIAS_PACKAGE_MANAGER_SUDO=
...
+| `emfu`  | `sudo emerge --sync && sudo emerge -uDN @world`  | Emerge Update & Upgrade.                                                                                                                          |
+| `ecd`   | `sudo eclean-dist -d`                            | Cleans repository source files                                                                                                                    |
+| `ecp`   | `sudo eclean-pkg -d`                             | Cleans binary packages                                                                                                                            |
+| `elip`  | `sudo eix-installed -a`                          | List all installed programs.                                                                                                                      |
+| `ers`   | `sudo emerge -c`                                 | Deprecated, retained for backwards compatibility. Cleans the system by removing packages that are not associated with explicitly merged packages. |
```
akinomyoga, line 125:
```suggestion
| `ers`   | `sudo emerge -c`                                 | (Deprecated, retained for backward compatibility. Use `er` instead.) |
```

aliases/package-manager.aliases.bash
blackteahamburger: [wikipedia](https://en.wikipedia.org/wiki/Backward_compatibility) says:

> Backward compatibility (sometimes known as backwards compatibility) is a property of an operating system, software, real-world product, or technology that allows for interoperability with an older legacy system, or with input designed for such a system, especially in telecommunications and computing.

So this shouldn't be a typo. Do you still think it needs to be modified?

akinomyoga: I understand that you intentionally chose "backwards", and I wouldn't say this *needs* to be modified. However, at least, backward compatibility is grammatically *more* natural and seems to be more common. It's just preferences to some extent, but I'd like to make it sound more natural.

akinomyoga: For example, [‘Backward’ Versus ‘Backwards’ - Quick and Dirty Tips](https://www.quickanddirtytips.com/articles/backward-versus-backwards/#:~:text=I%20do%20have%20to%20say%20I%E2%80%99ve%20seen%20it%20quite%20a%20bit%20online%2C%20as%20in%20%E2%80%9Cbackwards%20compatibility%2C%E2%80%9D%20but%20that%E2%80%99s%20online.%20If%20you%20want%20to%20write%20properly%20for%20your%20job%20or%20for%20schoolwork%20or%20just%20for%20life%2C%20stick%20with%20%E2%80%9Cbackward%E2%80%9D%20for%20the%20adjective.) says *"If you want to write properly for your job or for schoolwork or just for life, stick with “backward” for the adjective.*".

Just to clarify, I'm not requesting you to use "backward compatibility" instead of "backwards compatibility" in your daily conversations or in your documents. However, when we collaborate, we need to find the point of compromise.

akinomyoga (approved):

blackteahamburger:
Thank you for your suggestions!

akinomyoga:
Thank you for your quick responses and patience! I've merged it.

PR #514, opened 2024-01-09
Correction in the powerline-icon theme

TheWatcherMultiversal:
Found an error in the powerline-icon theme; apparently, it was not displaying the icon when a command was executed successfully and returned a status of 0.

PR #516, opened 2024-01-09
Error correction in the powerline-icon theme

TheWatcherMultiversal:
Fixed Bugs: 
- An error in one of the powerline functions has been corrected, which was preventing the display of the icon for a successfully executed command.

### DIFF
--- a/themes/powerline-icon/powerline-icon.base.sh
+++ b/themes/powerline-icon/powerline-icon.base.sh
@@ -41,3 +41,32 @@ function __powerline_last_status_prompt {
     echo "✅|${LAST_STATUS_THEME_PROMPT_COLOR_SUCCESS}"
   fi
 }
+
+function __powerline_prompt_command {
+  local last_status="$?" ## always the first
+  local separator_char="${POWERLINE_PROMPT_CHAR}"
+
+  LEFT_PROMPT=""
+  SEGMENTS_AT_LEFT=0
+  LAST_SEGMENT_COLOR=""
+
+  # The IFS (internal field seperator) may have been changed outside to not contain
+  # the space character ' ' whence we need to make sure that the space separated list
+  # stored in POWERLINE_PROMPT is converted into an array correctly.
+  IFS=' ' read -r -a POWERLINE_PROMPT_ARRAY <<< "${POWERLINE_PROMPT}"
+
+  ## left prompt ##
+  for segment in ${POWERLINE_PROMPT_ARRAY[@]}; do
+    local info="$(__powerline_${segment}_prompt)"
+    [[ -n "${info}" ]] && __powerline_left_segment "${info}"
+  done
+  __powerline_left_segment $(__powerline_last_status_prompt ${last_status})
+  [[ -n "${LEFT_PROMPT}" ]] && LEFT_PROMPT+="$(set_color ${LAST_SEGMENT_COLOR} -)${separator_char}${_omb_prompt_normal}"
+
+  PS1="${LEFT_PROMPT} "
+
+  ## cleanup ##
+  unset LAST_SEGMENT_COLOR \
+        LEFT_PROMPT \
+        SEGMENTS_AT_LEFT
+}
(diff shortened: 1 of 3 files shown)


akinomyoga:
Thanks. I actually expected a force push within the same PR instead of submitting another PR, but this is fine. I shall leave the reference to the original PR: #514.

Also, you have again submitted the PR with the `master` branch. You don't have to recreate the PR again this time, but please do not do that again.

Thanks for updating `THEMES.md`.

Also, this is independent of this PR, but please do not forget to update the wiki page.

themes/powerline-icon/powerline-icon.base.sh
```diff
@@ -41,3 +41,32 @@ function __powerline_last_status_prompt {
...
+  # the space character ' ' whence we need to make sure that the space separated list
+  # stored in POWERLINE_PROMPT is converted into an array correctly.
+  IFS=' ' read -r -a POWERLINE_PROMPT_ARRAY <<< "${POWERLINE_PROMPT}"
+
+  ## left prompt ##
+  for segment in ${POWERLINE_PROMPT_ARRAY[@]}; do
+    local info="$(__powerline_${segment}_prompt)"
+    [[ -n "${info}" ]] && __powerline_left_segment "${info}"
+  done
+  __powerline_left_segment $(__powerline_last_status_prompt ${last_status})
```
akinomyoga, line 63: Let's start the discussion. I rather think this should be fixed in `powerline.base.sh`. For the other segments, we first assign the results to the variable `info`, and only when the variable is no-empty, we call `__powerline_left_segment`. We can follow the same procedure in `powerline.base.sh`.

akinomyoga (changes requested):

TheWatcherMultiversal:
Of course, the wiki page has been updated. Thank you for the additional clarifications, I'm learning over time, and I appreciate that my contribution is accepted.

themes/powerline-icon/powerline-icon.base.sh
akinomyoga: Right.

TheWatcherMultiversal: Only to see if I understood correctly, the function called on line 63 should only be invoked if the 'info' variable for the segments is empty. However, in the original line, this function is executed only if the command executed by the user does not return a status of 0. Should I correct something specific in this line?

akinomyoga: Please modify that line in `powerline.base.sh`. As I have written, first assign the result of `$(__powerline_last_status_prompt ${last_status})` to the variable `info`. Then instead of checking the `last_status`, please check that `info` has a non-empty value. If yes, call `__powerline_left_segment` with the argument being `"$info"`.

TheWatcherMultiversal:
Ready, the changes have been made. It seems to be working as it should, and it doesn't affect other powerline-based themes. If I need to make any more changes, I'll be attentive. Have a great day

themes/powerline/powerline.base.sh
```diff
@@ -192,7 +192,11 @@ function __powerline_prompt_command {
     local info="$(__powerline_${segment}_prompt)"
     [[ -n "${info}" ]] && __powerline_left_segment "${info}"
   done
-  [[ "${last_status}" -ne 0 ]] && __powerline_left_segment $(__powerline_last_status_prompt ${last_status})
+
+  ## info status prompt ##
+  info="$(__powerline_last_status_prompt ${last_status})"
```
akinomyoga, line 197: Thank you for updating. Please make `info` local. The variable `info` is not localized when there are no entries in the array `POWERLINE_PROMPT_ARRAY` because `local info` in the loop is not executed.
```suggestion
  local info="$(__powerline_last_status_prompt ${last_status})"
```

Or, another possibility is to declare `local info` before the for loop.

akinomyoga (changes requested):

themes/powerline/powerline.base.sh
TheWatcherMultiversal: Ready.

akinomyoga (approved):

TheWatcherMultiversal:
Synchronize the PR. I'll be watching for any changes.

akinomyoga:
I've merged. Thank you.

PR #457, opened 2023-07-24
virtual enviroment add in kitsune theme

ankur-gupta-29:
last change does not working.so i do some change in this theme

PR #522, opened 2024-01-18
themes/bakke: Modify - python venv to be visible

msublee:
I modified the python venv to be visible in the terminal (Refer to #457).

### DIFF
--- a/themes/bakke/bakke.theme.sh
+++ b/themes/bakke/bakke.theme.sh
@@ -16,7 +16,15 @@ function _omb_theme_PROMPT_COMMAND() {
     #PS1="${_omb_prompt_bold_teal}$(scm_char)${_omb_prompt_green}$(scm_prompt_info)${_omb_prompt_purple}$(_omb_prompt_print_ruby_env) ${_omb_prompt_olive}\h ${_omb_prompt_reset_color}in ${_omb_prompt_green}\w ${_omb_prompt_reset_color}\n${_omb_prompt_green}→${_omb_prompt_reset_color} "
     #PS1="\n${_omb_prompt_purple}\h: ${_omb_prompt_reset_color} ${_omb_prompt_green}\w\n${_omb_prompt_bold_teal}$(scm_char)${_omb_prompt_green}$(scm_prompt_info) ${_omb_prompt_green}→${_omb_prompt_reset_color} "
     #PS1="\n${_omb_prompt_teal}\h: ${_omb_prompt_reset_color} ${_omb_prompt_olive}\w\n${_omb_prompt_brown}$(scm_char)${_omb_prompt_brown}$(scm_prompt_info) ${_omb_prompt_green}→${_omb_prompt_reset_color} "
-    PS1="\n${_omb_prompt_teal}\h: ${_omb_prompt_reset_color} ${_omb_prompt_olive}\w ${_omb_prompt_green}$(scm_prompt_info)\n${_omb_prompt_reset_color}→ "
+
+    local python_venv
+    _omb_prompt_get_python_venv
+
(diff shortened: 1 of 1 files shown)


themes/bakke/bakke.theme.sh
```diff
@@ -16,7 +16,15 @@ function _omb_theme_PROMPT_COMMAND() {
...
+
+    local python_venv
+    _omb_prompt_get_python_venv
+
+    PS1="\n$python_venv ${_omb_prompt_teal}\h: ${_omb_prompt_reset_color} ${_omb_prompt_olive}\w ${_omb_prompt_green}$(scm_prompt_info)\n${_omb_prompt_reset_color}→ "
 }
 
+OMB_PROMPT_SHOW_PYTHON_VENV=${OMB_PROMPT_SHOW_PYTHON_VENV:-false}
+OMB_PROMPT_VIRTUALENV_FORMAT="${_omb_prompt_bold_gray}(%s)${_omb_prompt_reset_color}"
+OMB_PROMPT_CONDAENV_FORMAT="${_omb_prompt_bold_gray}(%s)${_omb_prompt_reset_color}"
```
akinomyoga, line 28: Can you move these variables before the function? We typically collect these configuration variables at the beginning. Some existing themes do not follow it, but we want to fix them at some point.

themes/bakke/bakke.theme.sh
```diff
@@ -16,7 +16,15 @@ function _omb_theme_PROMPT_COMMAND() {
     #PS1="${_omb_prompt_bold_teal}$(scm_char)${_omb_prompt_green}$(scm_prompt_info)${_omb_prompt_purple}$(_omb_prompt_print_ruby_env) ${_omb_prompt_olive}\h ${_omb_prompt_reset_color}in ${_omb_prompt_green}\w ${_omb_prompt_reset_color}\n${_omb_prompt_green}→${_omb_prompt_reset_color} "
     #PS1="\n${_omb_prompt_purple}\h: ${_omb_prompt_reset_color} ${_omb_prompt_green}\w\n${_omb_prompt_bold_teal}$(scm_char)${_omb_prompt_green}$(scm_prompt_info) ${_omb_prompt_green}→${_omb_prompt_reset_color} "
     #PS1="\n${_omb_prompt_teal}\h: ${_omb_prompt_reset_color} ${_omb_prompt_olive}\w\n${_omb_prompt_brown}$(scm_char)${_omb_prompt_brown}$(scm_prompt_info) ${_omb_prompt_green}→${_omb_prompt_reset_color} "
-    PS1="\n${_omb_prompt_teal}\h: ${_omb_prompt_reset_color} ${_omb_prompt_olive}\w ${_omb_prompt_green}$(scm_prompt_info)\n${_omb_prompt_reset_color}→ "
+
+    local python_venv
+    _omb_prompt_get_python_venv
+
+    PS1="\n$python_venv ${_omb_prompt_teal}\h: ${_omb_prompt_reset_color} ${_omb_prompt_olive}\w ${_omb_prompt_green}$(scm_prompt_info)\n${_omb_prompt_reset_color}→ "
```
akinomyoga, line 23: > `...$python_venv ...`

In this way, an extra space is added at the beginning of the prompt even when there are no python environments. Can you instead add a space at the end of `OMB_PROMPT_{VIRTUALENV,CONDAENV}_FORMAT`?

I'm not sure if there is such a realistic situation, but would you like to avoid having a space between two `(%s)` when both the conda and virtualenv environments are enabled at the same time? If so, an alternative option is to use `${python_venv:+$python_venv }`.

akinomyoga (changes requested):
Thank you for your contribution! Could you check the following comments?

themes/bakke/bakke.theme.sh
msublee: Sure, I modified it.

themes/bakke/bakke.theme.sh
msublee: Thank you for your suggestion. I modified it and commited the changes.

akinomyoga (approved):

akinomyoga:
Thanks!

PR #526, opened 2024-02-11
themes: Add a new theme `developer` focusing on devs

marco-gallegos:
# what's new?

this PR includes a new theme, I have used it around a year, im a developer and is very helpful even better than any other complex configuration in oh_my_zsh  not fully finished but is good enough.

I think is a good idea share this with more people.


akinomyoga:
Can you check the last paragraph of [`CONTRIBUTING.md`](https://github.com/ohmybash/oh-my-bash/blob/master/CONTRIBUTING.md#you-have-an-addition)?

akinomyoga:
> please also update [themes/THEMES.md](https://github.com/ohmybash/oh-my-bash/blob/master/themes/THEMES.md).

marco-gallegos:
sorry I was working on it :D @akinomyoga

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+
+GIT_THEME_PROMPT_DIRTY=" ${_omb_prompt_brown}✗"
+GIT_THEME_PROMPT_CLEAN=" ${_omb_prompt_bold_green}✓"
+GIT_THEME_PROMPT_PREFIX=" ${_omb_prompt_green}|"
+GIT_THEME_PROMPT_SUFFIX="${_omb_prompt_green}|"
+
+RVM_THEME_PROMPT_PREFIX="|"
+RVM_THEME_PROMPT_SUFFIX="|"
+
+function return_delimited {
```
akinomyoga, line 16: Can you prefix `_omb_theme_developer_` to all the function names specific to this theme?

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+		printf "$2 $3---"
+	elif [[ $1 = "false" ]]; then
+		printf "$2 $3"
+	else
+		printf "$1 $2"
+	fi
+}
+
+function extract_key {
+	value=$(echo "$1" | grep -oP "(?<=$2)[^---]+")
```
akinomyoga, line 28: `grep -P` is not POSIX. We do not require the users to install GNU grep. Please use the POSIX-compatible feature.

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+	printf "$(clock_prompt) "
+
+	if [ "${THEME_SHOW_CLOCK_CHAR}" == "true" ]; then
+		printf "$(clock_char) "
+	fi
+}
+
+function node_version {
+	val_node=$(node --version)
+	if command -v nvm &>/dev/null; then
```
akinomyoga, line 42: use `_omb_util_command_exists` to test the existence of the command.

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+		return_delimited $1 "node" "nvm ${val_node}"
```
akinomyoga, line 43:
```suggestion
		return_delimited "$1" "node" "nvm ${val_node}"
```

Can you properly quote the arguments? If you intend to split the arguments, please consider using `_omb_util_split` for a safer splitting. If you intend to perform the glob expansion, please consider using `_omb_util_glob_expand`. (As far as I see the implementation of `return_delimited`, I guess neither word splitting nor glob expansions are expected here.) There are several similar cases in this PR.

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+	else
+		# Si nvm no está instalado, utilizar "njs"
+		return_delimited $1 "ruby" "rb ${val_rb}"
+	fi
+}
+
+function py_version {
+	local val_py=$(python --version | cut -d ' ' -f 2)
+	if command -v conda &>/dev/null; then
+		local condav=$(conda env list | grep '*' | awk '{print $1}')
```
akinomyoga, line 68:
```suggestion
		local condav=$(conda env list | awk '/\*/ {print $1}')
```

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+		return_delimited $1 "python" "conda<${condav}> ${val_py}"
+	else
+		# Si nvm no está instalado, utilizar "njs"
+		return_delimited $1 "python" "py ${val_py}"
+	fi
+}
+
+function OPi5p_Temp {
+	local temp_opi5p=$(cat /sys/class/thermal/thermal_zone4/temp &)
```
akinomyoga, line 77: What is the intent of `&`? There are many cases even for just `echo`. The command substitution anyway blocks until the stream is closed, so putting `&` doesn't change anything in my understanding.

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+	local temp_in_c=$((temp_opi5p / 1000))
+	printf "${temp_in_c}"
```
akinomyoga, line 79:
```suggestion
	printf %s "${temp_in_c}"
```

We don't want to directly specify parameter expansions to the format string of `printf` because it is unclear at a glance whether the expansion result contains the format `%...` that should be expanded. There are similar cases.

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,229 @@
...
+	# NEW way using paralellism
+	# this throws all inside the $() as a new thread but awaits the execution end_time
+	# so this takes the same time as the slower function
+	values=$(
+		getCpuLoad true & # this is very slow
+		getCpuTemp true &
+		py_version true &
+		node_version true &
+		go_version true &
+		getDefaultIp true &
```
akinomyoga, line 198: Is this robust? What happens if the outputs from multiple processes are interleaved? When the system is busy or the user is in the system where the atomicity of the IO operations is not ensured, the output can be mixed.

akinomyoga (changes requested):
Thank you for updating the necessary part for the new theme. Then, let's move on to the actual codes. I left comments. Could you update them?

themes/developer/developer.theme.sh
akinomyoga: Just out of curiosity, what does `[^---]` mean? Is it different from `[^-]` in PCRE or perlre?

marco-gallegos:
awesome, I can update this but not today, let me work on it :D

akinomyoga:
Thanks.

themes/developer/developer.theme.sh
akinomyoga: > Just out of curiosity, what does `[^---]` mean? Is it different from `[^-]` in PCRE or perlre?

I again looked at the code. This `[^---]` is probably intended to split the result by `---`, which is inserted by `return_delimited`. However, this doesn't work. `[...]` is a character class, which means that `[^---]` only sees a single character to check if it matches. This construct cannot be used to exclude the string `---`. This is actually equivalent to `[^-]` and matches any character except for `-`. See this:

```console
$ echo 'aa bb-cc---dd ee-ff' | grep -oP '(?<=aa)[^---]+'
 bb
```

It returns ` bb` instead of `bb-cc`. The correct way to exclude `---` is `(?:(?!---).)+`.

```console
$ echo 'aa bb-cc---dd ee-ff' | grep -oP '(?<=aa)(?:(?!---).)+'
 bb-cc
```

akinomyoga: Also, with this implementation, there is always a space at the beginning. I thought a possibility that this space is intentional so that it separates the element from the previous element, but as far as I see the place it is used, there is already a space, `(`, `)`, etc. to separate the elements. The extra space doesn't seem to be intentional.

Anyway, `grep -P` is not POSIX, and we need to use other way to achieve this. I will change it to use ASCII FS (`$'\x1c'`)to delimit the results.

themes/developer/developer.theme.sh
akinomyoga: cd7e914

themes/developer/developer.theme.sh
akinomyoga: Actually, one doesn't even need to use `$'\x1c'` to delimit the results. One can simply delimit the results with newlines. We do not expect a single prompt element contain newlines.

b4a9b9d

themes/developer/developer.theme.sh
akinomyoga: I now understand why this was not quoted; the first argument of `return_delimited` (`true` or `false`) has been an optional argument. Anyway, this interface was removed in commit b4a9b9d, and all those `$1` do not exist anymore. We can just mark this conversation as resolved.

themes/developer/developer.theme.sh
akinomyoga: 231fb70

themes/developer/developer.theme.sh
akinomyoga: Actually, one should explicitly check if the second column `$2` exactly matches `"*"`.

0fc211f

themes/developer/developer.theme.sh
akinomyoga: I just removed those meaningless backgrounding. As explained above, the command substitution `$()` blocks until the background process terminates, so it doesn't have the desired effect. We might achieve the original purpose by using temporary files, but it complicates the code excessively. Since these didn't work, we can just drop them.

cc6068b

themes/developer/developer.theme.sh
akinomyoga: In this PR, `printf` is used everywhere even for the cases where we know that there is no `%` or `\` to process. There are still other unclear cases where `%` or `\` might be intentionally specified, but as far as I quickly see the usage it doesn't seem to be the case. Rather `printf` seems to be used to print string without a terminating newline. `printf` cannot be simply used for this purpose because `printf` performs the extra processing of `%` or `\`, or it would cause problems. The correct way to perform this is in general, `printf %s "<string to print>"`. Actually, OMB has a utility `_omb_util_put` to encapsulate this. We here replace all the uses of `printf` with `_omb_util_put`.

3cdbf9d

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,214 @@
...
+}
+
+function _omb_theme_developer_cpu_load {
+  # Ejecutar el comando top en modo batch, filtrar por el nombre de usuario
+  # actual y almacenar la salida en la variable 'output'
+  local output=$(top -b -n 1 -u $USER | grep "Cpu(s)")
+
+  # Extraer el porcentaje de carga de la CPU excluyendo el estado "idle" usando
+  # awk
+  local cpu_load=$(echo "$output" | awk '{print 100.0-$8}' | cut -d '.' -f 1)
```
akinomyoga, line 109: This line seems to fail when the current CPU usage is `100.0%`, in which case the CPU usage is connected to the previous field, and `$8` points to the next field. This is an example output:

```
%Cpu(s):  0.0 us,  0.0 sy,  0.0 ni,100.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
```

where `$7` and `$8` were supposed to be `ni,` and `100.0`, respectively, but the actual results of `$7` and `$8` would become `ni,100.0` and `id`, respectively. We should split the line with `,` and pick the field corresponding to `id`.

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,223 @@
...
+    color="${_omb_prompt_red}!"
+  fi
+  _omb_theme_developer_return_delimited "cpuload" "${color}${current_cpu_load}"
+}
+
+function _omb_theme_developer_getCpuTemp {
+  local temp_in_c
+  local currentPlatform=$(_omb_theme_developer_currentPlatform)
+
+  if ((currentPlatform == "linux")); then
```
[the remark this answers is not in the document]
akinomyoga: 0c96cea

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,223 @@
...
+function _omb_theme_developer_get_py_version {
+  local val_py=$(python --version | cut -d ' ' -f 2)
+  if _omb_util_command_exists conda; then
+    local condav=$(conda env list | awk '$2 == "*" {print $1}')
+    _omb_theme_developer_return_delimited "python" "conda<${condav}> ${val_py}"
+  else
+    # Si nvm no está instalado, utilizar "njs"
+    _omb_theme_developer_return_delimited "python" "py ${val_py}"
+  fi
+}
```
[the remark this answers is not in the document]
akinomyoga: The design details are slightly different, and this is not so general. It's fine in the current way.

themes/developer/developer.theme.sh
akinomyoga: c3168eb

themes/developer/developer.theme.sh
akinomyoga: 4c65d8dd262589e263541bd8ccf9a329e7be9427

themes/developer/developer.theme.sh
```diff
@@ -0,0 +1,210 @@
...
+function __bobby_clock {
+  _omb_util_put "$(clock_prompt) "
+
+  if [[ $THEME_SHOW_CLOCK_CHAR == true ]]; then
+    _omb_util_put "$(clock_char) "
+  fi
+}
+
+function _omb_theme_developer_get_node_version {
+  local val_node=$(node --version)
```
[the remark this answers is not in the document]
akinomyoga: d43cf563f10b71efddf8e2932b0d49186d890d0b

Issue #517, opened 2024-01-10
Docker completion is not up-to-date


MonkeyKiki:
Docker Compose V2 is now used with 'docker compose' and no more 'docker-compose'.
But docker completion doesn't propose this feature.
Do you think it can be upgraded ?
I am not sure I can do it myself but i can try.

Thanks

marco-gallegos:
in the meanwhile you can try to  use a alias like 'alias docker-compose="docker compose"' in your .bashrc

akinomyoga:
I don't think updating this will make the completion for `docker compose` work. The completion for `docker compose` is a part of the completion for the `docker` command, so you can just set up the completion for the `docker`.

The `docker` CLI provides the [completion script](https://github.com/docker/cli/blob/master/contrib/completion/bash/docker), and it should be somewhere in your system if you install `docker`. For example, it might be located at something like `/usr/share/bash-completion/completions/docker` or `/etc/bash_completion.d/docker.sh`. If it is the case, as far as you have `bash-completion`, the completion should be automatically available without any plugin.

Do you have `bash-completion` in your system? Do you have any `docker` files in `/usr/share/bash-completion/completions` or in `/etc/bash_completion.d`?

akinomyoga:
Meanwhile, maybe I can still update this script to the latest version of Docker Compose v1, though it doesn't make the completion for `docker compose` work.

https://github.com/docker/compose/blob/v1/contrib/completion/bash/docker-compose

akinomyoga:
OK, the problem seems to be that the docker completion in this repository (`completions/docker.completion.sh`) is so old that it doesn't support the recent plugin completions. We can update this, though I think as far as `bash-completion` and `docker` are properly installed in the user's system, the user doesn't need to enable those completions bundled in Oh My Bash.

PR #527, opened 2024-02-11
completion/docker{,-compose}: Load completions from packages

akinomyoga:
Fxies #517

@MonkeyKiKi @marco-gallegos I've updated the completion scripts. I haven't tested it. Could you test it?

First, it's better to load the completion script coming with Docker and Docker Compose themselves if they are found in the system. So the new version first searches the system for the completion files from the packages. Only when they are not found, the OMB completions source the versions bundled with OMB. I also updated the completions bundled with OMB.


akinomyoga:
OK, then the easiest way is probably to clone the repository and temporarily set the OSH to that cloned directory.

1. Clone the repository into the temporary directory `omb-test`

```bash
$ git clone https://github.com/akinomyoga/oh-my-bash.git omb-test -b update-docker-completion
```

2. Change the value of OSH in your `~/.bashrc`

```bash
# your bashrc

...

# export OSH=~/.oh-my-bash # please comment out the original path
export OSH=/path/to/omb-test # add this line for testing

...
```

Please replace `/path/to/omb-test` to the actual path to the `omb-test` directory created in step 1.

After finishing the test, you can simply revert the change to your `~/.bashrc` and remove the directory `omb-test`.

MonkeyKiki:
Hi @akinomyoga. Thanks for the tutorial.
Your change works fine for me. Would you like me to do more tests ?

akinomyoga:
@MonkeyKiki Thank you for testing! I think it's enough for now. If you notice an issue later, you can any time open a new issue. I'll later merge the PR.

akinomyoga:
Thanks. I've merged it.

PR #529, opened 2024-02-20
feat: add support for ssh completion using files in .ssh/config.d too

typhoe:
From 7.3p1 and up, there is an Include keyword, which allows you to include additional configuration files.  

>Include  
Include the specified configuration file(s).  Multiple pathnames may be specified and each pathname may contain glob(3) wildcards and, for user configurations, shell-like “~” references to user home directories.  Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file or /etc/ssh if included from the system configuration file.  Include directive may appear inside a Match or Host block to perform conditional inclusion.
Source: [ssh_config(5)](http://man7.org/linux/man-pages/man5/ssh_config.5.html).

So if you have `Include config.d/*` in your `~/.ssh/config` file, this pull request allows files put in `~/.ssh/config.d/` to be also parsed for the `Host` keyword to use completion when calling ssh command (in addition to the default `~/.ssh/config`)   

Why would you want to do that: It's easier to maintain lists of hosts in dedicated files (e.g. separate private-home-lan, company1-project1, company2-project3, etc...)

### DIFF
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -5,6 +5,7 @@ _omb_module_require lib:omb-completion
 
 function _omb_completion_ssh {
   local cur
+  local additional_include_option
   local additional_include_defined_file
   _omb_completion_reassemble_breaks :
 
@@ -20,9 +21,14 @@ function _omb_completion_ssh {
   fi
 
   # check if .ssh/config contains Include options
-  for additional_include_defined_file in $(awk -F' ' '/^Înclude/{print $2}' 2>/dev/null) ;do
-    # parse all defined hosts from that file
-    [[ -s "$additional_include_defined_file" ]] &&COMPREPLY+=($(compgen -W "$(grep ^Host "$additional_include_defined_file" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
+  for additional_include_option in $(awk -F' ' '/^Include/{print $2}' "$HOME/.ssh/config" 2>/dev/null) ;do
+    # relative or absolute path, if relative transforms to absolute
+    [[ "${additional_include_option:0:1}" == "/" ]] ||additional_include_option="$HOME/.ssh/$additional_include_option"
+    # for loop to interpret possible globbing
+    for additional_include_defined_file in $additional_include_option ;do
+      # parse all defined hosts from that file
+      [[ -s "$additional_include_defined_file" ]] &&COMPREPLY+=($(compgen -W "$(grep ^Host "$additional_include_defined_file" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
+    done
   done
 
   # parse all hosts found in .ssh/known_hosts
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -5,7 +5,7 @@ _omb_module_require lib:omb-completion
 
 function _omb_completion_ssh {
   local cur
-  local file_in_config_d
+  local additional_include_defined_file
   _omb_completion_reassemble_breaks :
 
   if [[ $cur == *@*  ]] ; then
@@ -19,12 +19,11 @@ function _omb_completion_ssh {
     COMPREPLY=($(compgen -W "$(grep ^Host "$HOME/.ssh/config" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
   fi
 
-  # parse all defined hosts from .ssh/config.d/*
-  if [[ -d $HOME/.ssh/config.d ]]; then
-    for file_in_config_d in "$HOME/.ssh/config.d/"* ;do
-      [[ -s "$file_in_config_d" ]] &&COMPREPLY+=($(compgen -W "$(grep ^Host "$file_in_config_d" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
-    done
-  fi
+  # check if .ssh/config contains Include options
+  for additional_include_defined_file in $(awk -F' ' '/^Înclude/{print $2}' 2>/dev/null) ;do
+    # parse all defined hosts from that file
+    [[ -s "$additional_include_defined_file" ]] &&COMPREPLY+=($(compgen -W "$(grep ^Host "$additional_include_defined_file" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
+  done
 
   # parse all hosts found in .ssh/known_hosts
   if [[ -r $HOME/.ssh/known_hosts ]]; then
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -5,6 +5,7 @@ _omb_module_require lib:omb-completion
 
 function _omb_completion_ssh {
   local cur
+  local file_in_config_d
   _omb_completion_reassemble_breaks :
 
   if [[ $cur == *@*  ]] ; then
@@ -16,6 +17,13 @@ function _omb_completion_ssh {
   # parse all defined hosts from .ssh/config
   if [[ -r $HOME/.ssh/config ]]; then
     COMPREPLY=($(compgen -W "$(grep ^Host "$HOME/.ssh/config" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
+  fi
+
+  # parse all defined hosts from .ssh/config.d/*
+  if [[ -d $HOME/.ssh/config.d ]]; then
+    for file_in_config_d in "$HOME/.ssh/config.d/"* ;do
+      [[ -s "$file_in_config_d" ]] &&COMPREPLY+=($(compgen -W "$(grep ^Host "$file_in_config_d" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
+    done
   fi
 
   # parse all hosts found in .ssh/known_hosts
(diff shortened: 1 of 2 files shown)


akinomyoga (changes requested):
Can you extract the filenames from `Include` in `~/.ssh/config` (instead of assuming a specific directory `~/.ssh/config.d`)?

akinomyoga (changes requested):
Thank you!

You seem to directly source the extracted path `"$additional_include_defined_file"`, but the description you quoted says

> Files without absolute paths are assumed to be in ~/.ssh if included in a user configuration file 

Could you test whether each `additoinal_include_defined_path` is a relative path or an absolute path, and prepend "$HOME/.ssh/" if it's a relative path?

completions/ssh.completion.sh
```diff
@@ -18,6 +19,12 @@ function _omb_completion_ssh {
     COMPREPLY=($(compgen -W "$(grep ^Host "$HOME/.ssh/config" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
   fi
 
+  # check if .ssh/config contains Include options
+  for additional_include_defined_file in $(awk -F' ' '/^Înclude/{print $2}' 2>/dev/null) ;do
```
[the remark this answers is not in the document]
typhoe: yeah, my bad... It's my first pull request on github, I didn't realize that my commit where directly push to the pull request... I changed to WIP as soon as I received the email from github, sorry :'(

Anyway, you saw my first mistake (missing the file to awk) and my second (not checking against absolute or relative path)... funny as I just saw your replies and I stumble against these issues in my tests :-p
I also added a second 'for' loop to interpret globing in the path declared (otherwise config.d/* was not interpreted correctly)
This new commit works as expected with the various Include options I tried
>Include config.d/*
>Include file_in_.ssh_dir
>Include /absolute_path/test_file

typhoe:
Sorry, it's my first pull request using github and I didn't realized it would push my commit directly to the pull request, I change the subjet to WIP as soon as I received the github email.
My bad !

I indeed found the 2 issues you immediately saw (missing file for awk cmd and not checking against absolute or relative path), there was also a type Î instead of I for the "Include" regex in the awk cmd and the need for a second 'for' loop to interpret possible globing in the Include option.
The last commit seems to work correctly this time
Tested with various Include:
> Include config.d/*
> Include file_in_.ssh_dir
> Include /absolute_path/test_file

completions/ssh.completion.sh
```diff
@@ -18,6 +20,17 @@ function _omb_completion_ssh {
     COMPREPLY=($(compgen -W "$(grep ^Host "$HOME/.ssh/config" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
   fi
 
+  # check if .ssh/config contains Include options
+  for additional_include_option in $(awk -F' ' '/^Include/{print $2}' "$HOME/.ssh/config" 2>/dev/null) ;do
```
akinomyoga, line 24: I think this entire `for`-loop is valid only when `~/.ssh/config` is readable, so I think we can put it inside the previous `if` statement of `[[ -r $HOME/.ssh/config ]]`.

completions/ssh.completion.sh
```diff
@@ -18,6 +20,17 @@ function _omb_completion_ssh {
...
+    # relative or absolute path, if relative transforms to absolute
+    [[ "${additional_include_option:0:1}" == "/" ]] ||additional_include_option="$HOME/.ssh/$additional_include_option"
```
akinomyoga, line 26: One can use glob-pattern matching in `[[ ... ]]`.

```suggestion
    [[ "$additional_include_option" == /* ]] ||additional_include_option="$HOME/.ssh/$additional_include_option"
```

completions/ssh.completion.sh
```diff
@@ -5,6 +5,8 @@ _omb_module_require lib:omb-completion
 
 function _omb_completion_ssh {
   local cur
+  local additional_include_option
+  local additional_include_defined_file
```
akinomyoga, line 9: Can you put these declarations just before the corresponding `for` statements? The scope of the variables can be smaller in that way.

completions/ssh.completion.sh
```diff
@@ -18,6 +20,17 @@ function _omb_completion_ssh {
...
 
+  # check if .ssh/config contains Include options
+  for additional_include_option in $(awk -F' ' '/^Include/{print $2}' "$HOME/.ssh/config" 2>/dev/null) ;do
+    # relative or absolute path, if relative transforms to absolute
+    [[ "${additional_include_option:0:1}" == "/" ]] ||additional_include_option="$HOME/.ssh/$additional_include_option"
+    # for loop to interpret possible globing
+    for additional_include_defined_file in $additional_include_option ;do
+      # parse all defined hosts from that file
+      [[ -s "$additional_include_defined_file" ]] &&COMPREPLY+=($(compgen -W "$(grep ^Host "$additional_include_defined_file" | awk '{for (i=2; i<=NF; i++) print $i}' )" "${options[@]}"))
+    done
```
akinomyoga, line 31: We are now performing the command substitutions for every included file, but I think we can first collect the valid file names in an array, and then perform the search at once by specifying multiple files to awk as `awk '/^Host /{for (i=2; i<=NF; i++) print $i}' "${array[@]}"` (I think there is no strong reason to separate the processing in `grep` and `awk`. We can perform both in `awk`) (Note: the actual array name should be chosen properly instead of just `array`). This reduces the fork/exec cost. What do you think?

akinomyoga (commented):
Also, can you consider using utilities [`_omb_util_split`](https://github.com/ohmybash/oh-my-bash/blob/aa5deb804cf490b5b13f1c716c620fe3db8c506f/lib/utils.sh#L369) and [`_omb_util_glob_expand`](https://github.com/ohmybash/oh-my-bash/blob/aa5deb804cf490b5b13f1c716c620fe3db8c506f/lib/utils.sh#L377)?

completions/ssh.completion.sh
typhoe: Ok, I'll change that part

completions/ssh.completion.sh
typhoe: I think that I'll certainly learn a few things trying to improve that code :-p
I'll look at it!

typhoe:
> Also, can you consider using utilities [`_omb_util_split`](https://github.com/ohmybash/oh-my-bash/blob/aa5deb804cf490b5b13f1c716c620fe3db8c506f/lib/utils.sh#L369) and [`_omb_util_glob_expand`](https://github.com/ohmybash/oh-my-bash/blob/aa5deb804cf490b5b13f1c716c620fe3db8c506f/lib/utils.sh#L377)?

I just look at these functions... but I4ll be honest and I'm not sure I understood what/how they do what they do... I'll try to test them and see if I manage to understand.
I'll try my best !

akinomyoga:
Thanks.

> I just looked at these functions... but I4ll be honest and I'm not sure I understood what/how they do what they do...

Maybe I need to add some documentation. The word splitting and the pathname expansions by glob patterns can be affected by various shell settings. Those functions perform word splitting and pathname expansions, respectively, in safe ways.

In the present case, you first split the result of `awk` and, after modifying the relative path, perform the pathname expansion. In this case, we don't want to perform the pathname expansion in the first splitting stage. Also, the pathname expansions can be suppressed by shell settings of `set -f` or `GLOBIGNORE`. Other non-trivial shell options can also affect the behavior of the pathname expansions.

completions/ssh.completion.sh
akinomyoga: f1e0cc0

completions/ssh.completion.sh
akinomyoga: c9198a7

completions/ssh.completion.sh
akinomyoga: c66312e

completions/ssh.completion.sh
akinomyoga: > We are now performing the command substitutions for every included file, but I think we can first collect the valid file names in an array, and then perform the search at once by specifying multiple files to awk

6c221ff

> (Note: I also think there is no strong reason to separate the processing in `grep` and `awk`. We can perform both in `awk`)

a55af5b

akinomyoga:
> > I just looked at these functions... but I4ll be honest and I'm not sure I understood what/how they do what they do...
> 
> Maybe I need to add some documentation.

b007c732d910f3db7fe1ba699ae9ab3e0e6cede6

> Also, can you consider using utilities [`_omb_util_split`](https://github.com/ohmybash/oh-my-bash/blob/aa5deb804cf490b5b13f1c716c620fe3db8c506f/lib/utils.sh#L369) and [`_omb_util_glob_expand`](https://github.com/ohmybash/oh-my-bash/blob/aa5deb804cf490b5b13f1c716c620fe3db8c506f/lib/utils.sh#L377)?

f5d48c39fc978fb8492ee98dd3d5c40a54a9dd15

akinomyoga:
I'm going to rebase.

akinomyoga (approved):

akinomyoga:
@typhoe I have updated the PR. Could you check it?

typhoe:
I checked your modifications and there was a typo I think:
In your "relative or absolute path, if relative transforms to absolute" rewrite, you forgot that the var was an array

With the change, the ssh completion sorks as intended.
Thank you for finishing this PR, I'm sorry I didn't had the time to do it myself

akinomyoga:
Ah, you are right. Thank you.

akinomyoga (approved):

akinomyoga:
Thank you for your contribution! I've merged it.

PR #542, opened 2024-03-17
Add some useful aliases for portage

blackteahamburger:
There is still some useful aliases for portage that have not been added in #513.


akinomyoga (approved):

akinomyoga:
Thanks!

PR #555, opened 2024-04-19
 plugins/{battery,progress,xterm}: Add documentation

JoseMRT2004:
### Add the  _README.md_ file to include a description of the plugins  

1. plugins/xterm: Add documentation
2. plugins/progress.plugin.sh: Add documentation
3. plugins/battery.plugin.sh: Add documentation

### DIFF
--- a/plugins/battery/README.md
+++ b/plugins/battery/README.md
@@ -1,0 +1,61 @@
+# Battery Plugin for Oh My Bash
+
+## Overview
+
+This custom plugin for Oh My Bash enhances your terminal experience by providing functions to monitor and display battery status and information.
+
+## Functions
+
+### 1. `ac_adapter_connected`
+
+- **Description:** Checks if the AC adapter is currently connected.
+- **Returns:**
+  - `0` if the adapter is connected.
+  - Non-zero exit status otherwise.
+
+### 2. `ac_adapter_disconnected`
+
+- **Description:** Checks if the AC adapter is currently disconnected.
+- **Returns:**
+  - `0` if the adapter is disconnected.
+  - Non-zero exit status otherwise.
+
+### 3. `battery_percentage`
+
+- **Description:** Retrieves and displays the current battery charge as a percentage of full (100%).
+- **Returns:**
+  - Battery percentage as an integer.
+
+### 4. `battery_charge`
+
+- **Description:** Presents a graphical representation of the battery charge using ASCII characters.
+- **Returns:**
+  - Graphical representation of the battery charge.
+
+## Usage
+
+1. Ensure the plugin file is properly integrated into your `.bashrc` or `.bash_profile`.
+2. Utilize the provided functions to monitor battery status and information directly in your terminal.
+
+```bash
+# Check if AC adapter is connected
+ac_adapter_connected
+
+# Retrieve battery percentage
+battery_percentage
+
+# Visualize battery charge graphically
+battery_charge
+```
+
+## Dependencies
+
+This plugin relies on several utilities for retrieving battery information:
+
+- `upower`: Primary tool for battery information retrieval.
+- `acpi`, `pmset`, `ioreg`, `WMIC`: Fallback options for battery information retrieval.
+- `/sys/class/power_supply/`: Fallback option for battery information retrieval.
+
+```
+
+```
--- a/plugins/battery/README.md
+++ b/plugins/battery/README.md
@@ -2,60 +2,18 @@
 
 ## Overview
 
-This custom plugin for Oh My Bash enhances your terminal experience by providing functions to monitor and display battery status and information.
-
-## Functions
-
-### 1. `ac_adapter_connected`
-
-- **Description:** Checks if the AC adapter is currently connected.
-- **Returns:**
-  - `0` if the adapter is connected.
-  - Non-zero exit status otherwise.
-
-### 2. `ac_adapter_disconnected`
-
-- **Description:** Checks if the AC adapter is currently disconnected.
-- **Returns:**
-  - `0` if the adapter is disconnected.
-  - Non-zero exit status otherwise.
-
-### 3. `battery_percentage`
-
-- **Description:** Retrieves and displays the current battery charge as a percentage of full (100%).
-- **Returns:**
-  - Battery percentage as an integer.
-
-### 4. `battery_charge`
-
-- **Description:** Presents a graphical representation of the battery charge using ASCII characters.
-- **Returns:**
-  - Graphical representation of the battery charge.
+This custom plugin for Oh My Bash enhances your terminal experience by providing functions to monitor and display battery status and information. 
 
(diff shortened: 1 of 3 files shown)


plugins/battery/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### 4. `battery_charge`
+
+- **Description:** Presents a graphical representation of the battery charge using ASCII characters.
+- **Returns:**
+  - Graphical representation of the battery charge.
+
+## Usage
+
+1. Ensure the plugin file is properly integrated into your `.bashrc` or `.bash_profile`.
```
akinomyoga, line 37:
```suggestion
1. Ensure the plugin file is properly integrated into your `.bashrc` or `.bash_profile`.
```

What does this mean specifically? The plugin file is a separate file from `.bashrc` or `.bash_profile`.

plugins/battery/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+This plugin relies on several utilities for retrieving battery information:
+
+- `upower`: Primary tool for battery information retrieval.
+- `acpi`, `pmset`, `ioreg`, `WMIC`: Fallback options for battery information retrieval.
+- `/sys/class/power_supply/`: Fallback option for battery information retrieval.
+
+```
+
+```
```
akinomyoga, line 61: What is this empty code section?

plugins/progress/README.md
```diff
@@ -0,0 +1,47 @@
+### Using the Progress Bar Script
+
+#### Description
+
+This script provides a simple progress bar GUI for the terminal platform. It allows you to visualize the progress of a task being executed in your shell script.
```
akinomyoga, line 5:
```suggestion
This script provides a simple progress bar TUI for the terminal platform. It allows you to visualize the progress of a task being executed in your shell script.
```

plugins/progress/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+#### Usage
+
+1. **Copy Functions:**
+
+   - Copy the `delay` and `progress` functions directly into your shell script.
```
akinomyoga, line 11: I doubt this is the intended usage, but maybe I need to check the plugin file later.

plugins/xterm/README.md
```diff
@@ -1,3 +1,48 @@
...
+
+2. **`_omb_plugin_xterm_precmd_title`**
+
+   - Updates the xterm title before executing a command.
+   - Includes the user, host, directory, and prompt character.
+
+3. **`_omb_plugin_xterm_preexec_title`**
+
+   - Updates the xterm title after executing a command.
+   - Includes the user, host, directory, and the first part of the executed command.
```
akinomyoga, line 32: I believe these are internal functions and not a part of the public interface.

plugins/xterm/README.md
```diff
@@ -1,3 +1,48 @@
...
+
+4. **`set_xterm_title`**
+   - Manually sets the xterm title with the provided string.
+
+#### Usage
+
+1. **Installation:**
+
+   - Add this script to your shell configuration file (e.g., `.bashrc`, `.bash_profile`).
```
akinomyoga, line 41: What does this mean?

akinomyoga (commented):
Thank you for your contributions. I have questions as commented below.

Also, there are some instructions that the plugin files are *added* or *integrated* in `.bashrc` or `.bash_profile`, but those plugin files are intended to be automatically loaded by `oh-my-bash.sh`  when the plugin names are added in the `plugins` array in `.bashrc` in advance of sourcing `oh-my-bash.sh`. Are those description consistent with this intended usage? Could you elaborate them?

plugins/progress/README.md
JoseMRT2004: Ok

plugins/xterm/README.md
JoseMRT2004: In case you just want to use the script independently without installing oh-my-bash says add it here so it will work. it was understood

plugins/xterm/README.md
JoseMRT2004: You're cool, I thought they could be used externally

JoseMRT2004:
### I can eliminate the lines that do not interest you or are not correct, does this seem like a good option to you?

plugins/battery/README.md
JoseMRT2004: Damn, it was a message I don't remember exactly

akinomyoga: Thanks for updating. Should I wait until you remember it?

plugins/progress/README.md
akinomyoga: - First, the plugins in OMB are intended for interactive uses. For example, the users' shell functions defined in `.bashrc` can utilize the shell functions defined in the plugins in OMB. However, they are not primarily designed to be used within a standalone shell program, though it might still work.
- Next, the intended way is to specify the plugin name in the array `plugins` in `~/.bashrc` and let `oh-my-bash.sh` load them. Even if a user wants to use the plugin in a standalone shell program, the plugin file is supposed to be sourced using the `source` builtin. Its contents are not supposed to be directly copied into the main file of the shell program.

The "Usage" section of the suggested change should explain the standard usage by the interactive shell functions through the `plugins` array.

If you are sure that one can also use the `progress` plugin in a separate shell program, that usage should be introduced as an extra usage and should instruct users to use `source` instead of copying the contents.

plugins/xterm/README.md
akinomyoga: Technically, it is possible to use those functions externally, but it's not intended. It is important to draw a border between the private and public interface (even though we cannot programmatically prevent the external shell functions from using our private functions). In that way, we can maintain compatibility with the future changes. If everything is considered public, we cannot change anything in the future. A certain part needs to be left as a private interface to keep room for future changes.

plugins/xterm/README.md
akinomyoga: Thanks for the explanation. Maybe it is possible to use the `xterm` plugin in that way, but it's not the primary intended usage. Could you explain the intended usage first?

If you want, you may include this special usage with the clarification that it is a special usage.

akinomyoga:
> I can eliminate the lines that do not interest you or are not correct, does this seem like a good option to you?

I'm *interested* in what they mean. I cannot decide whether I want these lines or not unless I understand what they try to explain. Could you explain what they mean? After that, I'll decide whether to make them into the documentation or whether to eliminate them.

akinomyoga:
@JoseMRT2004 You've removed the branch, but can I pick some of the contents in this PR and include them in OMB?

JoseMRT2004:
Yes

akinomyoga:
Thank you!

akinomyoga:
6172dc33fdd81619013eb32e64d7427393c41ba4

Issue #568, opened 2024-05-07
ssh completion doesn't include global ssh hosts


plasmastorm:
It would be good if ssh completion could include global hosts from `/etc/ssh/ssh_known_hosts` and `/etc/ssh/ssh_config`

Would be good to also support `/etc/ssh/ssh_config.d/*` in the same way as https://github.com/ohmybash/oh-my-bash/pull/529 is about to

akinomyoga:
> Would be good to also support `/etc/ssh/ssh_config.d/*` in the same way as #529 is about to

Those files are supposed to be included by `/etc/ssh/ssh_config`, so they are covered by #569.

A problem could be that the `Include`s in `/etc/ssh/ssh_config.d` are not considered with the current implementation of #569. Maybe we can consider supporting the recursive includes, but I'm not sure if we would do that here. The `ssh` completion provided here is a kind of a lightweight version of `bash-completion`, and we already have a full implementation for this processing in the upstream (https://github.com/scop/bash-completion/pull/80 and https://github.com/scop/bash-completion/pull/976). Users who request a certain level of quality should consider move to `bash-completion`.

akinomyoga:
It's already loaded as a default set of configurations:

https://github.com/ohmybash/oh-my-bash/blob/f53beb64b8fe8a3f6ad9df09cc42fa5570406def/lib/bourne-shell.sh#L106-L115

However, there is an open PR #572 to move it to a plugin.

PR #569, opened 2024-05-08
completions/ssh: Include global ssh hosts

plasmastorm:
Partial fix for https://github.com/ohmybash/oh-my-bash/issues/568

### DIFF
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -52,13 +52,15 @@ function _omb_completion_ssh {
   local known_hosts_file
   for known_hosts_file in ~/.ssh/known_hosts /etc/ssh/ssh_known_hosts; do
     if [[ -r $known_hosts_file ]]; then
-      if grep -v -q -e '^ ssh-rsa' "$known_hosts_file"; then
-        known_hosts_files+=("$known_hosts_file")
-      fi
+      known_hosts_files+=("$known_hosts_file")
     fi
   done
   if ((${#known_hosts_files[@]} != 0)); then
-    COMPREPLY+=($(compgen -W "$(awk '{print $1}' "${known_hosts_files[@]}" | grep -v ^\| | cut -d, -f 1 | sed -e 's/\[//g' | sed -e 's/\]//g' | cut -d: -f1 | grep -v ssh-rsa)" "${options[@]}"))
+    COMPREPLY+=($(compgen -W "$(awk '
+      $1 !~ /^\|/ {
+        gsub(/[][]|[,:].*/, "", $1);
+        if ($1 !~ /ssh-rsa/) print $1;
+      }' "${known_hosts_files[@]}")" "${options[@]}"))
   fi
 
   # parse hosts defined in /etc/hosts
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -16,13 +16,14 @@ function _omb_completion_ssh {
 
   local IFS=$'\n'
 
+  local -a config_files=()
   local base_config_file
   for base_config_file in ~/.ssh/config /etc/ssh/ssh_config; do
     # parse all defined hosts from config file
     if [[ -r $base_config_file ]]; then
       local basedir
       basedir=${base_config_file%/*}
-      local -a config_files=("$base_config_file")
+      config_files+=("$base_config_file")
 
       # check if config file contains Include options
       local -a include_patterns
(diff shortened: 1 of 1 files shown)


akinomyoga (commented):
Thank you for your contribution. I merged another PR of ssh_config #529. The suggested addition of the support for `/etc/ssh/ssh_{config,known_hosts}` is good. However, I'd like to reduce the code duplicates before merging.

Could you rebase on top of the latest `master` and modify the code to store the found files in an array and run e.g. `grep ^HOST "${config_files[@]}"` at once?

```bash
local -a config_files=()
if [[ -r ~/.ssh/config ]]; then
  config_files+=(~/.ssh/config)
  ...
fi
if [[ -r /etc/ssh/ssh_config ]]; then
  config_files+=(/etc/ssh/ssh_config)
fi
COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
```

Similar for known_hosts.

plasmastorm:
Will do - `/etc/ssh/ssh_config` can also include files, worth trying to work it in with the new changes by looping over both config files?

akinomyoga:
Ah, right. I think the Include there can also be processed.

completions/ssh.completion.sh
```diff
@@ -16,37 +16,42 @@ function _omb_completion_ssh {
...
-    _omb_util_glob_expand include_files '${include_patterns[*]}'
-    local include_file
-    for include_file in "${include_files[@]}";do
-      # parse all defined hosts from that file
-      [[ -s $include_file ]] && config_files+=("$include_file")
-    done
-
-    COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
-  fi
+  for base_config_file in ~/.ssh/config /etc/ssh/ssh_config; do
```
akinomyoga, line 19: `base_config_file` should be localized.
```suggestion
  local base_config_file
  for base_config_file in ~/.ssh/config /etc/ssh/ssh_config; do
```

The same for the variable `known_hosts_file` used later.

completions/ssh.completion.sh
```diff
@@ -16,37 +16,42 @@ function _omb_completion_ssh {
...
+    # parse all defined hosts from config file
+    if [[ -r $base_config_file ]]; then
+      local basedir
+      basedir=$(dirname $base_config_file)
+      local -a config_files=("$base_config_file")
+
+      # check if config file contains Include options
+      local -a include_patterns
+      _omb_util_split include_patterns "$(awk -F' ' '/^Include/{print $2}' ${base_config_file} 2>/dev/null)" $'\n'
```
akinomyoga, line 28:
```suggestion
      _omb_util_split include_patterns "$(awk -F' ' '/^Include/{print $2}' "${base_config_file}" 2>/dev/null)" $'\n'
```

The parameter expansion should be properly quoted.

completions/ssh.completion.sh
```diff
@@ -16,37 +16,42 @@ function _omb_completion_ssh {
...
-      [[ -s $include_file ]] && config_files+=("$include_file")
-    done
-
-    COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
-  fi
+  for base_config_file in ~/.ssh/config /etc/ssh/ssh_config; do
+    # parse all defined hosts from config file
+    if [[ -r $base_config_file ]]; then
+      local basedir
+      basedir=$(dirname $base_config_file)
```
akinomyoga, line 23:
```suggestion
      basedir=$(dirname "$base_config_file")
```

Actually, we want to reduce the number of forks and execs for performance. We know that these paths contain directory names separated by a slash, so we can simply use

```suggestion
      basedir=${base_config_file%/*}
```

completions/ssh.completion.sh
```diff
@@ -16,37 +16,42 @@ function _omb_completion_ssh {
...
+      for include_file in "${include_files[@]}";do
+        # parse all defined hosts from that file
+        [[ -s $include_file ]] && config_files+=("$include_file")
+      done
 
-  # parse all hosts found in .ssh/known_hosts
-  if [[ -r ~/.ssh/known_hosts ]]; then
-    if grep -v -q -e '^ ssh-rsa' ~/.ssh/known_hosts; then
-      COMPREPLY+=($(compgen -W "$( awk '{print $1}' ~/.ssh/known_hosts | grep -v ^\| | cut -d, -f 1 | sed -e 's/\[//g' | sed -e 's/\]//g' | cut -d: -f1 | grep -v ssh-rsa)" "${options[@]}"))
+      COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
```
akinomyoga, line 44: We want to reduce the number of forks and execs, so we can run this outside the loop.

completions/ssh.completion.sh
akinomyoga: 7958763

completions/ssh.completion.sh
akinomyoga: 99a2f8c

completions/ssh.completion.sh
akinomyoga: 2fb782d

completions/ssh.completion.sh
akinomyoga: dc60e97

akinomyoga (approved):

PR #572, opened 2024-05-16
lib/bourne-shell.sh: Don't force-load bash-completion

ifireball:
Its already done better by the "system" completion module, and doing it there too makes disabling it impossible and debgging issues harder.


akinomyoga (commented):
Thanks. I added `system` to the default set of the completions in the `~/.bashrc` template because `bash-completion` is not loaded by default after the suggested change. The users who want to disable `bash-completion` can remove `system` from the `completions` array.

However, one issue is that we cannot enable the `system` completion for existing users because the updating process of Oh-My-Bash doesn't update the user's `~/.bashrc` (which may be already edited by the user). This means that by applying those change, we disable `bash-completion` of existing users, which is a problem.

ifireball:
Please let me know if you want me to add what I've described, since you've modified the PR I don't want to make further changes without your approval

akinomyoga:
> @akinomyoga is there a way to perhaps print a warning message during upgrade?

It seems no. I've checked the code to update OMB. The upgrading is performed by a script in the older version of OMB (before upgrading), and currently, there doesn't seem to be a way to change the behavior for the new version of OMB after upgrading.

> Wasn't "system" in the default `~/.bashrc` template already for a while now?

No. It's never been there. From the beginning when the `completions` array is added to the template, its content has been `git`, `composer`, and `ssh`, and has never been changed.

> I was sure I wasn't the one that put it there in my local file...

I'm not sure...

> If we absolutely want to keep the current behaviour, we can:
> 
> 1. Add an small hack in `oh-my-bash.sh` to force add `system` to the  `completions` array.
> 2. Have a variable to switch that hack off.
> 3. Set the variable by default in the `~/.bashrc` template.

That's technically possible, but if we would have such a variable, I'd rather do that the other way around. That is,

1. We keep the code in `lib/bourne-shell.sh`
2. Add a variable to turn off loading `bash-completion` in `lib/bourne-shell.sh`

> Please let me know if you want me to add what I've described. Since you've modified the PR I don't want to make further changes without your approval

I want to think about it more.

ifireball:
> That's technically possible, but if we would have such a variable, I'd rather do that the other way around. That is,
> 
>     1. We keep the code in `lib/bourne-shell.sh`
> 
>     2. Add a variable to turn off loading `bash-completion` in `lib/bourne-shell.sh`

That will mean we keep having duplicate code. Perhaps we can do one of:

1. Source `completions/system.completion.sh` from `lib/bourne-shell.sh`.
2. Put the code in a function defined in `lib/bourne-shell.sh` and then call it from `lib/bourne-shell.sh` (Conditioned by a variable) and from `completions/system.completion.sh`
 
> I want to think about it more.

Go ahead.
Perhaps consider adding a mechanism to inform users about changes they may want to make to their configuration after an upgrade. Having that would have been handy now. If you want, I can try to write down an issue to describe what I think it aught to look like and we can discuss the details there.

completions/system.completion.sh
```diff
@@ -1,27 +1,44 @@
...
-if [ $(uname) = "Darwin" ] && _omb_util_command_exists brew; then
+# If Homebrew is installed (OS X), its Bash completion module is loaded.
+if is_os darwin && _omb_util_command_exists brew; then
   BREW_PREFIX=$(brew --prefix)
 
-  if [ -f "$BREW_PREFIX"/etc/bash_completion ]; then
-    . "$BREW_PREFIX"/etc/bash_completion
+  # homebrew/versions/bash-completion2 (required for projects.completion.bash)
+  # is installed to this path
+  if [[ -f "$BREW_PREFIX"/share/bash-completion/bash_completion ]]; then
```
RobLoach, line 19: One request I would make here is the ability to set a `BASH_COMPLETION_DIR`, and check it as a potential location.

akinomyoga: I'm not sure if that is useful. If you know the location of bash_completion and needs to specify it explicitly, how is it different from `source '<path to bash_completion>/bash_completion'`? A plugin that allows users to write

```bash
BASH_COMPLETION_DIR='<path to bash_completion>/bash_completion'

plugins+=(system)
```

instead of

```bash
source '<path to bash_completion>/bash_completion'
```

doesn't seem useful. It actually complicates the configuration.

Also, the plugin name `system` implies loading `bash-completion` in a system package. Allowing the configuration for an arbitrary location of `bash-completion` doesn't match the plugin's name. If the configuration would be added, the plugin name should also be changed.

In addition, `BASH_COMPLETION_*` is a kind of variable namespace used by `bash-completion`, so we should avoid defining a variable of the form `BASH_COMPLETION_*` outside `bash-completion`. It might conflict in the future.

https://github.com/scop/bash-completion/blob/main/doc/api-and-naming.md#naming
https://github.com/scop/bash-completion/blob/main/doc/configuration.md

PR #583, opened 2024-08-09
add Wiki "Themes" page updater script

ZaLiTHkA:
I got bored, and I was looking through OMB themes, when I noticed that there were a great many discrepancies between the "Themes" page in Wiki and the "themes" folder in this repo...

tl;dr: I scripted an update process for the Wiki file that uses content as it currently exists inside this repo.

but I found I was still bored, so I made it configurable via either environment variables or CLI args.. and documented it's usage at the top of the script.

then I tested it a whole bunch to make sure I couldn't break it..

anyways, question for the repo maintainers: would this perhaps be helpful to you guys?

---

note: this does require one small change to the `Themes.md` file, specifically as follows:

```diff
diff --git a/Themes.md b/Themes.md
index 9917bb2..4e77dca 100644
--- a/Themes.md
+++ b/Themes.md
@@ -1,8 +1,10 @@
 ## Where to find the themes?
 
 The themes are all located inside the themes folder [here](https://github.com/ohmybash/oh-my-bash/tree/master/themes)
 
+<!-- THEME_GEN_START_MARKER -->
+
 ## `90210`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/90210/90210-dark.png)
 
@@ -275,7 +277,9 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `zork`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/zork/zork-dark.png)
 
+<!-- THEME_GEN_END_MARKER -->
+
 _*Credit:*_
 
 All credits go to [Israel-Laguan](https://github.com/Israel-Laguan). Thank you so much for your hard work with the screenshots here :)

```

### DIFF
--- a/tools/update-wiki-themes.sh
+++ b/tools/update-wiki-themes.sh
@@ -28,10 +28,10 @@
 
 # first process current env vars, with some sensible default fallback values...
 
-SCRIPT_WIKI_PATH=${OMB_WIKI_PATH:-../oh-my-bash.wiki}
-SCRIPT_WIKI_THEMES_FILE=${OMB_WIKI_THEMES_FILE:-Themes.md}
-SCRIPT_WIKI_THEMES_START_MARKER=${OMB_WIKI_THEMES_START_MARKER:-'<!-- THEME_GEN_START_MARKER -->'}
-SCRIPT_WIKI_THEMES_END_MARKER=${OMB_WIKI_THEMES_END_MARKER:-'<!-- THEME_GEN_END_MARKER -->'}
+OMB_WIKI_PATH=${OMB_WIKI_PATH:-../oh-my-bash.wiki}
+OMB_WIKI_THEMES_FILE=${OMB_WIKI_THEMES_FILE:-Themes.md}
+OMB_WIKI_THEMES_START_MARKER=${OMB_WIKI_THEMES_START_MARKER:-'<!-- OMB_WIKI_THEMES_START_MARKER -->'}
+OMB_WIKI_THEMES_END_MARKER=${OMB_WIKI_THEMES_END_MARKER:-'<!-- OMB_WIKI_THEMES_END_MARKER -->'}
 
 # then process any cli args, if provided...
 
@@ -44,22 +44,22 @@ while (($#)); do
   case $1 in
   -p | --wiki-path)
     # echo "Processing 'wiki-path' option. Input argument is '$2'"
-    SCRIPT_WIKI_PATH=$2
+    OMB_WIKI_PATH=$2
     shift 2
     ;;
   -f | --themes-file)
     # echo "Processing 'themes-file' option. Input argument is '$2'"
-    SCRIPT_WIKI_THEMES_FILE=$2
+    OMB_WIKI_THEMES_FILE=$2
     shift 2
     ;;
   -s | --start-marker)
     # echo "Processing 'start-marker' option. Input argument is '$2'"
-    SCRIPT_WIKI_THEMES_START_MARKER=$2
+    OMB_WIKI_THEMES_START_MARKER=$2
     shift 2
     ;;
   -e | --end-marker)
     # echo "Processing 'end-marker' option. Input argument is '$2'"
-    SCRIPT_WIKI_THEMES_END_MARKER=$2
+    OMB_WIKI_THEMES_END_MARKER=$2
     shift 2
     ;;
   --)
@@ -70,30 +70,30 @@ while (($#)); do
 done
 
 # debug: this will either be adapted for the final script, or removed entirely..
-printf '%s\n' "current SCRIPT_WIKI_PATH: $SCRIPT_WIKI_PATH"
-printf '%s\n' "current SCRIPT_WIKI_THEMES_FILE: $SCRIPT_WIKI_THEMES_FILE"
-printf '%s\n' "current SCRIPT_WIKI_THEMES_START_MARKER: $SCRIPT_WIKI_THEMES_START_MARKER"
-printf '%s\n' "current SCRIPT_WIKI_THEMES_END_MARKER: $SCRIPT_WIKI_THEMES_END_MARKER"
+printf '%s\n' "current OMB_WIKI_PATH: $OMB_WIKI_PATH"
+printf '%s\n' "current OMB_WIKI_THEMES_FILE: $OMB_WIKI_THEMES_FILE"
+printf '%s\n' "current OMB_WIKI_THEMES_START_MARKER: $OMB_WIKI_THEMES_START_MARKER"
+printf '%s\n' "current OMB_WIKI_THEMES_END_MARKER: $OMB_WIKI_THEMES_END_MARKER"
 
 # verify that the OMB Wiki project exists...
-if [[ ! -d $SCRIPT_WIKI_PATH ]]; then
-  printf '%s\n' "ERROR: Wiki project path '$SCRIPT_WIKI_PATH' does not exist."
+if [[ ! -d $OMB_WIKI_PATH ]]; then
+  printf '%s\n' "ERROR: Wiki project path '$OMB_WIKI_PATH' does not exist."
   exit 1
 fi
 
 # verify that the OMB Wiki contains the expected themes file...
-if [[ ! -f $SCRIPT_WIKI_PATH/$SCRIPT_WIKI_THEMES_FILE ]]; then
-  printf '%s\n' "ERROR: Wiki project has no themes file called '$SCRIPT_WIKI_THEMES_FILE'."
+if [[ ! -f $OMB_WIKI_PATH/$OMB_WIKI_THEMES_FILE ]]; then
+  printf '%s\n' "ERROR: Wiki project has no themes file called '$OMB_WIKI_THEMES_FILE'."
   exit 1
 fi
 
 # verify that the OMB Wiki contains the expected start and end markers...
-if ! grep -q "$SCRIPT_WIKI_THEMES_START_MARKER" "$SCRIPT_WIKI_PATH/$SCRIPT_WIKI_THEMES_FILE"; then
-  printf '%s\n' "ERROR: Wiki themes file does not contain start marker '$SCRIPT_WIKI_THEMES_START_MARKER'."
+if ! grep -q "$OMB_WIKI_THEMES_START_MARKER" "$OMB_WIKI_PATH/$OMB_WIKI_THEMES_FILE"; then
+  printf '%s\n' "ERROR: Wiki themes file does not contain start marker '$OMB_WIKI_THEMES_START_MARKER'."
   exit 1
 fi
-if ! grep -q "$SCRIPT_WIKI_THEMES_END_MARKER" "$SCRIPT_WIKI_PATH/$SCRIPT_WIKI_THEMES_FILE"; then
-  printf '%s\n' "ERROR: Wiki themes file does not contain end marker '$SCRIPT_WIKI_THEMES_END_MARKER'."
+if ! grep -q "$OMB_WIKI_THEMES_END_MARKER" "$OMB_WIKI_PATH/$OMB_WIKI_THEMES_FILE"; then
+  printf '%s\n' "ERROR: Wiki themes file does not contain end marker '$OMB_WIKI_THEMES_END_MARKER'."
   exit 1
 fi
 
@@ -103,7 +103,7 @@ fi
 theme_list=$(find "./themes" -mindepth 1 -maxdepth 1 -type d -print | sort | xargs -n1 basename)
 
 # prepare a variable to hold generated content, starting with with the "start" marker for next run...
-markdown_text="$SCRIPT_WIKI_THEMES_START_MARKER\n\n"
+markdown_text="$OMB_WIKI_THEMES_START_MARKER\n\n"
 
 # now we can loop through the list and find all images in each theme directory...
 for theme in $theme_list; do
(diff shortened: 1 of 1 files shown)


ZaLiTHkA:
for what it's worth, the `Themes.md` Wiki file doesn't actually have any major changes after running this script.. mostly just that `## font (default)` now generates as `## font`, everything else is just inserting themes that were previously not there.

patch output for reference:

```diff
diff --git a/Themes.md b/Themes.md
index 9917bb2..f2a05b0 100644
--- a/Themes.md
+++ b/Themes.md
@@ -1,16 +1,22 @@
 ## Where to find the themes?
 
 The themes are all located inside the themes folder [here](https://github.com/ohmybash/oh-my-bash/tree/master/themes)
 
+<!-- THEME_GEN_START_MARKER -->
+
 ## `90210`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/90210/90210-dark.png)
 
+## `absimple`
+
+![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/absimple/absimple-bash.png)
+
 ## `agnoster`
 
-![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/agnoster/agnoster-bash-sshot.png)
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/agnoster/agnoster-dark.png)
+![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/agnoster/agnoster-bash-sshot.png)
 
 ## `axin`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/axin/axin-dark.png)
@@ -94,16 +100,20 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `envy`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/envy/envy-dark.png)
 
-## `font` (default theme)
+## `font`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/font/font-dark.png)
 
 ## `gallifrey`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/gallifrey/gallifrey-dark.png)
 
+## `garo`
+
+WARNING: theme contains no example images.
+
 ## `half-life`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/half-life/half-life.theme.png)
 
@@ -171,8 +181,12 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `pete`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/pete/pete-dark.png)
 
+## `powerbash10k`
+
+![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/powerbash10k/powerbash10k-dark.png)
+
 ## `powerline`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/powerline/powerline-dark.png)
 
@@ -211,8 +225,12 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `purity`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/purity/purity-dark.png)
 
+## `pzq`
+
+![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/pzq/pzq-dark.png)
+
 ## `rainbowbrite`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/rainbowbrite/rainbowbrite-dark.png)
 
@@ -223,8 +241,12 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `rjorgenson`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/rjorgenson/rjorgenson-dark.png)
 
+## `robbyrussell`
+
+![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/robbyrussell/robbyrussell-dark.png)
+
 ## `roderik`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/roderik/roderik-dark.png)
 
@@ -263,8 +285,12 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `tylenol`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/tylenol/tylenol-dark.png)
 
+## `vscode`
+
+![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/vscode/vscode-dark.png)
+
 ## `wanelo`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/wanelo/wanelo-dark.png)
 
@@ -275,7 +301,9 @@ The themes are all located inside the themes folder [here](https://github.com/oh
 ## `zork`
 
 ![](https://raw.githubusercontent.com/ohmybash/oh-my-bash/master/themes/zork/zork-dark.png)
 
+<!-- THEME_GEN_END_MARKER -->
+
 _*Credit:*_
 
 All credits go to [Israel-Laguan](https://github.com/Israel-Laguan). Thank you so much for your hard work with the screenshots here :)
```

ZaLiTHkA:
I also just noticed that I left some debug output in there... that can be cleaned up if need be.

akinomyoga:
I adjusted coding styles.

Also, I normalized the variable names and interface names in commit 318d819. Originally, three different naming styles are mixed: E.g. `OMB_WIKI_THEMES_START_MARKER` for the environment variable name, `SCRIPT_WIKI_THEMES_START_MARKER` for the internal variable name, and `THEME_GEN_START_MARKER` for the actual marker string. I think there is no reason to use different names, and it is easier to use a single consistent name when understanding and maintaining them.

akinomyoga:
Currently, this script requires two separate options `-p` (`OMB_WIKI_PATH`) for the Wiki directory, and `-f` (`OMB_WIKI_THEMES_FILE`) for the `Themes.md` file under the directory specified by `-p`. In the actual processing, I see only the combination `$OMB_WIKI_PATH/$OMB_WIKI_THEMES_FILE` seems to be used. I'd like to combine these two options to one option `-f`, which directly specifies the path to the Markdown document of the themes. @ZaLiTHkA What do you think?

ZaLiTHkA:
I have no issue with any of that, naming things is hard... lol. as is trying to pick appropriate CLI arg flags. this was just a first pass at it really.

> Currently, this script requires two separate options `-p` (`OMB_WIKI_PATH`) for the Wiki directory, and `-f` (`OMB_WIKI_THEMES_FILE`) for the `Themes.md` file under the directory specified by `-p`. In the actual processing, I see only the combination `$OMB_WIKI_PATH/$OMB_WIKI_THEMES_FILE` seems to be used. I'd like to combine these two options to one option `-f`, which directly specifies the path to the Markdown document of the themes. @ZaLiTHkA What do you think? Is there a reason to leave a degree of freedom allowing users to specify them separately?

yeah, I'm not sure why I did it *that* way specifically.. if it's safe to assume that the Wiki's `Themes.md` file won't be moved or renamed (at least, not very often anyway), then by all means it can be simplified.

I was mostly trying to ensure this did not force any contributor into any specific workspace folder structure. if they want to clone the Wiki files to a completely different path, they're free to do so.

ZaLiTHkA:
> I think it is more explicit to let the user specify the filename rather than the directory name that would include it, e.g., `tools/update-wiki-themes.sh -f ../oh-my-bash.wiki/Themes.md` rather than `tools/update-wiki-themes.sh -p ../oh-my-bash.wiki`. It requires extra typing `/Themes.md`, but while inputting the command, it reminds the users (me) of which file would be overwritten.

that makes perfect sense. perhaps in that moment I was simply trying to cater for too many different potential use cases. at the end of the day, the goal here is to simplify things.

> By the way, I have a question: Do you have the "push" privilege to the Oh My Bash Wiki pages? I know everyone can clone or pull the wiki pages through the git command, but I thought the "push" privilege was only allowed for the maintainers/collaborators (so the normal users would need to edit the wiki pages from the web interface of GitHub). But maybe I'm not correct.

I didn't actually try push to the Wiki path, simply cloned it for testing...

I did just try push the Wiki's `master` branch now and SourceGit just closed on me with no error message. starting it up again I see it did not push anything. so I think it's safe to say that only authorised users will be able to push Wiki updates directly.

ZaLiTHkA:
one potentially sticky point here though: I wrote this under the assumption that the `oh-my-bash` repo root path would be the current working directory. so running it from anywhere else would fail.

should it perhaps get the script location and calculate an appropriate working directory first, just to be safe?

akinomyoga:
> > By the way, I have a question: Do you have the "push" privilege to the Oh My Bash Wiki pages? I know everyone can clone or pull the wiki pages through the git command, but I thought the "push" privilege was only allowed for the maintainers/collaborators (so the normal users would need to edit the wiki pages from the web interface of GitHub). But maybe I'm not correct.
> 
> I didn't actually try push to the Wiki path, simply cloned it for testing...
> 
> I did just try push the Wiki's `master` branch now and SourceGit just closed on me with no error message. starting it up again I see it did not push anything. so I think it's safe to say that only authorised users will be able to push Wiki updates directly.

Thank you for trying that! Then, I think I'll have to update the page regularly. I actually have a subdirectory `wiki` under the `oh-my-bash` tree, so it's personally useful that `OMB_WIKI_THEMES_FILE` would be defaulted to `./wiki/Themes.md`.

\# I thought that, if the normal users could push the page, we could ask the user who added a theme to run the command and push it, but it turned out to be difficult. We could still ask the user to copy and paste the content of the new `Themes.md` in the GitHub web interface, but that doesn't seem useful.

> one potentially sticky point here though: I wrote this under the assumption that the `oh-my-bash` repo root path would be the current working directory. so running it from anywhere else would most likely fail.

Yeah, I also notice that.

> should it perhaps get the script location and calculate an appropriate working directory first, just to be safe?

Yes, that is one option. At least, we can attempt automatic resolution of the default location. Maybe we can have an option for specifying the location of the OMB tree.

ZaLiTHkA:
lol. I was just about to mention that the default `Themes.md` path was missing a `/`, and then suddenly it was fixed.. 🍻 

---

with regards to streamlining the process of contributing a new theme to the repo, I could see something like the following:

1. create the theme, committed to a custom repo fork
2. create a PR here to get it merged
3. use a GitHub Action to trigger a Wiki page update on merge

but I must admit I do not have much experience with GitHub Actions. I'm looking into them for other reasons, but I have not actually implemented any as yet.

---

with regards to having the script calculate a safe working directory, what about something like this:

```sh
SELF_PATH=$(realpath "${BASH_SOURCE[-1]}")
OMB_REPO_ROOT=$(dirname "$SELF_PATH")
```

I *believe* this should work the same in all *nix shells, including Bash and Zsh. but having said that, if it will only ever be used from a GitHub Action, then perhaps it's not actually necessary?

ZaLiTHkA:
also just to throw a small spanner in the works: will there actually be any reason to pass a custom "start" or "end" marker for the generated output as a CLI arg? I'm struggling to come up with any valid reason to make those configurable.

akinomyoga:
> lol. I was just about to mention that the default `Themes.md` path was missing a `/`, and then suddenly it was fixed.. 🍻

Ah, yes. I've noticed it after pushing it. Anyway, thank you for pointing it out!

----

> with regards to streamlining the process of contributing a new theme to the repo, I could see something like the following:
> 
> 1. create the theme, committed to a custom repo fork
> 2. create a PR here to get it merged
> 3. use a GitHub Action to trigger a Wiki page update on merge

Ah, yeah. If the GitHub Action can update the wiki page, that'd be useful.

> but I must admit I do not have much experience with GitHub Actions.

Hmm, me neither. In such a case, so that the contents edited by other users will not be automatically overwritten, we may additionally need to put a warning "Do NOT EDIT THIS SECTION!" in `Themes.md`.

----

> with regards to having the script calculate a safe working directory, what about something like this as a baseline:
> 
> ```shell
> SELF_PATH=$(realpath "${BASH_SOURCE[-1]}")
> OMB_REPO_ROOT=$(dirname "$SELF_PATH")
> ```

Oh, thank you for your suggestion! I already started to implement it without noticing it. The current version is in commit 5bc810251190dcc2a3f0d1c719cb60c510c78714. If you have any suggestion, please feel free to leave a comment!

> I _believe_ this should work the same in all *nix shells, including Bash and Zsh. but having said that, if it will only ever be used from a GitHub Action, then perhaps it's not actually necessary?

Ah, right. However, it would still be nice to allow executing it by hand. Anyway, I've already implemented it.

----

> also just to throw a small spanner in the works: will there actually be any reason to pass a custom "start" or "end" marker for the generated output as a CLI arg? I'm struggling to come up with any valid reason to make those configurable.

Hmm, good points.

akinomyoga:
@ZaLiTHkA I tested the script locally. This is very useful! Thank you! I added several additional adjustments that I noticed while testing.

ZaLiTHkA:
I'm technically on work-time right now, so I *really* shouldn't be doing this right now.. but since I am also investigating GitHub Actions for work stuff, so I couldn't resist doing a little research in the mean time.

TL;DR: as far as I can tell, it seems like it is possible to configure a GitHub Action to clone the wiki and execute a script file from the code repo when a PR is merged, pushing changes back to the wiki afterwards.

but I don't think we should block this PR with that "nicety" right now. if the script is help as it is, then maybe rather just slot it into your existing workflow for adding new themes. I'll dig deeper into this idea later on today and see what comes of it.

akinomyoga (approved):

akinomyoga:
Thank you! I merged it. I also updated `Themes.md` in the wiki using thsi script.

PR #584, opened 2024-08-09
themes/ht: Added ht theme

HarrisonGagnon:


themes/ht/ht.theme.sh
```diff
@@ -0,0 +1,80 @@
...
+# Harrison's Theme (ht)
+
+# Description:
+# Simple prompt that shows basic working directory and exit code info
+# Works with 8 color terminals
+
+# Example prompt:
+# ● [harrison@ubN2] in src ± |dev ✗|
+
+function exit_color {
```
akinomyoga, line 12: Can we change the function name `exit_color` to `_omb_theme_ht_exit_color`?

akinomyoga (commented):
Thanks!

themes/ht/ht.theme.sh
```diff
@@ -0,0 +1,80 @@
...
+  PS1+="${user_host_prefix}${user}${at}${host}${user_host_suffix}"
+
+  # Working directory
+  PS1+=" in ${_omb_prompt_magenta}\W${_omb_prompt_reset_color}"
+
+  # SCM
+  local SCM_GIT_SHOW_MINIMAL_INFO="true"
+  local SCM_THEME_PROMPT_DIRTY=" ${_omb_prompt_red}✗${_omb_prompt_normal}"
+  local SCM_THEME_PROMPT_CLEAN=" ${_omb_prompt_green}✓${_omb_prompt_normal}"
+  local SCM_NONE_CHAR=""
```
akinomyoga, line 72: Can we set these variables globally so that the users of `ht` can override them?

themes/ht/ht.theme.sh
```diff
@@ -0,0 +1,82 @@
...
+
+  # Start prompt blank
+  PS1=""
+
+  # Exit code indicator
+  PS1+="$(_omb_theme_ht_exit_color "$EXIT_CODE")●$_omb_prompt_reset_color"
+
+  # Environment info
+  PS1+=" $_omb_prompt_red$(_omb_prompt_get_ruby_env)$_omb_prompt_reset_color"
+  PS1+=" $_omb_prompt_green$(_omb_prompt_get_python_env)$_omb_prompt_reset_color"
```
akinomyoga, line 61: @HarrisonGagnon I tried `ht` in my environment, but I noticed that the appearance is slightly different from the presented picture:

> 

In my environment, it looks like

image

A large space exists between `●` and `[user@host]`.

This is caused by the above parts for the Ruby and Python environments. They add two spaces regardless of whether we are currently in the Ruby or Python environment. However, in the presence of the Ruby or Python environment, `_omb_prompt_get_ruby_env` and `_omb_prompt_get_python_env` are already configured to include a space at the beginning by default:

https://github.com/ohmybash/oh-my-bash/blob/f77615fe938f2809bcd6016bf34ddd793d226664/lib/omb-prompt-base.sh#L61-L90

akinomyoga: Can I remove the spaces?

```suggestion
  # Environment info
  PS1+="$_omb_prompt_red$(_omb_prompt_get_ruby_env)$_omb_prompt_reset_color"
  PS1+="$_omb_prompt_green$(_omb_prompt_get_python_env)$_omb_prompt_reset_color"
```

Instead, I think I need to add a space to the next `PS1+=`.

akinomyoga:
Thanks

akinomyoga (approved):

PR #588, opened 2024-08-20
plugins/cargo: Added a new plugin for working with cargo

leonovk:
Added a plugin for working with cargo, made aliases for cargo itself and also for the most frequently used commands.

PR #590, opened 2024-08-24
plugins/rbenv: Added a new plugin for working with rbenv

leonovk:
Added a new plugin for working with rbenv. It registers the necessary paths, initializes the environment, and sets a useful alias.

### DIFF
--- a/plugins/rbenv/rbenv.plugin.sh
+++ b/plugins/rbenv/rbenv.plugin.sh
@@ -1,0 +1,9 @@
+#! bash oh-my-bash.module
+
+if [ -d ~/.rbenv ]; then
+  export RBENV_ROOT=~/.rbenv
+	export PATH=$RBENV_ROOT/bin:$PATH
+  eval "$(rbenv init - bash)"
+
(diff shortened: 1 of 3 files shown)


plugins/rbenv/rbenv.plugin.sh
```diff
@@ -0,0 +1,9 @@
+#! bash oh-my-bash.module
+
+if [ -d ~/.rbenv ]; then
```
akinomyoga, line 3:
```suggestion
if [[ -d ~/.rbenv ]]; then
```

akinomyoga (approved):

akinomyoga:
Thank you. I merged this PR.

By the way, I noticed that your previous PR #588, which added `plugins/cargo`, actually only contains aliases. Maybe it should have been `aliases/cargo`. What do you think?

leonovk:
@akinomyoga Yes, you are right. A little later I will move cargo to aliases

akinomyoga:
Thank you!

PR #593, opened 2024-08-26
plugins/asdf: Add `asdf` extendable version manager plugin

RobLoach:
This adds a plugin to load the [asdf version manager](https://github.com/asdf-vm/asdf) through *oh-my-bash*. It's roughly based off of the [original pull request](https://github.com/ohmybash/oh-my-bash/pull/310), moves to the bash syntax, and allows checking for the `ASDF_DIR` variable before hitting the file system.

```
asdf plugin add nodejs
asdf install nodejs latest
```

### DIFF
--- a/completions/asdf.completion.sh
+++ b/completions/asdf.completion.sh
@@ -3,6 +3,6 @@
 # Depends on the `asdf` plugin.
 
 # Only load the completions if the ASDF_DIR variable was set.
-if [[ -v ASDF_DIR ]]; then
+if [[ ${ASDF_DIR+set} ]]; then
   . "$ASDF_DIR/completions/asdf.bash"
 fi
--- a/completions/asdf.completion.sh
+++ b/completions/asdf.completion.sh
@@ -1,0 +1,8 @@
+#! bash oh-my-bash.module
+# Bash completion support for the `asdf` command.
+# Depends on the `asdf` plugin.
+
+# Only load the completions if the ASDF_DIR variable was set.
+if [[ -v ASDF_DIR ]]; then
+  . "$ASDF_DIR/completions/asdf.bash"
+fi
--- a/plugins/asdf/README.md
+++ b/plugins/asdf/README.md
@@ -19,7 +19,7 @@ Add [oh-my-bash](https://ohmybash.github.io) integration with [asdf](https://git
     completions=(asdf)
     ```
 
-### Usage
+## Usage
 
 See the [asdf usage documentation](https://github.com/asdf-vm/asdf#usage) for information on how to use asdf:
 
--- a/plugins/asdf/README.md
+++ b/plugins/asdf/README.md
@@ -2,7 +2,7 @@
 
 Add [oh-my-bash](https://ohmybash.github.io) integration with [asdf](https://github.com/asdf-vm/asdf), the extendable version manager, with support for Ruby, Node.js, Elixir, Erlang and more.
 
-### Installation
+## Installation
 
 1. [Install asdf](https://github.com/asdf-vm/asdf#setup) by running the following:
     ``` bash
--- a/plugins/asdf/README.md
+++ b/plugins/asdf/README.md
@@ -1,0 +1,29 @@
+# asdf plugin
+
+Add [oh-my-bash](https://ohmybash.github.io) integration with [asdf](https://github.com/asdf-vm/asdf), the extendable version manager, with support for Ruby, Node.js, Elixir, Erlang and more.
+
+### Installation
+
+1. [Install asdf](https://github.com/asdf-vm/asdf#setup) by running the following:
+    ``` bash
+    git clone https://github.com/asdf-vm/asdf.git ~/.asdf
+    ```
+
+2. Enable the plugin by adding it to your oh-my-bash `plugins` definition in `~/.bashrc`.
+    ``` sh
+    plugins=(asdf)
+    ```
+
+2. Enable the completions by adding it to your oh-my-bash `completions` definition in `~/.bashrc`.
+    ``` sh
+    completions=(asdf)
+    ```
+
+### Usage
+
+See the [asdf usage documentation](https://github.com/asdf-vm/asdf#usage) for information on how to use asdf:
+
+``` bash
+asdf plugin add nodejs
+asdf install nodejs latest
+```
--- a/plugins/asdf/asdf.plugin.sh
+++ b/plugins/asdf/asdf.plugin.sh
@@ -1,0 +1,26 @@
+#! bash oh-my-bash.module
+# asdf.plugin.sh: asdf plugin for oh-my-bash.
+
+# Custom ASDF_DIR location
+if [[ -v ASDF_DIR ]]; then
+  . "$ASDF_DIR/asdf.sh"
+
+# Standard install location
+elif [[ -f "${XDG_CONFIG_HOME:-$HOME}/.asdf/asdf.sh" ]]; then
+  ASDF_DIR="${XDG_CONFIG_HOME:-$HOME}/.asdf"
+  . "${ASDF_DIR}/asdf.sh"
+
+# Arch Linux / AUR Package
+elif [[ -f "/opt/asdf-vm/asdf.sh" ]]; then
+  ASDF_DIR="/opt/asdf-vm"
+  . /opt/asdf-vm/asdf.sh
+
+# Homebrew
+elif _omb_util_command_exists brew; then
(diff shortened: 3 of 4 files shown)


completions/asdf.completion.sh
```diff
@@ -0,0 +1,8 @@
+#! bash oh-my-bash.module
+# Bash completion support for the `asdf` command.
+# Depends on the `asdf` plugin.
+
+# Only load the completions if the ASDF_DIR variable was set.
+if [[ -v ASDF_DIR ]]; then
```
akinomyoga, line 6: Oh My Bash targets Bash >= 3.2 while `[[ -v ASDF_DIR ]]` is only available in Bash >= 4.2.

```suggestion
if [[ ${ASDF_DIR+set} ]]; then
```

plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,26 @@
+#! bash oh-my-bash.module
+# asdf.plugin.sh: asdf plugin for oh-my-bash.
+
+# Custom ASDF_DIR location
+if [[ -v ASDF_DIR ]]; then
```
akinomyoga, line 5: Similar to the previous comment.

plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,26 @@
...
+  . "${ASDF_DIR}/asdf.sh"
+
+# Arch Linux / AUR Package
+elif [[ -f "/opt/asdf-vm/asdf.sh" ]]; then
+  ASDF_DIR="/opt/asdf-vm"
+  . /opt/asdf-vm/asdf.sh
+
+# Homebrew
+elif _omb_util_command_exists brew; then
+  __ASDF_PREFIX="$(brew --prefix asdf)"
```
akinomyoga, line 20: Is this an internal variable of this plugin? Can we change the variable name as follows?

```suggestion
  _omb_plugin_asdf__prefix="$(brew --prefix asdf)"
```

If so, the other occurrences of `__ASDF_PREFIX` should also be updated.

plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,26 @@
...
+  if [[ -f "$__ASDF_PREFIX/libexec/asdf.sh" ]]; then
+    ASDF_DIR="$__ASDF_PREFIX/libexec"
+    . "$ASDF_DIR/asdf.sh"
+  fi
+  unset __ASDF_PREFIX
```
akinomyoga, line 25: We want to explicitly specify `-v` to remove the variable (but not a function of the same name).

```suggestion
  unset -v __ASDF_PREFIX
```

plugins/asdf/README.md
```diff
@@ -0,0 +1,29 @@
+# asdf plugin
+
+Add [oh-my-bash](https://ohmybash.github.io) integration with [asdf](https://github.com/asdf-vm/asdf), the extendable version manager, with support for Ruby, Node.js, Elixir, Erlang and more.
+
+### Installation
```
akinomyoga, line 5: I guess it must be heading level 2. The same for `### Usage`.

```suggestion
## Installation
```

akinomyoga (changes requested):
> It's roughly based off of the [original pull request](https://github.com/ohmybash/oh-my-bash/pull/310), moves to the bash syntax, and allows checking for the `ASDF_DIR` variable before hitting the file system.

Thank you. Could you mention the original author (https://github.com/prodrigues1912) and URL of  PR #310 in the code comment around the top of `asdf.plugin.sh`? Since it will not remain in the git history, we need to leave the information manually.

RobLoach:
Thanks a lot for the review, I'll fix these up.

plugins/asdf/asdf.plugin.sh
RobLoach:
``` suggestion
if [[ -v ${ASDF_DIR+set} ]]; then
```

plugins/asdf/README.md
akinomyoga: > The same for `### Usage`.

plugins/asdf/asdf.plugin.sh
akinomyoga: > If so, the other occurrences of `__ASDF_PREFIX` should also be updated.

plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,26 @@
+#! bash oh-my-bash.module
+# asdf.plugin.sh: asdf plugin for oh-my-bash.
```
RobLoach, line 2:
``` suggestion
# asdf.plugin.sh: asdf plugin for oh-my-bash.
# Author: Rob Loach (https://robloach.net)
# Author: prodrigues1912 (https://github.com/prodrigues1912)
# Originally suggested in https://github.com/ohmybash/oh-my-bash/pull/310
# Fork of the oh-my-zsh asdf plugin
```

plugins/asdf/README.md
```diff
@@ -0,0 +1,29 @@
...
+    ``` sh
+    plugins=(asdf)
+    ```
+
+2. Enable the completions by adding it to your oh-my-bash `completions` definition in `~/.bashrc`.
+    ``` sh
+    completions=(asdf)
+    ```
+
+### Usage
```
RobLoach, line 22:
``` suggestion
## Usage
```

plugins/asdf/asdf.plugin.sh
RobLoach: Good catch, thanks. Fixed up.

plugins/asdf/asdf.plugin.sh
RobLoach: Good catch, thanks. Fixed.

akinomyoga: Thanks

plugins/asdf/asdf.plugin.sh
akinomyoga: Thanks.

plugins/asdf/asdf.plugin.sh
```diff
@@ -0,0 +1,30 @@
...
+# Originally suggested in https://github.com/ohmybash/oh-my-bash/pull/310
+# Fork of the oh-my-zsh asdf plugin
+
+# Custom ASDF_DIR location
+if [[ ${ASDF_DIR+set} ]]; then
+  . "$ASDF_DIR/asdf.sh"
+
+# Standard install location
+elif [[ -f "${XDG_CONFIG_HOME:-$HOME}/.asdf/asdf.sh" ]]; then
+  ASDF_DIR="${XDG_CONFIG_HOME:-$HOME}/.asdf"
```
akinomyoga, line 14: Is this location correct? `XDG_CONFIG_HOME` is defaulted to `$HOME/.config`. It's not normal to have a dotdir under `~/.config` (i.e. `~/.config/.asdf`). The global code search in GitHub only finds only your setting:

- [Code search results `/${XDG_CONFIG_HOME:-$HOME}\/.asdf/`](https://github.com/search?q=%2F%24%7BXDG_CONFIG_HOME%3A-%24HOME%7D%5C%2F.asdf%2F&amp;type=code)

The upstream discussions (https://github.com/asdf-vm/asdf/issues/687, https://github.com/asdf-vm/asdf/pull/1351) .only mention `$XDG_CONFIG_HOME/asdf`

akinomyoga (commented):
Thank you for updating

plugins/asdf/asdf.plugin.sh
RobLoach: Perhaps something like this?
``` suggestion
# Home
elif [[ -f "$HOME/.asdf/asdf.sh" ]]; then
  ASDF_DIR="$HOME/.asdf"
  . "${ASDF_DIR}/asdf.sh"

# Config
elif [[ -f "$XDG_CONFIG_HOME/asdf/asdf.sh" ]]; then
  ASDF_DIR="$XDG_CONFIG_HOME/asdf"
```

akinomyoga: That works, but please replace `$XDG_CONFIG_HOME` with `${XDG_CONFIG_HOME:-$HOME/.config}`.

akinomyoga (approved):

akinomyoga:
Thank you.

RobLoach:
Thank you so much for the reviews and merge :+1: ... Happy to have this in.

PR #600, opened 2024-09-04
Enable THEME_CLOCK_FORMAT support from .bashrc

elg:
powerline-icon has a hardcoded time format while apparently THEME_CLOCK_FORMAT is the way to go. If not set, it seems to still be %H%M%S or something like that so there is nothing broken here.


RobLoach (approved):
This approach is used elsewhere, so should be good.

themes/powerline-icon/powerline-icon.base.sh
```diff
@@ -32,7 +32,7 @@ function __powerline_user_info_prompt {
       fi
       ;;
   esac
-  [[ -n "${user_info}" ]] && echo "$OMB_THEME_POWERLINE_ICON_USER ${user_info} `date +%X\ %D`|${color}|${secondary_color}"
+  [[ -n "${user_info}" ]] && echo "$OMB_THEME_POWERLINE_ICON_USER ${user_info} `date +${THEME_CLOCK_FORMAT}`|${color}|${secondary_color}"
```
akinomyoga, line 35: I suggest using a new variable name `OMB_THEME_POWERLINE_ICON_CLOCK`.

```suggestion
  if [[ -n "${user_info}" ]]; then
    local clock=$(date +"${OMB_THEME_POWERLINE_ICON_CLOCK-%X %D}")
    _omb_util_print "$OMB_THEME_POWERLINE_ICON_USER ${user_info}${clock:+ $clock}|${color}|${secondary_color}"
  fi
```

Also, this new variable should be documented in README.md. However, I now noticed that `powerline-icon` doesn't have a README...

akinomyoga:
rebased and resolved a conflict.

themes/powerline-icon/powerline-icon.base.sh
akinomyoga: > I suggest using a new variable name `OMB_THEME_POWERLINE_ICON_CLOCK`.

https://github.com/ohmybash/oh-my-bash/pull/600/commits/f8895234f3f3bebac3086ac4eaf2dd9cee67e410

> Also, this new variable should be documented in README.md. However, I now noticed that `powerline-icon` doesn't have a README...

I added README: https://github.com/ohmybash/oh-my-bash/pull/600/commits/154157c3e9860e264f6aee13e1dc927a43eb8458

PR #602, opened 2024-09-05
aliases: Move alias plugins to aliases

RobLoach:
This moves all the remaining plugins that just implement aliases to the aliases directory.


akinomyoga:
Thank you for these suggestions!

There is a problem that is difficult for me to decide. Although I'd like to move these plugins to aliases as suggested, there is a compatibility issue. The existing users who have already enabled these plugins will see errors. In addition, for the users who've forked a branch and modified the plugins, it would be more non-trivial to merge the changes. Technically, we can inform such users to fix their `.bashrc` and merge conflicts to make it consistent with our changes. However, I'm not sure if it would be worth moving these plugins to aliases at the expense of breaking existing users' settings and requiring the users to fix the settings.

The reason that I suggested moving "plugin `cargo`" to "aliases `cargo`" was because it is a new plugin just added a few weeks ago and disabled by default. Therefore, we expect there are not so many existing users of the plugin `cargo`.

A possible workaround would be to leave symbolic links to the new locations in the plugin directories. However, I'm wondering if such a complication is needed at all.

Honestly, there are many inconsistencies in this project, and the present inconsistency of plugins/aliases is merely one of them. To solve them, all kinds of destructive changes in Oh My Bash are needed. Continuously breaking compatibilities doesn't seem nice, but it requires a large amount of work to solve everything all at once. Even if it is done all at once, we would anticipate the questions and reports from the users who are troubled by the incompatible changes. I currently don't have the bandwidth to handle these issues. What do you think about the compatibility?

RobLoach:
I have no strong opinion either way. Backwards compatibility is a good reason to keep them where they are.

Would you like me to go through all the alias plugins and bring them into one pull request?

aliases/brew.aliases.md
```diff
@@ -1,14 +1,6 @@
-# brew plugin
+# Aliases `brew`
```
RobLoach, line 1:
``` suggestion
# Aliases: `brew`
```

akinomyoga:
Thank you for updating, but I forgot to mention another point to consider. I'm sorry, I should have commented on that earlier.

Since this project is a Bash version of Oh My Zsh (OMZ) and is actually derived from OMZ, we want to keep some consistency with OMZ even if we do not offer perfect parity with OMZ. This is because the OMB users are typically also the users of OMZ and use OMB to set up a similar environment as their environment in Zsh with OMZ. The upstream OMZ doesn't have a category "aliases", and everything including aliases is inside "plugins". The `golang` and `kubectl` plugins that this PR tries to move have corresponding files in the upstream OMZ. We wouldn't like to break this correspondence.

- [ohmyzsh/plugins/golang/golang.plugin.zsh at master · ohmyzsh/ohmyzsh](https://github.com/ohmyzsh/ohmyzsh/blob/80fa5e137672a529f65a05e396b40f0d133b2432/plugins/golang/golang.plugin.zsh)
- [ohmyzsh/plugins/kubectl/kubectl.plugin.zsh at master · ohmyzsh/ohmyzsh](https://github.com/ohmyzsh/ohmyzsh/blob/80fa5e137672a529f65a05e396b40f0d133b2432/plugins/kubectl/kubectl.plugin.zsh)

OMZ also seems to have the `brew` plugin, but the upstream `brew` seems different from ours, so it's actually better to move not to confuse the users of both OMB/OMZ:

- [ohmyzsh/plugins/brew/brew.plugin.zsh at master · ohmyzsh/ohmyzsh](https://github.com/ohmyzsh/ohmyzsh/blob/80fa5e137672a529f65a05e396b40f0d133b2432/plugins/brew/brew.plugin.zsh)

Another upstream of this project is Bash-it. The category `aliases` originates in Bash-it. Bash-it seems to have `kubectl.aliases.bash`, but its contents appear to show that it has a different origin.

- [bash-it/aliases/available/kubectl.aliases.bash at master · Bash-it/bash-it](https://github.com/Bash-it/bash-it/blob/d985e4c96b1eb4e7374b3ab53eea3513fc9f50ce/aliases/available/kubectl.aliases.bash)

RobLoach:
Didn't know parity was desired, thanks for the explanation. Does this sound like a good path forwards?
- Move our Chezmoi to aliases, since neither ohmyzsh or bash-it implement it as a plugin.
- Move our brew plugin to an alias, since the brew of ohmyzsh and bash-it are different.
- Keep the others where they are

Again, I have no strong opinion either way. Just looking to help out where I can.

akinomyoga:
> Didn't know parity was desired, thanks for the explanation. Does this sound like a good path forwards?
> 
> * Move our Chezmoi to aliases, since neither ohmyzsh or bash-it implement it as a plugin.
> * Move our brew plugin to an alias, since the brew of ohmyzsh and bash-it are different.
> * Keep the others where they are

Yes, that is what I'm thinking (if we are going to resolve these).

Anyway, I probably won't merge this PR soon. As explained above, destructive changes should be performed properly. At least we would need to do that when we increment the major version of OMB with other destructive changes.

akinomyoga:
Thank you! LGTM for now. We'll keep this PR until we perform the next major update.

RobLoach:
Thanks, brought these change into a different branch at https://github.com/ohmybash/oh-my-bash/pull/609

akinomyoga:
Thank you for continuing to pay attention to the PR, but may I ask about your intent? It was just fine with this PR, which contains the related discussion.

RobLoach:
I had accidently opened the branch off of `master` instead of creating a branch. We can keep this one instead. I'll just have to remember not to mess with my fork's `master`.

akinomyoga:
Ah, OK. Makes sense. Then, let's leave a reference to here in #609.

PR #603, opened 2024-09-05
aliases/golang: Move golang to aliases

RobLoach:
This moves the golang plugin to an aliases file, since it's all aliases.


akinomyoga:
Discussed in #602

RobLoach:
If we do this, we'll do it over there. Thanks.

PR #604, opened 2024-09-05
aliases/chezmoi: Move chezmoi to aliases

RobLoach:
This moves another aliases file from plugins to aliases.


akinomyoga:
Discussed in #602

RobLoach:
If we do this, we'll do it over there. Thanks.

Issue #470, opened 2023-08-24
Add option to disable all aliases


klimadavid:
I disabled all aliases in the configuration with aliases=(). When I do alias i see that there are aliases like egrep='egrep --color=auto --exclude-dir={.bzr,CVS,.git,.hg,.svn,.idea,.tox}', which I have no idea about. How can I disable these aliases?

It would be nice if I could set a variable like OMB_ALIASES=false, so that i can disable aliases.

Unwanted aliases can lead to people wondering what they are doing wrong when they aren't doing anything wrong, and its the alias they never knew they had causing the problem.

akinomyoga:
Use
```bash
OMB_DEFAULT_ALIASES=disable
```

klimadavid:
Unfortunately it does not work for me and I dont know why:
```
 david  ~  grep ALIAS .bashrc
OMB_DEFAULT_ALIASES="disable"
 david  ~  grep aliases= .bashrc
# Example format: aliases=(vagrant composer git-avh)
aliases=()
 david  ~  alias | grep grep
alias gbda='command git branch --no-color --merged | command grep -vE "^(\*|\s*(master|develop|dev)\s*$)" | command xargs -n 1 git branch -d'
alias gignored='command git ls-files -v | grep "^[[:lower:]]"'
alias gunwip='command git log -n 1 | grep -q -c "\-\-wip\-\-" && git reset HEAD~1'

```

akinomyoga:
> It looks like the aliases that remain come from the git plugin.

Yes. If you don't want these plugins, you also need to remove the `git` plugin from the `plugins` array.

klimadavid:
I need the git plugin, but I don't want aliases... :(

akinomyoga:
In such a case, you can copy the `git` plugin to the custom directory and edit it. The way to copy and edit a plugin is described in *["Customization of Plugins and Themes" (README)](https://github.com/ohmybash/oh-my-bash#customization-of--plugins-and-themes)*.

PR #605, opened 2024-09-05
aliases/git: Move git plugin aliases to their own file

RobLoach:
In an attempt to fix the issue brought up in #470, this moves the aliases defined in the git plugin to their own git aliases file.

It allows having the git shell functions, while keeping the git aliases disabled.


akinomyoga:
Hmm, I'm opposed to this change. A significant number of users use this plugin since it is enabled by default in Oh My Bash (OMB). So the compatibility issue that I raised in #602 applies here even more . Even if we forget about the compatibility issue, this breaks correspondence to the upstream OMZ plugin [`git`](https://github.com/ohmyzsh/ohmyzsh/blob/80fa5e137672a529f65a05e396b40f0d133b2432/plugins/git/git.plugin.zsh).

Also, many of the left functions are actually helper functions used to implement the aliases. These aliases depend on shell functions, such as `git_{current,main,develop}_branch`. I also suspect that the other functions are for similar purposes. In that sense, this move of the partial content breaks the aliases relying on them.

RobLoach:
Sounds good with me :+1:

PR #607, opened 2024-09-09
defaults.completion.sh Syntax Error on macOS with Bash version 3.2.57

cyberjames:
After updating Oh My Bash, I encountered a syntax error in the `defaults.completion.sh` file, specifically at line 69. 
The error message is:

```bash
-bash: /Users/username/.oh-my-bash/completions/defaults.completion.sh: line 69: syntax error in conditional expression: unexpected token `('
-bash: /Users/username/.oh-my-bash/completions/defaults.completion.sh: line 69: syntax error near `@($'
-bash: /Users/username/.oh-my-bash/completions/defaults.completion.sh: line 69: `  if [[ ${COMP_WORDS[1]} == @(${cmds// /|}) ]]; then'
```

This error seems to be related to the use of `@(...)` syntax, which might not be fully supported in my Bash version.

**Steps to Reproduce**

1. Update Oh My Bash to the latest version.
2. Open a new terminal session.
3. The error message appears.

**Expected Behavior**

Oh My Bash should load without errors, and tab completion should work correctly.

**Actual Behavior**

The syntax error prevents Oh My Bash from loading properly, and tab completion might be affected.

**Environment**

* **Operating System:** macOS Sonoma 14.6.1
* **Bash Version:** 3.2.57(1)-release
* **Oh My Bash Version:** (latest)

**Workaround**

I was able to temporarily fix the issue by reverting commit `74012c548c2fbd3f38a9c04dbf4e2cb3e38493b1`.

**Possible Solution**

* Consider using a more compatible syntax in `defaults.completion.sh` or adding a version check to avoid using features that might not be supported in older Bash versions.

### DIFF
--- a/completions/defaults.completion.sh
+++ b/completions/defaults.completion.sh
@@ -66,7 +66,7 @@ function _defaults {
 
   # Both a domain and command have been specified
 
-  if [[ ${COMP_WORDS[1]} == @(${cmds// /|}) ]]; then
+  if [[ ${COMP_WORDS[1]} == [${cmds// /|}] ]]; then
     cmd=${COMP_WORDS[1]}
     domain=${COMP_WORDS[2]}
     key_index=3
@@ -78,7 +78,7 @@ function _defaults {
       domain="-app ${COMP_WORDS[3]}"
       key_index=4
     fi
-  elif [[ ${COMP_WORDS[2]} == "-currentHost" && ${COMP_WORDS[2]} == @(${cmds// /|}) ]]; then
+  elif [[ ${COMP_WORDS[2]} == "-currentHost" && ${COMP_WORDS[2]} == [${cmds// /|}] ]]; then
     cmd=${COMP_WORDS[2]}
     domain=${COMP_WORDS[3]}
     key_index=4
@@ -90,7 +90,7 @@ function _defaults {
       domain="-app ${COMP_WORDS[4]}"
       key_index=5
     fi
-  elif [[ ${COMP_WORDS[3]} == "-host" && ${COMP_WORDS[3]} == @(${cmds// /|}) ]]; then
+  elif [[ ${COMP_WORDS[3]} == "-host" && ${COMP_WORDS[3]} == [${cmds// /|}] ]]; then
     cmd=${COMP_WORDS[3]}
     domain=${COMP_WORDS[4]}
     key_index=5


akinomyoga:
The reported problem is supposed to be already fixed by commit 1a7ef75.

akinomyoga:
> This error seems to be related to the use of `@(...)` syntax, which might not be fully supported in my Bash version.

Also, the reported error message should never happen in Bash 5.2. The problem is caused by the failure to restore the option `extglob`; the reported construct (the extended glob patterns in conditional commands) fails to be parsed in Bash <= 4.0. However, in Bash >= 4.1, it works regardless of the option `extgob`.

I suspect you are actually using Bash 3.2 shipped with macOS instead of Bash 5.2. How did you check the Bash version? `bash --version` is not the right way to check the Bash version of the current session. The command `bash --version` tells the Bash version in the current `PATH`, which is unrelated to the version of the login shell of the current user. You need to use `echo "$BASH_VERSION"` to check the version of the current Bash session.

```console
$ echo "$BASH_VERSION"
```

If you intended to change your login shell, it's not sufficient to install the new Bash version. You additionally need to add the path to the new Bash in `/etc/shells` and then perform `chsh -s <path to the new Bash>` to change the login shell of the current user account.

> **Workaround**
> 
> I was able to temporarily fix the issue by reverting commit `74012c548c2fbd3f38a9c04dbf4e2cb3e38493b1`.

This is just a workaround. The previous version was just wrong; it didn't do what is supposed to do.

> **Possible Solution**
> 
> * Consider using a more compatible syntax in `defaults.completion.sh` or adding a version check to avoid using features that might not be supported in older Bash versions.

As far as the option `extglob` is turned on, the current syntax has been already compatible since Bash 2.02 (26 years ago). It is still a compatible syntax in the devel version of Bash (5.3-alpha). The problem was just the option `extglob` was unexpectedly turned off.

cyberjames:
Yeah I was using `bash --version`. 

This is what I got:
```bash
$ echo $BASH_VERSION
3.2.57(1)-release
```

akinomyoga:
Thanks for the confirmation of the Bash version! Then, I think this is exactly the issue that was solved by commit 1a7ef75664de1241009f10d5f0fe3d08e699526d. Could you confirm if the problem is fixed with the current `master` branch?

cyberjames:
Thanks @akinomyoga. Unfortunately, I just tested with the latest master branch, and I'm still encountering the same error.
```bash
-bash: /Users/username/.oh-my-bash/completions/defaults.completion.sh: line 69: syntax error in conditional expression: unexpected token `('
-bash: /Users/username/.oh-my-bash/completions/defaults.completion.sh: line 69: syntax error near `@($'
-bash: /Users/username/.oh-my-bash/completions/defaults.completion.sh: line 69: `  if [[ ${COMP_WORDS[1]} == @(${cmds// /|}) ]]; then'
```

akinomyoga:
OK, thanks for the confirmation. I could reproduce the problem.

I'm testing it with Bash 3.2 locally installed in my Linux, but somehow another syntax error caused by the system bash-completion (which requires Bash 4.2) has hidden the syntax error of `defaults.completion.sh`. When I turned off the bash-completion, I could reproduce the problem.

akinomyoga:
I've been thinking we would have the setting `shopt -s extglob` in our codebase, but it turned out to be not the case. We load `bash-completion` if we find it in the system, and `bash-completion` has been turning on `extglob`. This means that in the systems where `bash-completion` is not installed, `extglob` is not turned on and causes the reported issue.

I pushed a fix 08e297c in the master branch. We decided to include `shopt -s extglob` in the default setting. It is anyway turned on in the systems with bash-completion, so it should be harmless. Furthermore, we already enable other non-trivial glob options such as `nocaseglob` and `globstar`, so there doesn't seem to be a reason to refraining from the rather basic option `extglob`.

Could you check if the problem is solved in the current master branch?

cyberjames:
Yep seems the error has gone. That fix works best! Cheers! 🍻

akinomyoga:
Thank you for your quick responses!

PR #609, opened 2024-09-09
aliases: Move `brew` and `chezmoi` plugins to aliases

RobLoach:
This moves the brew and chezmoi plugins to the aliases directory as they just implement aliases.

As this is a breaking change, hold this until next release.


akinomyoga:
The discussion is continued from #602.

PR #619, opened 2024-09-30
Add Colored Man Pages plugin

RobLoach:
Inspired by the [`colored-man-pages`](https://github.com/ohmyzsh/ohmyzsh/tree/a72a26406ad3aa9a47c3f5227291bad23494bed0/plugins/colored-man-pages) zsh plugin.

This is likely not the best design, as I'm grabbing the specific colors. If there's a better method, feel free to let me know.

Screenshot from 2024-09-30 16-53-12

### DIFF
--- a/plugins/colored-man-pages/colored-man-pages.plugin.sh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.sh
@@ -1,0 +1,47 @@
+#! bash oh-my-bash.module
+
+# Safety get the colors.
+if tput setaf 1 &> /dev/null; then
+  COLORED_MAN_PAGES_BLACK=$(tput setaf 0)
+  COLORED_MAN_PAGES_RED=$(tput setaf 1)
+  COLORED_MAN_PAGES_GREEN=$(tput setaf 2)
+  COLORED_MAN_PAGES_YELLOW=$(tput setaf 3)
+  COLORED_MAN_PAGES_BLUE=$(tput setaf 4)
+  COLORED_MAN_PAGES_MAGENTA=$(tput setaf 5)
+  COLORED_MAN_PAGES_CYAN=$(tput setaf 6)
+  COLORED_MAN_PAGES_WHITE=$(tput setaf 7)
+  COLORED_MAN_PAGES_BOLD=$(tput bold)
+  COLORED_MAN_PAGES_RESET=$(tput sgr0)
+
+  # Extra Styles
+  export LESS_TERMCAP_mr=$(tput rev) # Reverse
+  export LESS_TERMCAP_mh=$(tput dim) # Half Bright
+  export LESS_TERMCAP_ZN=$(tput ssubm) # Subscript
+  export LESS_TERMCAP_ZV=$(tput rsubm) # Reverse Superscript
+  export LESS_TERMCAP_ZO=$(tput ssupm) # Subscript
+  export LESS_TERMCAP_ZW=$(tput rsupm) # Reverse Superscript
+else
+  COLORED_MAN_PAGES_BLACK="\033[1;30m"
+  COLORED_MAN_PAGES_RED="\033[1;31m"
+  COLORED_MAN_PAGES_GREEN="\033[1;32m"
+  COLORED_MAN_PAGES_YELLOW="\033[1;33m"
+  COLORED_MAN_PAGES_BLUE="\033[1;34m"
+  COLORED_MAN_PAGES_MAGENTA="\033[1;35m"
+  COLORED_MAN_PAGES_CYAN="\033[1;36m"
+  COLORED_MAN_PAGES_WHITE="\033[1;37m"
+  COLORED_MAN_PAGES_BOLD=""
+  COLORED_MAN_PAGES_RESET="\033[m"
+fi
+
+# Bold and Blinking
+export LESS_TERMCAP_mb=$COLORED_MAN_PAGES_GREEN
+export LESS_TERMCAP_md=$COLORED_MAN_PAGES_BOLD$COLORED_MAN_PAGES_GREEN
+export LESS_TERMCAP_me=$COLORED_MAN_PAGES_RESET
+
+# Standout Mode
+export LESS_TERMCAP_so=$COLORED_MAN_PAGES_YELLOW
+export LESS_TERMCAP_se=$COLORED_MAN_PAGES_RESET
+
+# Underline
+export LESS_TERMCAP_us=$COLORED_MAN_PAGES_MAGENTA
+export LESS_TERMCAP_ue=$COLORED_MAN_PAGES_RESET
--- a/plugins/colored-man-pages/colored-man-pages.plugin.sh
+++ b/plugins/colored-man-pages/colored-man-pages.plugin.sh
@@ -7,17 +7,17 @@
 
 # Adds the LESS_TERMCAP_* variables to color the man pages.
 function colored() {
-  local -x LESS_TERMCAP_mb=${_omb_term_green}
-  local -x LESS_TERMCAP_md=${_omb_term_bold_green}
+  local -x LESS_TERMCAP_mb=${_omb_term_red}
+  local -x LESS_TERMCAP_md=${_omb_term_bold_red}
   local -x LESS_TERMCAP_me=${_omb_term_reset}
   local -x LESS_TERMCAP_so=${_omb_term_bold_yellow}
   local -x LESS_TERMCAP_se=${_omb_term_reset}
-  local -x LESS_TERMCAP_us=${_omb_term_purple}
+  local -x LESS_TERMCAP_us=${_omb_term_green}
   local -x LESS_TERMCAP_ue=${_omb_term_reset}
 
   # Prefer `less` whenever available, since we specifically configured
   # environment for it.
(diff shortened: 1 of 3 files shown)


plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,47 @@
...
+  COLORED_MAN_PAGES_BLACK=$(tput setaf 0)
+  COLORED_MAN_PAGES_RED=$(tput setaf 1)
+  COLORED_MAN_PAGES_GREEN=$(tput setaf 2)
+  COLORED_MAN_PAGES_YELLOW=$(tput setaf 3)
+  COLORED_MAN_PAGES_BLUE=$(tput setaf 4)
+  COLORED_MAN_PAGES_MAGENTA=$(tput setaf 5)
+  COLORED_MAN_PAGES_CYAN=$(tput setaf 6)
+  COLORED_MAN_PAGES_WHITE=$(tput setaf 7)
+  COLORED_MAN_PAGES_BOLD=$(tput bold)
+  COLORED_MAN_PAGES_RESET=$(tput sgr0)
```
akinomyoga, line 14: We already have variables `_omb_term_{black,brown,green,olive,navy,purple,teal,silver,bold,reset}`. We don't want to call a bunch of `tput` for every single plugin. It slows down the initialization. Since those variables `COLORED_MAN_PAGES_*` are not intended to be used by users, we can simply remove these variables. Instead one can directly specify `$_ble_term_*` to `LESS_TERMCAP_*`.

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,47 @@
...
+
+  # Extra Styles
+  export LESS_TERMCAP_mr=$(tput rev) # Reverse
+  export LESS_TERMCAP_mh=$(tput dim) # Half Bright
+  export LESS_TERMCAP_ZN=$(tput ssubm) # Subscript
+  export LESS_TERMCAP_ZV=$(tput rsubm) # Reverse Superscript
+  export LESS_TERMCAP_ZO=$(tput ssupm) # Subscript
+  export LESS_TERMCAP_ZW=$(tput rsupm) # Reverse Superscript
```
akinomyoga, line 22: `tput` doesn't necessarily recognize the terminfo names. In some (non-Linux) systems, `tput` only accepts termcap names. For such a case, could you redirect `stderr` and also attempt a call with the termcap name? You can find examples in

https://github.com/ohmybash/oh-my-bash/blob/f53beb64b8fe8a3f6ad9df09cc42fa5570406def/lib/utils.sh#L150-L160

akinomyoga (changes requested):

plugins/colored-man-pages/colored-man-pages.plugin.sh
RobLoach: Didn't know of those. Will update accordingly. Thanks.

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# Extra Styles
+if _omb_util_binary_exists tput; then
+  export LESS_TERMCAP_mr=$(tput rev 2>/dev/null || tput mr 2>/dev/null)
+  export LESS_TERMCAP_mh=$(tput dim 2>/dev/null || tput mh 2>/dev/null)
+  export LESS_TERMCAP_ZN=$(tput ssubm 2>/dev/null)
+  export LESS_TERMCAP_ZV=$(tput rsubm 2>/dev/null)
+  export LESS_TERMCAP_ZO=$(tput ssupm 2>/dev/null)
+  export LESS_TERMCAP_ZW=$(tput rsupm 2>/dev/null)
+fi
```
RobLoach, line 24: Updated it, not sure if these are done correctly, but using the omb api does clean up a lot of the code. Thanks.

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,24 @@
+#! bash oh-my-bash.module
+
+# Bold and Blinking
+export LESS_TERMCAP_mb=$_omb_term_green
+export LESS_TERMCAP_md=$_omb_term_bold$_omb_term_green
```
akinomyoga, line 5:
```suggestion
export LESS_TERMCAP_md=$_omb_term_bold_green
```
There are actually variables for foreground colors with boldface.

plugins/colored-man-pages/colored-man-pages.plugin.sh
akinomyoga: Why don't they use the termcap names `ZN`, `ZV`, `ZO`, and `ZW` while `mr` and `mh` are used?

akinomyoga (changes requested):
This plugin affects the global behavior of `less` instead of just affecting `man`.

I thought we might need to accept it if the upstream OMZ plugin of the same name were behaving in the same way. However, I checked the upstream plugin and realized that the upstream version correctly implements it using a shell function with the local environment variable so that it only affects `man`.

Could you adjust it so that it doesn't affect the global behavior of `less`?

plugins/colored-man-pages/colored-man-pages.plugin.sh
RobLoach:
``` suggestion
```

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Standout Mode
+export LESS_TERMCAP_so=$_omb_term_olive
+export LESS_TERMCAP_se=$_omb_term_reset
+
+# Underline
+export LESS_TERMCAP_us=$_omb_term_purple
+export LESS_TERMCAP_ue=$_omb_term_reset
+
+# Extra Styles
```
akinomyoga, line 16:
```suggestion
```

akinomyoga:
> Could you adjust it so that it doesn't affect the global behavior of `less`?

The upstream seems to be using `env`, but what I'm thinking is something like this:

```bash
function _omb_plugin_colored_man_pages_colored {
  local -x LESS_TERMCAP_mb=...
  ...

  command "$@"
}

_omb_util_binary_exists man &&
  function man { _omb_plugin_colored_man_pages_colored "$FUNCNAME" "$@"; }
_omb_util_binary_exists debman &&
  function debman { _omb_plugin_colored_man_pages_colored "$FUNCNAME" "$@"; }
...
```

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,44 @@
...
+# Wrapper for man to colorize the output.
+_omb_util_binary_exists man &&
+function man {
+  colored "$FUNCNAME" "$@"
+}
+
+# Wrapper for dman to colorize the output.
+_omb_util_binary_exists dman &&
+function dman {
+  colored dman "$@"
```
RobLoach, line 37:
``` suggestion
  colored "$FUNCNAME" "$@"
```

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,44 @@
...
+}
+
+# Wrapper for debman to colorize the output.
+_omb_util_binary_exists debman &&
+function debman {
+  colored debman "$@"
```
RobLoach, line 43:
``` suggestion
  colored "$FUNCNAME" "$@"
```

plugins/colored-man-pages/colored-man-pages.plugin.sh
```diff
@@ -0,0 +1,42 @@
...
+  local -x LESS_TERMCAP_md=${_omb_term_bold_green}
+  local -x LESS_TERMCAP_me=${_omb_term_reset}
+  local -x LESS_TERMCAP_so=${_omb_term_bold_yellow}
+  local -x LESS_TERMCAP_se=${_omb_term_reset}
+  local -x LESS_TERMCAP_us=${_omb_term_purple}
+  local -x LESS_TERMCAP_ue=${_omb_term_reset}
+
+  # Prefer `less` whenever available, since we specifically configured
+  # environment for it.
+  local -x PAGER="${commands[less]:-$PAGER}"
```
akinomyoga, line 20: Zsh's `commands` doesn't exist in Bash. The associative array `commands` is a special variable of Zsh, and it resolves to the path to the specified command name when an element is accessed. I think something like this should work:
```suggestion
  local -x PAGER=$(type -P less || _omb_util_print "$PAGER")
```

akinomyoga (commented):
Thanks for updates!

I have a question. This plugin and also the upstream one define the function `colored`, but is this something that this plugin intentionally exposes to the user? Is the function `colored` intended to be also directly used by users? It doesn't seem to be the case to me. If so, can we rename the function as `_omb_plugin_colored_man_pages_colored`? We try to avoid polluting the namespace of the commands that the users directly type and run.  What do you think?

akinomyoga (approved):

akinomyoga (approved):

akinomyoga:
@RobLoach Thanks!

PR #620, opened 2024-10-02
Implement `omb theme`

RobLoach:
This introduces a `themes` plugin which allows changing the theme on the fly. This mimics the [themes oh-my-zsh plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/themes).

## Usage

`theme <theme_name>` - Changes the theme to specified theme.

`theme ` - Changes the theme to some random theme.

`lstheme ` - Lists installed themes.


akinomyoga:
Thank you for the PR! I have a question. How is the upstream OMZ `themes` plugin different from `omz theme` [1]?

- [1] https://github.com/ohmyzsh/ohmyzsh/blob/f4423ebd09fbc7670815c3c20cc86c81b7319e5f/lib/cli.zsh#L658-L796

Since this is related to the OMB specific feature for themes, I feel this should be implemented as a subcommand of the `omb` command similar to `omz theme`. However, the upstream OMZ seems to contain quite similar features in two locations, `plugins/themes` and `omz theme`, which appears to be weird to me.

RobLoach:
Namespacing it within `omb` makes so much sense. Thanks for linking the stub.

RobLoach:
Not sure how I feel about manipulating the `.bashrc` file with `omb theme set zork`.

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
...
 }
 function _omb_cmd_theme {
-  echo 'Not yet implemented'
+  case "$1" in
+  list)
+      local -a available_themes
+      _comp_cmd_omb__get_available_themes
+      for theme in "${available_themes[@]}"; do
+        echo "$theme"
+      done
```
akinomyoga, line 27:
```suggestion
      _omb_util_print_lines "${available_themes[@]}"
```

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
   echo 'Not yet implemented'
 }
 function _omb_cmd_theme {
-  echo 'Not yet implemented'
+  case "$1" in
```
akinomyoga, line 21: In case the user sets `set -u`:

```suggestion
  case "${1-}" in
```

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
...
+  list)
+      local -a available_themes
+      _comp_cmd_omb__get_available_themes
+      for theme in "${available_themes[@]}"; do
+        echo "$theme"
+      done
+    ;;
+  use)
+    local theme=$2
```
akinomyoga, line 30:
```suggestion
    local theme=${2-}
```

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
...
+    if [[ -z "$theme" ]]; then
+      echo 'Usage: omb theme use <theme>'
+      return 2
+    fi
+    local -a available_themes
+    _comp_cmd_omb__get_available_themes
+    for i in "${available_themes[@]}"; do
+      if [ "$i" == "$theme" ]; then
+        _omb_module_require_theme "$theme"
```
akinomyoga, line 39: `_omb_module_require_theme` is kind of "include once"; it checks whether the module is already loaded and loads the module only when it hasn't been loaded. So, with this implementation, one cannot switch to the theme that was loaded before.

To handle this, we may need to factorize a part of `_omb_module_require` (`oh-my-bash.sh`) into `_omb_module_search` and use it to obtain the location of the theme file.

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
...
+        return 0
+      fi
+    done
+    echo "Theme '$theme' not found"
```
akinomyoga, line 43:
```suggestion
    _omb_util_print "Theme '$theme' not found"
```

We avoid using `echo` because `echo` is affected by the setting `shopt -s xpg_echo` and is not trustable.

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
...
+  set)
+    echo 'Not yet implemented'
+    ;;
+  *)
+    echo 'Usage: omb theme <command>'
+    echo ''
+    echo 'Available commands:'
+    echo '  list          List all available themes'
+    echo '  set <theme>   Set the given theme in your .bashrc file'
+    echo '  use <theme>   Load the given theme'
```
akinomyoga, line 54:
```suggestion
    _omb_util_print_lines \
      'Usage: omb theme <command>' \
      '' \
      'Available commands:' \
      '  list          List all available themes' \
      '  set <theme>   Set the given theme in your .bashrc file' \
      '  use <theme>   Load the given theme'
```

lib/cli.bash
```diff
@@ -18,7 +18,43 @@ function _omb_cmd_reload {
...
+      local -a available_themes
+      _comp_cmd_omb__get_available_themes
+      for theme in "${available_themes[@]}"; do
+        echo "$theme"
+      done
+    ;;
+  use)
+    local theme=$2
+    if [[ -z "$theme" ]]; then
+      echo 'Usage: omb theme use <theme>'
```
akinomyoga, line 32:
```suggestion
      _omb_util_print 'Usage: omb theme use <theme>' >&2
```

Error messages should be output to fd 2 (`stderr`).

PR #622, opened 2024-10-03
progress: Improve the Progress Plugin

RobLoach:
This change improves the progress plugin by...
- Allows reseting the progress bar to allow running another progress bar after the first one is done
- Replaces the message whitespace so that there's no overlap in messages
- Scopes the variables within the function
- Renames the global variable so that it's namespaced a bit

Here's a good way to test it...
``` bash
progress 0 "Doing something cool!" && progress 40 "Hello?" && progress 60 "Almost done" && progress 100 "Close" && delay
```

### DIFF
--- a/plugins/progress/progress.plugin.sh
+++ b/plugins/progress/progress.plugin.sh
@@ -4,7 +4,7 @@
 # Summary       : Show a progress bar GUI on terminal platform                 #
 # Support       : destro.nnt@gmail.com                                         #
 # Created date  : Aug 12,2014                                                  #
-# Latest Modified date : Aug 13,2014                                           #
+# Latest Modified date : Dec 19,2024                                           #
 #                                                                              #
 ################################################################################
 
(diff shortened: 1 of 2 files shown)


plugins/progress/progress.plugin.sh
```diff
@@ -29,32 +31,57 @@ function delay()
...
+    if [ $_omb_plugin_progress_value -le 65 -a $value -ge 65 ]; then echo -ne "[#############-------------] (65%) $message$whitespace \r"  ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 70 -a $value -ge 70 ]; then echo -ne "[###############-----------] (70%) $message$whitespace \r"  ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 75 -a $value -ge 75 ]; then echo -ne "[#################---------] (75%) $message$whitespace \r"  ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 80 -a $value -ge 80 ]; then echo -ne "[####################------] (80%) $message$whitespace \r"  ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 85 -a $value -ge 85 ]; then echo -ne "[#######################---] (90%) $message$whitespace \r"  ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 90 -a $value -ge 90 ]; then echo -ne "[##########################] (100%) $message$whitespace \r" ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 100 -a $value -ge 100 ];then echo -ne 'Done!                                            \n' ; delay; fi;
```
RobLoach, line 84: Regarding `printf`, I was getting `bash: [: too many arguments` with printf. I believe this broke last month. We should likely switch to using `_omb_util_print` instead.

akinomyoga: Ah, yeah. `(0%)` should be `(0%%)` in the printf format.

plugins/progress/progress.plugin.sh
```diff
@@ -29,32 +31,57 @@ function delay()
...
+    fi
+
+    # Reset the progress value
+    if [ $value -eq 0 ]; then
+      _omb_plugin_progress_value=0;
+      return 0
+    fi
+
+    # Clear the message whitespace
+    local size=40-${#message}
```
akinomyoga, line 59:
```suggestion
    local size=$((40-${#message}))
```
It doesn't need to be re-calculated every time `i<size` is evalauted.

plugins/progress/progress.plugin.sh
```diff
@@ -29,32 +31,57 @@ function delay()
...
+    local whitespace=""
+    for ((i=0; i<size; i++)); do
+      whitespace="$whitespace "
+    done
+
+    if [ $_omb_plugin_progress_value -le 0 -a $value -ge 0 ]  ; then echo -ne "[--------------------------] (0%)  $message$whitespace \r"  ; delay; fi;
```
akinomyoga, line 65: What is the purpose of the padding by `$whitespace`?

RobLoach: To clear out any previous message text so it doesn't overlap.

If you have Hello World on one progress. And then Yo on the second, the message ends up being "Yollo World"

akinomyoga: OK, in this case we usually put EL [`tput el` (terminfo), `tput ce` (termcap), or `\e[K` (ANSI)].

RobLoach: What I did is absolutely gross.Mind having a look? lol

plugins/progress/progress.plugin.sh
```diff
@@ -29,32 +31,73 @@ function delay()
...
+    local clear_line
+    if _omb_util_command_exists 'tput'; then
+      clear_line=$(tput el)
+      if [ -z $clear_line ]; then
+        clear_line=$(tput el1)
+        if [ -z $clear_line ]; then
+          clear_line=$(tput ce)
+        fi
+      fi
+    fi
```
akinomyoga, line 68: You can simply use 

```suggestion
    clear_line=$(tput el 2>/dev/null || tput ce 2>/dev/null)
```

When the terminfo names are not recognized, `tput` outputs the error message. So the redirection `2>/dev/null` is anyway needed even if the existence of the command `tput` is checked.

Also, `el1` does the opposite. It erases the content from the beginning of the current line to the current position. In this context, the cursor is located at the beginning of the line, so `el1` only erases the first character of the current line. This cannot be used to achieve the present purpose.

plugins/progress/progress.plugin.sh
```diff
@@ -29,32 +31,73 @@ function delay()
...
+    if [ -z $clear_line ]; then
+      clear_line="\e[K"
```
akinomyoga, line 70: `clear_line="\e[K"` will display the literal string `\e[K` in the terminal since the argument of `printf` is not processed as the format string of `printf`.
```suggestion
      clear_line=$'\e[K'
```

plugins/progress/progress.plugin.sh
```diff
@@ -29,32 +31,73 @@ function delay()
...
+    if [ $_omb_plugin_progress_value -le 85 -a $value -ge 85 ]; then printf "%s[#################...] (85%%) %s\r" "$clear_line" "$message" ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 90 -a $value -ge 90 ]; then printf "%s[##################..] (90%%) %s\r" "$clear_line" "$message" ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 95 -a $value -ge 95 ]; then printf "%s[###################.] (95%%) %s\r" "$clear_line" "$message" ; delay; fi;
+    if [ $_omb_plugin_progress_value -le 100 -a $value -ge 100 ]; then
+      # Display the finished progress bar, and then clear with a new line.
+      printf "%s[####################] (100%%) %s\r" "$clear_line" "$message"
+      delay
+      printf "%s\n" "$clear_line"
+      value=0
+    fi;
```
akinomyoga, line 100: The message `Done!` seems to have been dropped. Is this an intentional behavioral change? What is the rationale?

plugins/progress/progress.plugin.sh
akinomyoga: I left comments in new conversations.

plugins/progress/progress.plugin.sh
RobLoach: It's kind of clear when it's done, because the application continues?

akinomyoga: That explains that the new behavior works fine, but it is the same for the previous behavior. Rather, the previous behavior was even more clear about that. Also, does that "?" mean you don't remember why you did it? So, can I think it was not the intentional behavioral change? I avoid breaking changes unless there is a clear reason. Could you revert the behavior?

akinomyoga (commented):
Another thing that this PR changes without explanation is the accelerating growth of the progress bar. Now it is linear. What is the reason for this change? Did you remove that knowing the original intent? Or did you just remove it because you don't understand?

plugins/progress/progress.plugin.sh
RobLoach: When comparing other progress bar approaches, like that in apt, I found that it didn't present a Done! For each progress bar that was displayed. Tried to mimic that behavior.

The removal was intentional, but I'm happy to restore it.

RobLoach:
Thanks for following up on this. Completely forgot about it.

RobLoach:
Tested it out, and it works really well. Thanks so much :rocket:

akinomyoga (approved):

akinomyoga (approved):

akinomyoga:
Thanks!

PR #625, opened 2024-10-04
feat: Added Gradle plugin

chriptus13:
Added Gradle plugin similar to what we have in oh-my-zsh


plugins/gradle/README.md
```diff
@@ -0,0 +1,25 @@
+# Gradle plugin
+
+This plugin adds completions and aliases for [Gradle](https://gradle.org/).
```
akinomyoga, line 3: This plugin doesn't seem to provide completions. Is there a plan to add them?

plugins/gradle/gradle.plugin.sh
```diff
@@ -0,0 +1,24 @@
...
+  while [[ "$dir" != / ]]; do
+    if [[ -x "$dir/gradlew" ]]; then
+      project_root="$dir"
+      break
+    fi
+    dir="${dir:h}"
+  done
+
+  # if gradlew found, run it instead of gradle
+  if [[ -f "$project_root/gradlew" ]]; then
```
akinomyoga, line 16:
```suggestion
  if [[ -x "$project_root/gradlew" ]]; then
```

plugins/gradle/gradle.plugin.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "executing gradlew instead of gradle"
```
akinomyoga, line 17: In case the user pipes the commands, I think this type of informative message should go to stderr. What do you think?

```suggestion
    echo "executing gradlew instead of gradle" >&2
```

plugins/gradle/gradle.plugin.sh
```diff
@@ -0,0 +1,24 @@
+#! bash oh-my-bash.module
+
+function gradle-or-gradlew() {
+  # find project root
+  # taken from https://github.com/gradle/gradle-completion
```
akinomyoga, line 5: That code seems to be licensed under the MIT license, meaning that [the original copyright notice](https://github.com/gradle/gradle-completion/blob/25da917cf5a88f3e58f05be3868a7b2748c8afe6/LICENSE#L1) should be retained. Please include it.

akinomyoga:
> similar to what we have in oh-my-zsh

I checked the upstream OMZ plugin and realized that this PR is actually not just similar but based on the OMZ plugin. In that case, could you also mention the upstream OMZ plugin in the code comment? Since the upstream is a moving target, it is better to keep the permalink that ours is based on:

https://github.com/ohmyzsh/ohmyzsh/blob/62cf1201b031399e7251abeee859e895ee825a48/plugins/gradle/gradle.plugin.zsh

PR #626, opened 2024-10-09
grep - fix option availability check.

allan-kg:
# Description
Checks if the desired _grep_ option is available, properly.

# Cause
Executing `: | grep "$1" "" >/dev/null 2>&1` **always** return `false` on _bash_ `5.2.15(1)-release (x86_64-pc-linux-gnu)` @ _linux_ `6.1.0-25-amd64`.

It causes problems like:
- no color for matches
- including undesired directories

### DIFF
--- a/lib/grep.sh
+++ b/lib/grep.sh
@@ -2,7 +2,7 @@
 
 # is x grep argument available?
 function _omb_grep_flag_available {
-  : | grep "$1" "" >/dev/null 2>&1
+  grep "$1" --help >/dev/null 2>&1
 }
 
 _omb_grep_options=()
--- a/lib/grep.sh
+++ b/lib/grep.sh
@@ -2,7 +2,7 @@
 
 # is x grep argument available?
 function _omb_grep_flag_available {
-  grep "$1" --help >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


lib/grep.sh
```diff
@@ -2,7 +2,7 @@
 
 # is x grep argument available?
 function _omb_grep_flag_available {
-  : | grep "$1" "" >/dev/null 2>&1
+  grep "$1" --help >/dev/null 2>&1
```
akinomyoga, line 5: `--help` is not specified by POSIX, so it fails in some other environments. I'd rather suggest the original version before the bug is introduced.

```suggestion
  echo | grep "$1" "" >/dev/null 2>&1
```

akinomyoga (changes requested):
Thank you for your contribution. This is a regression of [`4b95670`](https://github.com/ohmybash/oh-my-bash/commit/4b9567033052f0e81537ec51f8c58d0e5b0cfaab#diff-c439f098bf3a3a324cfa4eb4cdd1d0843bb7dab8f56f3592a3d72d250f33d64f).

akinomyoga (approved):

PR #628, opened 2024-10-14
[Plugin] adding starship

### LINKED ISSUE #627: [Plugin] Support to Starship
https://starship.rs/

Heus-Sueh:
close my own issue: https://github.com/ohmybash/oh-my-bash/issues/627

I believe I didn't do anything wrong in the code

----

Closes #627

### DIFF
--- a/plugins/starship/starship.plugin.sh
+++ b/plugins/starship/starship.plugin.sh
@@ -1,0 +1,8 @@
+#! bash oh-my-bash.module
+
+# Check if starship is installed
+if _omb_util_command_exists starship; then
+    eval "$(starship init bash)"
+else
+    echo '[oh-my-bash] starship not found, please install it from https://github.com/starship/starship'
+fi
--- a/plugins/starship/starship.plugin.sh
+++ b/plugins/starship/starship.plugin.sh
@@ -3,7 +3,7 @@
 # Check if starship is installed
 if _omb_util_command_exists starship; then
     # ignore oh-my-bash theme
-    unset OSH_THEME
+    unset -v OSH_THEME
     eval "$(starship init bash)"
 else
     echo '[oh-my-bash] starship not found, please install it from https://github.com/starship/starship'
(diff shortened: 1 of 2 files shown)


akinomyoga:
`starship` conflicts with `OSH_THEME` since both provide `PS1`. Shouldn't this be implemented as a theme?

akinomyoga:
In addition, our themes are supposed to be able to switch to another theme when requested, i.e. a theme can be unloaded and another theme can be loaded. Starship can be enabled by `eval "$(starship init bash)"`, but is there a way to unload Starship in a Bash session where Starship was loaded?

Heus-Sueh:
> `starship` conflicts with a theme specified by `OSH_THEME` since both provide `PS1`. Shouldn't this be implemented as a theme?

I based it on what [ohmyzsh did](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/starship), which is a plugin instead of a conventional theme

> but is there a way to unload Starship in a Bash session where Starship was loaded?

Looking at the starship commands, I didn't find a way to turn it off in an active session


maybe I should add this giant warning:

`⚠️ ENABLING THIS PLUGIN WILL UNSET YOUR OSH_THEME VARIABLE`

?

plugins/starship/starship.plugin.sh
```diff
@@ -0,0 +1,10 @@
+#! bash oh-my-bash.module
+
+# Check if starship is installed
+if _omb_util_command_exists starship; then
+    # ignore oh-my-bash theme
+    unset OSH_THEME
```
akinomyoga, line 6:
```suggestion
    unset -v OSH_THEME
```

akinomyoga (approved):

akinomyoga:
Although providing an external theme through a "plugin" instead of a "theme" seems wrong, it should probably better for the consistency with the upstream Oh My Zsh.

PR #635, opened 2024-11-13
Fix powerline multiline last status check

DIG-:
Using the same check syntax used in `powerline.base.sh`, in same function (`__powerline_last_status_prompt`) at line [173](https://github.com/ohmybash/oh-my-bash/blob/5ce9fadcde08c5751c6da008ae3a1d4053516caf/themes/powerline/powerline.base.sh#L173)

### DIFF
--- a/themes/powerline-multiline/powerline-multiline.base.sh
+++ b/themes/powerline-multiline/powerline-multiline.base.sh
@@ -3,7 +3,7 @@
 source "$OSH/themes/powerline/powerline.base.sh"
 
 function __powerline_last_status_prompt {
-  (($1 != 0)) && _omb_util_print "$(set_color $LAST_STATUS_THEME_PROMPT_COLOR -) $1 $_omb_prompt_normal"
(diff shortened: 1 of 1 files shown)


akinomyoga:
`(( ))` is preferred to `[[ ]]` in this context. Can you change `powerline.base.sh` to use `(( ... ))` instead of modifying `powerline-multiline.base.sh`?

DIG-:
The syntax `(( $1 != 0 ))` was producing bellow error
> bash: ((: != 0: syntax error: operand expected (error token is "!= 0")

While debugging I found that I have wrong config.
I put `last_status` in `POWERLINE_LEFT_PROMPT`, but it resolves with error.. There is no need to it, since last_result is always printed.

Meanwhile.. In `powerline.base.sh`, they set the last status as string:
https://github.com/ohmybash/oh-my-bash/blob/5ce9fadcde08c5751c6da008ae3a1d4053516caf/themes/powerline/powerline.base.sh#L178
I suggest a new PR to update there.

PR #638, opened 2024-11-23
Copied duru theme

SuperInfinity:
New Theme called "copied-duru", where segment blocks of  "Duru" are kept the same but the color scheme of theme "sexy" are used. In addition time block is added too


themes/copied-duru/copied-duru.theme.sh
```diff
@@ -0,0 +1,28 @@
...
+source "$OSH/themes/sexy/sexy.theme.sh"
+
+if [[ $COLORTERM = gnome-* && $TERM = xterm ]] && infocmp gnome-256color &>/dev/null; then
+  export TERM=gnome-256color
+elif [[ $TERM != dumb ]] && infocmp xterm-256color &>/dev/null; then
+  export TERM=xterm-256color
+fi
+
+function _omb_theme_PROMPT_COMMAND {
+	local current_time=$(date +"%H:%M:%S")
```
akinomyoga, line 13:
```suggestion
	local current_time='\D{%H:%M:%S}'
```

A prompt sequence `\D{...}` can be used to format the date/time, so the fork cost of `date` can be reduced.

akinomyoga (changes requested):
Thanks. LGTM, with one request.

themes/copied-duru/copied-duru.theme.sh
SuperInfinity: Yup, done that. But do check it once though.

akinomyoga: Yeah, I haven't checked it at my side, so please tell me after checking it.

SuperInfinity: Yeah, I have checked and it works!

akinomyoga: Thanks!

akinomyoga (approved):

Issue #644, opened 2024-12-12
Case-sensitive ssh completion


jornodberg:
I had trouble getting ssh-completion to work with hosts defined in .ssh/config

Until I changed the syntax from 'host myhost' to 'Host myhost'.

Why is it case-sensitive, when ssh-config is not?

akinomyoga:
Because the `ssh` completion shipped with Oh My Bash is not perfect.

https://github.com/ohmybash/oh-my-bash/blob/410b6c943fc9af635ad80cdc3537e9e5c42f99e5/completions/ssh.completion.sh#L48

Actually, there are many places to improve. Probably you want to consider using the completion setting for `ssh` from the [`bash-completion`](https://github.com/scop/bash-completion) project.

akinomyoga:
If you already have `bash-completion` installed in your system (specifically at `/usr/share/bash-completion/bash_completion`), you can simply remove `ssh` from the array `completions` in your `~/.bashrc`. Then, the `ssh` completion from the `bash-completion` project would be usually enabled automatically. If your `bash-completion` is not installed at `/usr/share/bash-completion/bash_completion`, you can manually `source /path/to/bash-completion` in your `~/.bashrc`.

jornodberg:
Thanks.

I've edited the code in ssh.completion.sh to allow different cases for 'host': https://github.com/ohmybash/oh-my-bash/pull/645

Never really contributed code by Git earlier, so please tell if I've done this incorrect. :)
Or also if my code is shait.

akinomyoga:
> Never really contributed code by Git earlier, so please tell if I've done this incorrect. :) Or also if my code is shait.

Don't worry. You are on the right path.

akinomyoga:
Since #645 has already been merged, I think we can simply close this issue.

PR #645, opened 2024-12-12
Update ssh.completion.sh to be case-insensitive

jornodberg:
Also find 'host myhost' and 'HOST myhost' in ssh-config

### DIFF
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -45,7 +45,7 @@ function _omb_completion_ssh {
     fi
   done
   if ((${#config_files[@]} != 0)); then
-    COMPREPLY+=($(compgen -W "$(awk '/^[Hh][Oo][Ss][Tt] /{for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
+    COMPREPLY+=( $(compgen -W "$(awk 'BEGIN {IGNORECASE=1}/^[Hh][Oo][Ss][Tt] / {current_host=""; for (i=2; i<=NF; i++) hosts[current_host=$i]=""}/^[Hh][Oo][Ss][Tt][Nn][Aa][Mm][Ee] / && current_host {hosts[current_host]=$2} END {for (host in hosts) print hosts[host] ? hosts[host] : host}' "${config_files[@]}")" "${options[@]}") )
   fi
 
   local -a known_hosts_files=()
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -48,7 +48,7 @@ function _omb_completion_ssh {
     COMPREPLY+=($(compgen -W "$(awk '
       sub(/^[ \t]*[Hh][Oo][Ss][Tt]([Nn][Aa][Mm][Ee])?[ \t=]+/, "") {
         n = split($0, fields, /[ \t]+/);
-        for (i = 1; i <= n; i++) if (fields[i] && !visited[fields[i]]++) print fields[i];
+        for (i = 1; i <= n; i++) if (fields[i] != "" && !visited[fields[i]]++) print fields[i];
       }' "${config_files[@]}")" "${options[@]}") )
   fi
 
--- a/completions/ssh.completion.sh
+++ b/completions/ssh.completion.sh
@@ -45,7 +45,7 @@ function _omb_completion_ssh {
     fi
   done
   if ((${#config_files[@]} != 0)); then
-    COMPREPLY+=( $(compgen -W "$(awk 'BEGIN {IGNORECASE=1}/^[Hh][Oo][Ss][Tt] / {current_host=""; for (i=2; i<=NF; i++) hosts[current_host=$i]=""}/^[Hh][Oo][Ss][Tt][Nn][Aa][Mm][Ee] / && current_host {hosts[current_host]=$2} END {for (host in hosts) print hosts[host] ? hosts[host] : host}' "${config_files[@]}")" "${options[@]}") )
+    COMPREPLY+=($(compgen -W "$(awk '/^[Hh][Oo][Ss][Tt] / { current=""; for (i=2; i<=NF; i++) { hosts[$i]=1; current=$i; }}/^[ \t]*[Hh][Oo][Ss][Tt][Nn][Aa][Mm][Ee][ \t]+/ && current { if (!($2 in hosts)) hosts[$2]=1;}END { for (host in hosts) print host;}' "${config_files[@]}")" "${options[@]}"))
   fi
 
   local -a known_hosts_files=()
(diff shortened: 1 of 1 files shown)


jornodberg:
This is a commit to solve my own problem https://github.com/ohmybash/oh-my-bash/issues/644

completions/ssh.completion.sh
```diff
@@ -45,7 +45,7 @@ function _omb_completion_ssh {
     fi
   done
   if ((${#config_files[@]} != 0)); then
-    COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
+    COMPREPLY+=($(compgen -W "$(awk '/^[Hh][Oo][Ss][Tt] /{for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
```
akinomyoga, line 48: You may also check [the regular expression used in `bash-completion`](https://github.com/scop/bash-completion/blob/8a8880db78e9b04fb3d66681028af701701f93f2/bash_completion#L2743-L2746).

By the way, the original regular expression `^Host` also matched `HostName`, which is followed by the full names of the hosts (in contrast to `Host` where the user can give a short name). This change would disable showing the full name of the hostnames.

jornodberg: Ah. Didn't take into account that reading 'HostName' would be a good idea.
Have now changed the code to also read HostName in different cases. But without hitting other entries, like HashKnownHosts, HostbasedAcceptedAlgorithms, HostbasedAuthentication or HostKeyAlgorithms like it did before.

akinomyoga: Thanks! Yeah, right. It's even better to check `HostName` explicitly to avoid hitting other options.

completions/ssh.completion.sh
```diff
@@ -45,7 +45,7 @@ function _omb_completion_ssh {
     fi
   done
   if ((${#config_files[@]} != 0)); then
-    COMPREPLY+=($(compgen -W "$(awk '/^[Hh][Oo][Ss][Tt] /{for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
+    COMPREPLY+=( $(compgen -W "$(awk 'BEGIN {IGNORECASE=1}/^[Hh][Oo][Ss][Tt] / {current_host=""; for (i=2; i<=NF; i++) hosts[current_host=$i]=""}/^[Hh][Oo][Ss][Tt][Nn][Aa][Mm][Ee] / && current_host {hosts[current_host]=$2} END {for (host in hosts) print hosts[host] ? hosts[host] : host}' "${config_files[@]}")" "${options[@]}") )
```
akinomyoga, line 48: - `IGNORECASE` is the GNU extension, while Oh My Bash doesn't only target GNU systems. Actually, since you've specified the patterns as `[Hh][Oo]...`, I think you don't need to set that variable.
- As far as I understand the updated version, the name specified by `Host` seems to be disabled if `HostName` is defined. However, suggesting both `Host` and `HostName` is better. If only either one is shown, it should suggest `Host` (but not `HostName`). Could you fix this?

akinomyoga (changes requested):

akinomyoga:
- `Host` can be indented
- Also, what do you think of `=` in [the corresponding code in `bash-completion`](https://github.com/scop/bash-completion/blob/8a8880db78e9b04fb3d66681028af701701f93f2/bash_completion#L2743-L2746)?

jornodberg:
I really didn't know Host could be indented. But I've confirmed that now - so I have adjusted for that.
Never used = myself, so did not take that into account either. But have done that now.

Have tried many different syntaxes in my own ssh-config now, and also confirmed it will suggest IP-adresses from HostName-lines.

completions/ssh.completion.sh
```diff
@@ -45,7 +45,7 @@ function _omb_completion_ssh {
     fi
   done
   if ((${#config_files[@]} != 0)); then
-    COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
+    COMPREPLY+=( $(compgen -W "$(awk '/^[ \t]*[Hh][Oo][Ss][Tt][ \t=]+/ {current=""; n=split($0, fields, /[ \t=]+/); for (i=2; i<=n; i++) if (fields[i] && !(fields[i] in hosts)) hosts[fields[i]]=1;} /^[ \t]*[Hh][Oo][Ss][Tt][Nn][Aa][Mm][Ee][ \t=]+/ {n=split($0, fields, /[ \t=]+/); for (i=2; i<=n; i++) if (fields[i] && !(fields[i] in hosts)) hosts[fields[i]]=1;} /^[ \t]*$/ {current="";} END {for (host in hosts) print host;}' "${config_files[@]}")" "${options[@]}") )
```
akinomyoga, line 48: Hmm, I think `n=split($0, fields, /[ \t=]+/);` for `$0 = "  Host myhost"` (with indentation) would produce an empty element as the first field; i.e. it would produce `fields[1] = ""; fields[2] = "Host"; fields[3] = "myhost"`.

akinomyoga:
```suggestion
    COMPREPLY+=($(compgen -W "$(awk '
      sub(/^[ \t]*[Hh][Oo][Ss][Tt]([Nn][Aa][Mm][Ee])?[ \t=]+/, "") {
        n = split($0, fields, /[ \t=]+/);
        for (i = 1; i <= n; i++) if (fields[i] && !visited[fields[i]]++) print fields[i];
      }' "${config_files[@]}")" "${options[@]}") )
```

akinomyoga (approved):

akinomyoga:
Do you have an explanation on 551f28286bf34b123525552b27804f040907ed3d?

jornodberg:
I added = as a possible separator

akinomyoga:
As far as I tried, `=` can be used to separate the key `Host{,Name}` and the value (i.e., the list of hostnames), but `=` cannot be used to separate hostnames. Rather, `=` can be used as a part of a hostname. Also, `man ssh_config` mention that `=` (up to once) can be used to separate the option name and the value, but there is no mention about `=` as a separator inside the value.

jornodberg:
You are right. This last commit should be abandoned.

akinomyoga (approved):
Thank you

completions/ssh.completion.sh
```diff
@@ -45,7 +45,11 @@ function _omb_completion_ssh {
     fi
   done
   if ((${#config_files[@]} != 0)); then
-    COMPREPLY+=($(compgen -W "$(awk '/^Host/ {for (i=2; i<=NF; i++) print $i}' "${config_files[@]}")" "${options[@]}"))
+    COMPREPLY+=($(compgen -W "$(awk '
+      sub(/^[ \t]*[Hh][Oo][Ss][Tt]([Nn][Aa][Mm][Ee])?[ \t=]+/, "") {
+        n = split($0, fields, /[ \t]+/);
+        for (i = 1; i <= n; i++) if (fields[i] && !visited[fields[i]]++) print fields[i];
```
akinomyoga, line 51: Hmm, if we just use `fields[i]` as a condition, the hostname shortcut `0` would be excluded. We should explicitly check that the field is a non-empty string.

```suggestion
        for (i = 1; i <= n; i++) if (fields[i] != "" && !visited[fields[i]]++) print fields[i];
```

akinomyoga (approved):

akinomyoga:
Thanks!

PR #651, opened 2025-01-19
Add unicode support for brainy prompt

jocyber:
The brainy terminal prompt could not process unicode characters correctly. Unicode characters get processed as a string literal due to the printf function. 

added the following to .bashrc:
`export THEME_PROMPT_CHAR_PS1="\u03bb -> "`

### DIFF
--- a/themes/brainy/brainy.theme.sh
+++ b/themes/brainy/brainy.theme.sh
@@ -172,7 +172,7 @@ function ___brainy_prompt_exitcode {
 function ___brainy_prompt_char {
   local color=$_omb_prompt_bold_white
   local prompt_char="${__BRAINY_PROMPT_CHAR_PS1}"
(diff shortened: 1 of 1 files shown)


akinomyoga:
> added the following to .bashrc:
> `export THEME_PROMPT_CHAR_PS1="\u03bb -> "`

This does not produce a Unicode character. This is an escape sequence of the Unicode character, i.e., an ASCII representation. To assign an Unicode character, one needs to use the string literal of the form `$'...'`:

```bash
THEME_PROMPT_CHAR_PS1=$'\u03bb -> '
```

Also, `export` is unnecessary. `export` is used to mark it as an environment variable that should be inherited by the external programs, but `THEME_PROMPT_CHAR_PS1` is used only in the current Bash session.

Issue #259, opened 2021-11-19
Feature Request: Spack environment name in prompt


prehensilecode:
[Spack](https://github.com/spack/spack) is a software build/management system similar to Homebrew. It's mostly used in HPC environments. 

With Spack, you can define "environments" which behave like Conda envs or pipenvs. The Spack env is given in an environment variable, e.g. for an environment called `foobar`
```
SPACK_ENV=/usr/local/spack/var/spack/environments/foobar
```

I would like to request a new feature to add the Spack env name to the prompt, similar to what is done currently with Conda env or pipenv.

prehensilecode:
This is done in theme.

PR #657, opened 2025-02-12
feat: add spack environment in prompt

thomas-saigre:
[Spack](https://spack.io/) is a packaging manager allowing to create an environment (like Python virtual environment).
I added a function allowing the display of the spack environment name in the prompt.

image

The theme `font` was also updated with this modification (see the screenshot above - I have modified the colors on my local installation, but not on this version !)

Note that if Spack is not installed, this should not cause any error (at least it did not on my local tests)

I found issue #259 about this, but no actual solution was proposed (actually, I think he just did what I have done on his local installation :wink:)

I tried to follow the coding guidelines, if there is something that need to be improved, just tell me :)

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -80,6 +80,7 @@ CHRUBY_THEME_PROMPT_SUFFIX='|'
 # OMB_PROMPT_CONDAENV_USE_BASENAME=true
 # OMB_PROMPT_PYTHON_VERSION_FORMAT=' |%s|'
 # OMB_PROMPT_SHOW_PYTHON_VENV=true
+OMB_PROMPT_SPACK_ENV_FORMAT='[%s] '
 
 # deprecate
 VIRTUALENV_THEME_PROMPT_PREFIX=' |'
@@ -536,6 +537,16 @@ function _omb_prompt_get_python_env {
   _omb_prompt_get_python_version
   python_env=$virtualenv$condaenv$python_version
   [[ $python_env ]]
+}
+
+## @fn _omb_prompt_get_spack_env
+##   @var[out] spack_env
+##   @exit
+function _omb_prompt_get_spack_env {
+  spack_env=
+  [[ ${OMB_PROMPT_SHOW_SPACK_ENV-} == true ]] || return 1
+  [[ ${SPACK_ENV-} ]] || return 1
+  _omb_prompt_format spack_env "$(basename "$SPACK_ENV")" OMB_PROMPT_SPACK_ENV
 }
 
 _omb_util_defun_print _omb_prompt_{print,get}_virtualenv virtualenv
--- a/README.md
+++ b/README.md
@@ -239,6 +239,20 @@ Some themes turn on it by default.  If you would like to turn it off, you may di
 OMB_PROMPT_SHOW_PYTHON_VENV=false
 ```
 
+#### Enable/disable Spack environment information
+
+To enable the Spack environment information in the prompt, please set the
+following shell variable in `~/.bashrc`:
+
+```bash
+OMB_PROMPT_SHOW_SPACK_ENV=true
+```
+
+If the theme supports it, the information of the currently active Spack
+environment will be shown.  If the theme you use does not support the Spack
+environment information, a pull request to add it is welcome.  See the `font`
+theme as an example implementation of including the Spack environment.
+
 #### Disable internal uses of `sudo`
 
 Some plugins of oh-my-bash internally use `sudo` when it is necessary.  However, this might clutter with the `sudo` log.  To disable the use of `sudo` by oh-my-bash, `OMB_USE_SUDO` can be set to `false` in `~/.bashrc`.
--- a/templates/bashrc.osh-template
+++ b/templates/bashrc.osh-template
@@ -79,6 +79,10 @@ OMB_USE_SUDO=true
 # OMB_PROMPT_SHOW_PYTHON_VENV=true  # enable
 # OMB_PROMPT_SHOW_PYTHON_VENV=false # disable
 
+# To enable/disable Spack environment information
+# OMB_PROMPT_SHOW_SPACK_ENV=true  # enable
+# OMB_PROMPT_SHOW_SPACK_ENV=false # disable
+
 # Which completions would you like to load? (completions can be found in ~/.oh-my-bash/completions/*)
 # Custom completions may be added to ~/.oh-my-bash/custom/completions/
 # Example format: completions=(ssh git bundler gem pip pip3)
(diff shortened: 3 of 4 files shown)


lib/omb-prompt-base.sh
```diff
@@ -538,6 +538,15 @@ function _omb_prompt_get_python_env {
   [[ $python_env ]]
 }
 
+## @fn _omb_prompt_get_spack_env
+##   @var[out] spack_env
+##   @exit
+function _omb_prompt_get_spack_env {
+  spack_env=
+  [[ ${SPACK_ENV-} ]] || return 1
+  _omb_prompt_format spack_env "[$(basename "$SPACK_ENV")] " OMB_PROMPT_SPACK_ENV:SPACK_ENV_THEME_PROMPT
```
akinomyoga, line 547:
```suggestion
  _omb_prompt_format spack_env "$(basename "$SPACK_ENV")" OMB_PROMPT_SPACK_ENV
```

- Any formatting such as enclosing the value within "[...]" should be performed through the corresponding formatting variable, `OMB_PROMPT_SPACE_ENV_FORMAT='[%s] '`. Please add this default value around line 83.
- The name after the colon in the last argument of `_omb_prompt_format` is for the deprecated variable names. In this case, `:SPACK_ENV_THEME_PROMPT` would imply that it also checks the *deprecated* variables `SPACK_ENV_THEME_PROMPT_PREFIX` and `SPACK_ENV_THEME_PROMPT_SUFFIX`. However, we shouldn't need to support these variables since we have never provided a feature to use `SPACK_ENV_THEME_PROMPT_PREFIX` and `SPACK_ENV_THEME_PROMPT_SUFFIX`. We can simply drop it.

akinomyoga (commented):
Thank you for your contribution! I'd like to include the changes in `lib/omb-prompt-base.sh`.

> The theme font was also updated with this modification (see the screenshot above - I have modified the colors on my local installation, but not on this version !)
>
> Note that if Spack is not installed, this should not cause any error (at least it did not on my local tests)

I'm not sure if I should accept the change in the `font` theme. When modifying an existing theme, I need to be careful about the impact on the existing users. When Spack is not installed, it doesn't affect the behavior as you correctly clarified. However, when Spack is installed in the system, this change would affect the prompt. Existing users may not want to show the Spack environment even if it is activated.

As described in [README](https://github.com/ohmybash/oh-my-bash?tab=readme-ov-file#customization-of--plugins-and-themes), when a user wants to modify the items and the layout of the prompt, the user is expected to make a modified version in `.oh-my-bash/custom/themes`.

akinomyoga:
> However, when Spack is installed in the system, this change would affect the prompt. Existing users may not want to show the Spack environment even if it is activated.

One solution used in the codebase is to prepare another variable, `OMB_PROMPT_SHOW_SPACK_VENV`, which can be used to opt in the Spack environment information in the prompt. See the following line for the example of Python environment.

https://github.com/ohmybash/oh-my-bash/blob/2651f35600fbf3452b1d1652ccfb709b1a1692d7/lib/omb-prompt-base.sh#L521

This way, when a user wants to enable the Spack information, the user can specify the following in their `~/.bashrc`:

```bash
OMB_PROMPT_SHOW_SPACK_VENV=yes
```

thomas-saigre:
For the record, here are the new feature that has been added regarding the display of Spack environment in prompt:

If a spack environment is activated, and if the variable `OMB_PROMPT_SHOW_SPACK_VENV` is set to `true` (in `.bashrc`), the name of the environment can be loaded and included in prompt by setting these lines in the theme file:

```sh
local spack_env; _omb_prompt_get_spack_env
spack_env=$_omb_prompt_red$spack_env
```

The name is then displayed by adding `$spack_env` in the prompt:

```sh
PS1="$(clock_prompt)$spack_env$python_venv${hostname} ${_omb_prompt_bold_teal}\w $(scm_prompt_char_info)${ret_status}→ ${_omb_prompt_normal}"
```

Note that the format can be personalized by setting the variable `OMB_PROMPT_SPACK_ENV_FORMAT` :
```sh
OMB_PROMPT_SPACK_ENV_FORMAT='[%s] '
```

lib/omb-prompt-base.sh
```diff
@@ -80,6 +80,7 @@ CHRUBY_THEME_PROMPT_SUFFIX='|'
 # OMB_PROMPT_CONDAENV_USE_BASENAME=true
 # OMB_PROMPT_PYTHON_VERSION_FORMAT=' |%s|'
 # OMB_PROMPT_SHOW_PYTHON_VENV=true
+# OMB_PROMPT_SPACK_ENV_FORMAT='[%s] '
```
akinomyoga, line 83:
```suggestion
OMB_PROMPT_SPACK_ENV_FORMAT='[%s] '
```

lib/omb-prompt-base.sh
```diff
@@ -538,6 +539,16 @@ function _omb_prompt_get_python_env {
   [[ $python_env ]]
 }
 
+## @fn _omb_prompt_get_spack_env
+##   @var[out] spack_env
+##   @exit
+function _omb_prompt_get_spack_env {
+  spack_env=
+  [[ ${SPACK_ENV-} ]] || return 1
+  [[ ${OMB_PROMPT_SHOW_PYTHON_VENV-} == true ]] || return 1
```
akinomyoga, line 548:
```suggestion
  [[ ${OMB_PROMPT_SHOW_SPACK_ENV-} == true ]] || return 1
```

akinomyoga: Also, I think we should test this before testing `${SPACE_ENV-}`

akinomyoga (changes requested):
I'm not sure if you understood my previous comment, but the behavior of the existing users should be preserved. There are two options.

- We do not modify the `font` theme within this PR.
- We modify the `font` theme, but the new Spack environment information is only shown when the user explicitly asks to show it by setting`OMB_PROMPT_SHOW_SPACK_ENV`. In this way, the behavior would be kept for the existing users..

In the current version, the `font` theme is not modified, but`OMB_PROMPT_SHOW_SPACE_ENV` is supported.

lib/omb-prompt-base.sh
thomas-saigre: What option would be the best for oh-my-bash ?
Intuitively, I'd say that not changing the theme would be better (avoiding changing all themes), but the users will have to activate by hand

akinomyoga: Now you've added a mechanism to opt-in `OMB_PROMPT_SHOW_SPACK_ENV`, so we can include the spack environment information in the prompt of the `font` theme without affecting the existing users. I re-added the changes to the `font` theme in commit 0559dc7.

thomas-saigre: ok great! Thanks :)

akinomyoga:
I added a fix for the variable name in commit da54f81. I also added a description in `README` and the default `.bashrc`.

akinomyoga:
I've merged it. Thanks for your contribution!

PR #659, opened 2025-02-19
feat(git): add autocompletion for git aliases in git.plugin.sh

codernirdesh:
This commit improves the Git plugin in Oh My Bash by adding autocompletion support for various Git aliases. The enhancements include:
- Adding `__git_complete` for multiple Git aliases to improve shell autocompletion.
- Ensuring better developer experience by providing context-aware suggestions for Git commands.
- Improving efficiency when using Git shortcuts by enabling autocompletion where applicable.

These changes enhance usability, making it easier for users to navigate and execute Git commands quickly.

Fixes: #658, #633

### DIFF
--- a/plugins/git/git.plugin.sh
+++ b/plugins/git/git.plugin.sh
@@ -2,6 +2,7 @@
 #
 # Functions
 #
+_omb_module_require completion:git
 
 # The name of the current branch
 # Back-compatibility wrapper for when this function was defined here in
--- a/plugins/git/git.plugin.sh
+++ b/plugins/git/git.plugin.sh
@@ -57,60 +57,100 @@ function git_main_branch() {
 # https://github.com/ohmyzsh/ohmyzsh/blob/f36c6db0eac17b022eee87411e6996a5f5fc8457/plugins/git/git.plugin.zsh
 #
 
+# Git aliases
 alias g='command git'
+__git_complete g _git
 
 alias ga='command git add'
+__git_complete ga _git_add
 alias gaa='command git add --all'
+__git_complete gaa _git_add
 alias gapa='command git add --patch'
+__git_complete gapa _git_add
 alias gau='command git add --update'
+__git_complete gau _git_add
 alias gav='command git add --verbose'
+__git_complete gav _git_add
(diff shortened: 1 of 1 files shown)


akinomyoga (changes requested):
In general, the user may not have loaded any git completion settings; `__git_complete` may not be defined. To make sure that the git completion setting is loaded, could you put the following line at the beginning of the file?

```bash
_omb_module_require completion:git
```

codernirdesh:
Thanks for the suggestion!

I've added `_omb_module_require completion:git` at the beginning of `git.plugin.sh` to ensure Git completion settings are loaded before using `__git_complete`.  

Let me know if any further improvements are needed!

plugins/git/git.plugin.sh
```diff
@@ -2,6 +2,7 @@
 #
 # Functions
 #
+_omb_module_require completion:git
```
akinomyoga, line 5:
```suggestion

_omb_module_require completion:git

#
# Functions
#
```

We should probably put the dependency before the section "Functions". These comments are the section title for the subsequent functions, `current_branch`, `work_in_progress`, etc.

codernirdesh: I've made changes regarding this.

akinomyoga (commented):
Thanks! I'll later check it in my local environment.

plugins/git/git.plugin.sh
```diff
@@ -57,60 +58,100 @@ function git_main_branch() {
 # https://github.com/ohmyzsh/ohmyzsh/blob/f36c6db0eac17b022eee87411e6996a5f5fc8457/plugins/git/git.plugin.zsh
 #
 
+# Git aliases
```
akinomyoga, line 61: This seems like a duplicate of `# Aliases` 6 lines above. May I ask about what is the intent of adding this new line?

codernirdesh: I didn't noticed there was `#Aliases` before. Thanks for pointing it out. I've removed the duplicate.

akinomyoga:
I now checked it in my local environment. I added a blank line in commit 7b9d3d5.

akinomyoga (approved):

akinomyoga:
I merged it, thanks!

Hmm, did you use AI to generate commit messages? I tried to format the commit message, but I realized that these contain superficial statements essentially stating only that "it becomes efficient and useful" using several sentences. I significantly shortened the message.

PR #669, opened 2025-03-27
new simple cargo plugin

sjplanet:


plugins/cargo/cargo.plugin.sh
```diff
@@ -0,0 +1,9 @@
+#! bash oh-my-bash.module
+# @sjplanet 2025
+
+if [ -d ~/.cargo ]; then
```
akinomyoga, line 4:
```suggestion
if [[ ! -d ${CARGO_HOME-} && -d ~/.cargo ]]; then
```

sjplanet: I just want to extend PATH if .cargo is present, but you're right, it needs a few more prerequisite queries.

plugins/cargo/cargo.plugin.sh
```diff
@@ -1,9 +1,14 @@
...
 
-fi
+# Check if the directory exists
+if [ -d "$CARGO_HOME" ]; then
+    # Add $CARGO_HOME/bin to PATH if it's not already included
+    case ":$PATH:" in
+        *":$CARGO_HOME/bin:"*) ;;  # already in PATH, do nothing
+        *) export PATH="$CARGO_HOME/bin${PATH:+:$PATH}" ;;
+    esac
+fi
```
akinomyoga, line 14: Thanks for updating. I think this part is supposed to be provided by `$CARGO_HOME/env`.

```suggestion
# Check if the directory exists
if [[ -f $CARGO_HOME/env && -r $CARGO_HOME/env ]]; then
  source "$CARGO_HOME/env"
fi
```

Or is there a case where `$CARGO_HOME/env` doesn't exist?

sjplanet: I'm not a rust programmer, so on my system there isn't $CARGO_HOME/env . Also in the man page of cargo is nothing about $CARGO_HOME/env. I'm using cargo to build and install rust programs and cargo installs the rust programs into $CARGO_HOME/bin

akinomyoga: Hmm, OK. I'm similar to you, meaning I also just build Rust programs using `cargo`. Maybe it depends on how you got `cargo`. The recommended way is to use `rustup`, but if `cargo` is installed through a different way, `~/.cargo/env` may not be there. How did you get `cargo` in your system?

sjplanet: openSuSE Leap 15.6: zypper install cargo

akinomyoga:
```suggestion
# Check if the directory exists
if [[ -d "$CARGO_HOME" ]]; then
  if [[ -f $CARGO_HOME/env && -r $CARGO_HOME/env ]]; then
    source "$CARGO_HOME/env"
  else
    # Add $CARGO_HOME/bin to PATH if it's not already included
    case ":$PATH:" in
    *":$CARGO_HOME/bin:"*) ;;  # already in PATH, do nothing
    *) export PATH="$CARGO_HOME/bin${PATH:+:$PATH}" ;;
    esac
  fi
fi
```

plugins/cargo/cargo.plugin.sh
```diff
@@ -0,0 +1,19 @@
+#! bash oh-my-bash.module
+# @sjplanet 2025
+
+# Set CARGO_HOME to default if not already set
+: "${CARGO_HOME:=$HOME/.cargo}"
+
+# Check if the directory exists
+if [[ -d "$CARGO_HOME" ]]; then
+  if [[ -f "$CARGO_HOME/env" && -r "$CARGO_HOME/env" ]]; then
```
akinomyoga, line 9: `[[ ... ]]` is not a normal command, but a special syntax of Bash that was introduced to solve the quoting issues with `[ ... ]`. One of the issues is the word splitting and the pathname expansions that are applied to the arguments of the normal commands such as `[`. To prevent the word splitting and the pathname expansions, users needed to quote the arguments as `"$var"`. `[[ ... ]]` was introduced to solve such an issue so that the users do not need to quote the expansions as `"$var"`. Thus, quoting parameter expansions in `[[ ... ]]` does not make sense.

plugins/cargo/cargo.plugin.sh
```diff
@@ -0,0 +1,19 @@
...
+    # Note: Rustup sets PATH in $CARGO/env
+    source "$CARGO_HOME/env"
+  else
+    # Add $CARGO_HOME/bin to PATH if it's not already included
+    case :$PATH: in
+    *":$CARGO_HOME/bin:"*) ;;  # already in PATH, do nothing
+    *) export PATH="$CARGO_HOME/bin${PATH:+:$PATH}" ;;
```
akinomyoga, line 16: The right-hand side of the variable assignment is not subject to word splitting and pathname expansions either.

plugins/cargo/cargo.plugin.sh
akinomyoga: In fact, [the page you provided](https://guide.bash.academy/expansions/?=Command_Substitution#a1.3.0_1) includes the following example:

```bash
[[ $- = *i* ]]
```

while there are no examples of quoting the parameter expansion of `[[ ... ]]`.

plugins/cargo/cargo.plugin.sh
akinomyoga: Also for this case, the page you provided includes the following example:

```bash
first=${1%% *} last=${1##* } middle=${1#$first} middle=${middle%$last}
```

plugins/cargo/cargo.plugin.sh
akinomyoga: I'll drop those commits since there are no replies from the OP, I'm pretty sure these are not problems, and these are inconsistent with the existing styles of the codebase.

akinomyoga (approved):

PR #673, opened 2025-04-10
Improved python tool support

marco-gallegos:
just a small update :)

### DIFF
--- a/themes/developer/developer.theme.sh
+++ b/themes/developer/developer.theme.sh
@@ -54,12 +54,24 @@ function _omb_theme_developer__read_ruby_version {
 ## @var[out] REPLY
 function _omb_theme_developer__readPyVersion {
   local val_py=$(python --version 2>/dev/null | cut -d ' ' -f 2)
+  local val_uv=$(uv --version 2>/dev/null | cut -d ' ' -f 2)
+  local val_venv=""
+
+  ## according some docs this env var just exist when a venv/uv venv is active
+  if [[ -n "$VIRTUAL_ENV" ]]; then 
+    val_venv=" ($(echo "$VIRTUAL_ENV" | awk -F/ '{print $(NF-1)"/"$NF}'))"
+  fi
+
   if _omb_util_command_exists conda; then
     local condav=$(conda env list | awk '$2 == "*" {print $1}')
     val_py="conda<$condav> $val_py"
   else
-    # Si nvm no está instalado, utilizar "njs"
-    val_py="py $val_py"
+    if _omb_util_command_exists uv; then
+      val_py="uv($val_uv) $val_py$val_venv"
+    else
+      # if conda and uv is not installed use "py"
+      val_py="py $val_py"
+    fi
   fi
   REPLY=$val_py
 }


themes/developer/developer.theme.sh
```diff
@@ -54,12 +54,24 @@ function _omb_theme_developer__read_ruby_version {
 ## @var[out] REPLY
 function _omb_theme_developer__readPyVersion {
   local val_py=$(python --version 2>/dev/null | cut -d ' ' -f 2)
+  local val_uv=$(uv --version 2>/dev/null | cut -d ' ' -f 2)
+  local val_venv=""
+
+  ## according some docs this env var just exist when a venv/uv venv is active
+  if [[ -n "$VIRTUAL_ENV" ]]; then
+    val_venv=" ($(echo "$VIRTUAL_ENV" | awk -F/ '{print $(NF-1)"/"$NF}'))"
```
akinomyoga, line 62: I'm wondering what happens when `VIRTUAL_ENV` contains a newline in the path. The above would output multiple `aaa/bbb` pairs. Also, what happens when `VIRTUAL_ENV` doesn't contain any `/`? It would output `/` + `$VIRTUAL_ENV`, but I'm not sure if it's the intended behavior. I'd suggest simply doing

```suggestion
  if [[ $VIRTUAL_ENV =~ [^/]+/[^/]+$ ]]; then
    val_venv=" ($BASH_REMATCH)"
```

themes/developer/developer.theme.sh
```diff
@@ -54,12 +54,24 @@ function _omb_theme_developer__read_ruby_version {
...
+  fi
+
   if _omb_util_command_exists conda; then
     local condav=$(conda env list | awk '$2 == "*" {print $1}')
     val_py="conda<$condav> $val_py"
   else
-    # Si nvm no está instalado, utilizar "njs"
-    val_py="py $val_py"
+    if _omb_util_command_exists uv; then
+      val_py="uv($val_uv) $val_py$val_venv"
```
akinomyoga, line 70: The values of `val_uv` and `val_venv` are only used inside this if-branch. The values of these variables should be calculated inside this branch.

themes/developer/developer.theme.sh
```diff
@@ -54,12 +54,24 @@ function _omb_theme_developer__read_ruby_version {
...
+    val_venv=" ($(echo "$VIRTUAL_ENV" | awk -F/ '{print $(NF-1)"/"$NF}'))"
+  fi
+
   if _omb_util_command_exists conda; then
     local condav=$(conda env list | awk '$2 == "*" {print $1}')
     val_py="conda<$condav> $val_py"
   else
-    # Si nvm no está instalado, utilizar "njs"
-    val_py="py $val_py"
+    if _omb_util_command_exists uv; then
```
akinomyoga, line 69: This `else` and `if` can be combined to `elif`.

marco-gallegos:
Thank you about your review @akinomyoga  I just pushed these changes to my branch

themes/developer/developer.theme.sh
akinomyoga: > Also, what happens when `VIRTUAL_ENV` doesn't contain any `/`?

This hasn't been answered. Or is it ensured that it `VIRTUAL_ENV` contains `/`?

PR #675, opened 2025-04-17
Ensure rmdir doen't fail / return non zero

senderic:
This is meant to fix this error:

```sh
         __                          __               __
  ____  / /_     ____ ___  __  __   / /_  ____ ______/ /_
 / __ \/ __ \   / __ `__ \/ / / /  / __ \/ __ `/ ___/ __ \
/ /_/ / / / /  / / / / / / /_/ /  / /_/ / /_/ (__  ) / / /
\____/_/ /_/  /_/ /_/ /_/\__, /  /_.___/\__,_/____/_/ /_/
                        /____/
Hooray! Oh My Bash has been updated and/or is at the current version.
To keep up on the latest news and updates, follow us on GitHub: https://github.com/ohmybash/oh-my-bash
rmdir: failed to remove '/home/esender/.oh-my-bash/log/update.lock': No such file or directory
```

That `rmdir` failure can be avoided with `command rmdir "$OSH"/log/update.lock || true`

```c
01:52:00 esender@machine .oh-my-bash ±|rmdir-no-fail|→ command rmdir xxx
rmdir: failed to remove 'xxx': No such file or directory
01:52:02 esender@machine .oh-my-bash ±|rmdir-no-fail|→ echo $?
1 # <--- Bad!
```
```c
01:52:04 esender@machine .oh-my-bash ±|rmdir-no-fail|→ command rmdir xxx || :
rmdir: failed to remove 'xxx': No such file or directory
01:52:14 esender@machine .oh-my-bash ±|rmdir-no-fail|→ echo $?
0 # <--- Not Bad!
```

### DIFF
--- a/tools/check_for_upgrade.sh
+++ b/tools/check_for_upgrade.sh
@@ -57,7 +57,7 @@ type -P git &>/dev/null || return 0
 
 if command mkdir -p "$OSH/log/update.lock" 2>/dev/null; then
   _omb_upgrade_check
-  command rmdir "$OSH"/log/update.lock
+  command rmdir "$OSH"/log/update.lock || :
 else
   printf '%s\n' \
     'oh-my-bash/check_for_upgrade: Failed to get a lock.  Please make sure that no' \


tools/check_for_upgrade.sh
```diff
@@ -57,7 +57,7 @@ type -P git &>/dev/null || return 0
 
 if command mkdir -p "$OSH/log/update.lock" 2>/dev/null; then
   _omb_upgrade_check
-  command rmdir "$OSH"/log/update.lock
+  command rmdir "$OSH"/log/update.lock || :
```
senderic, line 60: FYI - `:` is a posix null command that will ensure a return value of `0` no matter what.

senderic:
> When there is an error, just ignoring the error cannot be a proper solution in general.
> 


Agree in general, but the context of this particular error was to remove a lock. If the lock didn't exist in the first place, why fully fail? Let's log an error and move forward.

akinomyoga:
> the context of this particular error was to remove a lock.

No, unfortunately. As I've already written above, this was a regression introduced in bdd222055e7adab7a8b8185848d9706f4c63a6cd, and I already fixed it in b71d5d3f0612ab162d065b6139546471900bebca in the right way.

PR #676, opened 2025-04-27
feat: add fzf plugin

matlads:


plugins/fzf/fzf.plugin.sh
```diff
@@ -0,0 +1,8 @@
+#! bash oh-my-bash.module
+
+# Check if fzf is installed
+if _omb_util_command_exists fzf; then
+    [ -f ~/.fzf.bash ] && source ~/.fzf.bash
+else
+    echo '[oh-my-bash] fzf not found, please install it from https://github.com/junegunn/fzf'
```
akinomyoga, line 7:
```suggestion
    _omb_util_print '[oh-my-bash] fzf not found, please install it from https://github.com/junegunn/fzf' >&2
```

plugins/fzf/fzf.plugin.sh
```diff
@@ -0,0 +1,8 @@
+#! bash oh-my-bash.module
+
+# Check if fzf is installed
+if _omb_util_command_exists fzf; then
+    [ -f ~/.fzf.bash ] && source ~/.fzf.bash
```
akinomyoga, line 5:
```suggestion
    [[ -f ~/.fzf.bash ]] && source ~/.fzf.bash
```

akinomyoga (commented):
Thanks. Could you check the following suggestions?

plugins/fzf/fzf.plugin.sh
akinomyoga: In addition to the above suggestion, in the current setup of this PR, the error message is printed when the `fzf` command is not found, but it is not printed when `~/.fzf.bash` does not exist while the `fzf` command exists. In both cases, the feature provided by this `fzf.plugin.sh` is not available, so I think it is more consistent to print the error message also when `~/.fzf.bash` does not exist.

matlads: I have done some further research, and the fzf github repo recommends this way
```
# Set up fzf key bindings and fuzzy completion
eval "$(fzf --bash)"
```
of loading fzf into bash

akinomyoga: OK. That makes sense.

One concern might be that `fzf --bash` was introduced in [0.48.0](https://github.com/junegunn/fzf/releases/tag/0.48.0) 2024-03-14, so users in some older systems might have a problem with it. Nevertheless, such users can manually set it up, so it shouldn't be a big problem.

Thank you for updating!

plugins/fzf/README.md
```diff
@@ -0,0 +1,15 @@
+## fzf plugin
+
+The fzf plugin enables fzf on bash
```
akinomyoga, line 3:
```suggestion
The fzf plugin enables fzf keybindings and completions on bash.
```

plugins/fzf/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+To use it, install
+[fzf](https://github.com/junegunn/fzf?tab=readme-ov-file#installation), add fzf
```
akinomyoga, line 6:
```suggestion
[fzf](https://github.com/junegunn/fzf?tab=readme-ov-file#installation) and add fzf
```

plugins/fzf/README.md
```diff
@@ -0,0 +1,15 @@
...
+to the plugins array of your bashrc file:
+
+```bash
+plugins=(... fzf)
+```
+
+## Aliases
+
+None
```
akinomyoga, line 15: I think we may simply drop the Aliases section.
```suggestion
```

akinomyoga (commented):
These are just cosmetic fixes.

akinomyoga (approved):

akinomyoga (changes requested):
Ah, I completely forgot about `plugins/README.md`. Could you add a row of `fzf` in the table of `plugins/README.md`?

matlads:
> Ah, I completely forgot about `plugins/README.md`. Could you add a row of `fzf` in the table of `plugins/README.md`?
> 
> This request is also described in [`CONTRIBUTING.md`](https://github.com/ohmybash/oh-my-bash/blob/master/CONTRIBUTING.md#new-plugin).

Yeah, I had totally missed that as well. Thanks for catching it.

I have added a line for fzf in `plugins/README.md`

akinomyoga (approved):

plugins/fzf/README.md
```diff
@@ -0,0 +1,11 @@
+## fzf plugin
```
akinomyoga, line 1:
```suggestion
# fzf plugin
```

akinomyoga (approved):

akinomyoga (approved):

akinomyoga:
Thanks! I merged it.

PR #683, opened 2025-05-12
feat(lib/omb-prompt-base): Add support for distrobox

anthqiu:
[Distrobox](https://distrobox.it/) is "a container wrapping layer that allows the user to install containerised versions of Linux that are different to the host while providing tight integration with the host allowing the use of binaries designed for one distribution to run on another [(Arch wiki)](https://wiki.archlinux.org/title/Distrobox)".

This PR implements the basic function to get distrobox container name, and modified the default `font` theme to add distrobox info field.

Screenshot_20250512_160909

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -81,6 +81,7 @@ CHRUBY_THEME_PROMPT_SUFFIX='|'
 # OMB_PROMPT_PYTHON_VERSION_FORMAT=' |%s|'
 # OMB_PROMPT_SHOW_PYTHON_VENV=true
 OMB_PROMPT_SPACK_ENV_FORMAT='[%s] '
+OMB_PROMPT_DISTROBOX_FORMAT='[📦%s] '
 
 # deprecate
 VIRTUALENV_THEME_PROMPT_PREFIX=' |'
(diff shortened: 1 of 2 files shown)


akinomyoga:
I'm not familiar with this field, so I don't have a way to tell whether the suggested change does the right thing. My naive question is whether the environment variable `CONTAINER_ID` is really specific to the Distrobox container. A quick search seems to suggest that `CONTAINER_ID` is not the variable provided by Distrobox but a more general variable associated with Docker. Wouldn't this show the prompt information also for a Docker container that is not managed by Distrobox?

akinomyoga:
[ChatGPT-CONTAINER_ID.pdf](https://github.com/user-attachments/files/20176051/ChatGPT-CONTAINER_ID.pdf)

akinomyoga:
Another thing is whether the Distrobox information in a prompt has a significant demand by a large number of users. For example, Starship (which provides a wide range of options to show various information) doesn't seem to offer a module for Distrobox. Instead, Starship seems to include module [`container`](https://github.com/starship/starship/blob/87f9a2367ba2c7ee81409a969cfa222b9177f217/src/modules/container.rs). This module seems to inspect `/run/*` and `/proc/*` to identify the container information without relying on environment variables. However, I'm not sure if what Starship's `container` tries to achieve is consistent with what this PR tries to achieve.

akinomyoga (changes requested):
If we decide to include the suggested Distrobox information, small adjustments are needed.

- Could you check the use of `OMB_PROMPT_SHOW_SPACK_ENV` and prepare a similar variable for the Distrobox information? We don't want to change the default behavior of the font theme as long as the old default doesn't cause problems. In the present case, we haven't shown the Distrobox information by default, and the existing users of the `font` theme might not want to show the Distrobox information in the prompt. If the `font` theme starts to show it by default and there is not way to turn off, it could be a trouble for existing users. To avoid such a conflict, we want to introduce a variable to enable/disable the prompt information.

izaac:
This would be really helpful. The same it is already for pyenvs with `OMB_PROMPT_SHOW_PYTHON_VENV`

PR #684, opened 2025-05-12
themes/font: add condaenv format

anthqiu:
This PR matches condaenv format with the existing virtualenv format that's already in `font` theme.


akinomyoga (approved):
I am usually very careful about the behavioral/style changes, which largely depend on the users' preferences. There is no need to change the default behavior because each user can configure it as they like. I usually reject such a request, but I think I can accept this.

akinomyoga:
Thank you! I merged this one.

PR #685, opened 2025-05-21
add new theme lucky

meizikeai:
Hi, everyone, 
I like oh-my-bash, I created a theme lucky by myself, this is a pull request, hope it can be merged into the main branch, thank you~


meizikeai:
> @meizikeai Thank you for your contribution. Could you check the instructions in `CONTRIBUTING.md`? In particular, please carefully check the [New Theme](https://github.com/ohmybash/oh-my-bash/blob/master/CONTRIBUTING.md#new-theme) section.

Yes, I was careless and made changes to themes/THEMES.md. 
The files involved in the whole merge request are themes/lucky/lucky-dark.png, themes/lucky/lucky.theme.sh and themes/THEMES.md. 
Maybe it's correct, thanks for your time.

themes/lucky/lucky.theme.sh
```diff
@@ -0,0 +1,59 @@
+#! bash oh-my-bash.module
+
+_omb_deprecate_const 20000 red "$_omb_prompt_brown" "${_omb_deprecate_msg_please_use/'%s'/_omb_prompt_brown}"
+_omb_deprecate_const 20000 blue "$_omb_prompt_navy" "${_omb_deprecate_msg_please_use/'%s'/_omb_prompt_navy}"
+_omb_deprecate_const 20000 green "$_omb_prompt_green" "${_omb_deprecate_msg_please_use/'%s'/_omb_prompt_green}"
+_omb_deprecate_const 20000 purple "$_omb_prompt_purple" "${_omb_deprecate_msg_please_use/'%s'/_omb_prompt_purple}"
```
akinomyoga, line 6: Could you explain these lines? These define variables `red`, `blue`, `green`, and `purple`, but none of them seem to be used in the theme. In the first place, these variables are prepared by `lib/omb-deprecated.sh`, which is always loaded. So these variables already exist.

themes/lucky/lucky.theme.sh
```diff
@@ -0,0 +1,59 @@
...
+L_WHITE=$_omb_prompt_white
+
+L_BOLD_RAD=$_omb_prompt_bold_brown
+L_BOLD_YELLOW=$_omb_prompt_bold_olive
+
+L_PREFIX=$SCM_THEME_PROMPT_PREFIX
+L_SUFFIX=$SCM_THEME_PROMPT_SUFFIX
+
+# reset git_prompt_minimal_info
+function git_prompt_minimal_info {
```
akinomyoga, line 27: Could you also check [Section "*Naming convention of functions and global variables*" in `CONTRIBUTING.md`](https://github.com/ohmybash/oh-my-bash/blob/master/CONTRIBUTING.md#naming-convention-of-functions-and-global-variables)? There are existing old variable and function names in old themes, but we should follow the naming convention introduced in new codes. Please change the variable names of `L_*` . Or you can directly use the variables `_omb_prompt_*`. Or you can declare these variables inside the functions with the `local` keyword.

Please also change the function name `git_prompt_minimal_info`. Not only does it not follow the naming convention, `git_prompt_minimal_info` is already defined in `lib/omb-prompt-base.sh`. The existing functions should not be overwritten, or it would affect the other themes (when a theme is switched later) and other modules. You are supposed to make a new function with a distinct name.

Note: `SCM_*` is defined and used by old codes, so one cannot change the names of the variables `SCM_*`. We want to migrate these old variable names all at once someday, but we just don't have time.

themes/lucky/lucky.theme.sh
```diff
@@ -0,0 +1,59 @@
...
+  local ref
+  local status
+  local git_status_flags=('--porcelain')
+  SCM_STATE=${L_GREEN}${SCM_THEME_PROMPT_CLEAN}
```
akinomyoga, line 31:
```suggestion
  local SCM_STATE=${L_GREEN}${SCM_THEME_PROMPT_CLEAN}
```

Unless you want to expose this variable to the user or the caller functions so that they can directly reference the variable content, please localize it. This variable is used only in this function. The original function defined in `lib/omb-prompt-base.sh` intentionally exposed the variables to the caller, but it doesn't apply to the present case. You might think unlocalized variables don't affect the main shell because this function is only used inside a subshell, but in the Oh-My-Bash codebase, functions should be designed so that they don't leave unnecessary side effects.

themes/lucky/lucky.theme.sh
```diff
@@ -0,0 +1,59 @@
...
+    status=$(_omb_prompt_git status ${git_status_flags} 2>/dev/null | tail -n1)
+
+    if [[ -n ${status} ]]; then
+      SCM_DIRTY=1
+      SCM_STATE=${L_RED}${SCM_THEME_PROMPT_DIRTY}
+    fi
+
+    # Output the git prompt
+    echo -e " ${L_WHITE}on ${L_BLUE}git:${L_PURPLE}${SCM_BRANCH}${SCM_STATE}"
+  fi
```
akinomyoga, line 50: So as far as I understand it correctly, the difference from the original `git_prompt_minimal_info` in `lib/omb-prompt-base.sh` is these color codes and the prefix "` on `"?

I have a naive question. Couldn't that be achieved just by modifying the global variables `SCM_THEME_PROMPT_DIRTY`, `SCM_THEME_PROMPT_PREFIX`, and `SCM_THEME_PROMPT_SUFFIX`? If so, it should be implemented by the global variables instead of defining a new function.

akinomyoga (changes requested):

meizikeai:
@akinomyoga Thank you for pointing out the mistakes and giving suggestions. This is indeed where I haven't done enough. I have modified it according to the agreement in CONTRIBUTING.md and hope it meets the requirements. May good luck always be with you.

themes/lucky/lucky.theme.sh
akinomyoga: I checked the `lucky` theme on my local host. We can just call `scm_prompt_info` with custom `SCM_THEME_*` variables to achieve the same behavior (unless you want to allow users to change `SCM_THEME_PROMPT_{CLEAN,DIRTY,BRANCH_PREFIX}` after loading the theme). I pushed the changes to the PR branch  (i.e. `themes-lucky`) in commit 2c68bf61bba61bbbe623a44a86a389684fc9e09b. @meizikeai It's supposed to keep the result, but could you pull the branch and check if the result is what you expect?

I see some of your changes in `_omb_theme_lucky_git_prompt_minimal_info` are valuable. I applied them to `git_prompt_minimal_info` in `lib/omb-prompt-base.sh` in commit fa7e58ceab55196f6adb7826ac6b41896392df33.

akinomyoga: Thanks!

akinomyoga (approved):

akinomyoga:
Thanks for your contribution! I've merged it!

PR #694, opened 2025-07-10
Supports more network interfaces on macOS

AlexGh12:
I adjusted the script to obtain the IP, because when it has more than one network interface and the first one doesn't have a network, it gives an error, so I go through the next network interface if it doesn't return an IP until it finishes.


AlexGh12:
> Could you rebase on top of the current `master` branch?

Sorry, I don't understand you.

akinomyoga:
A pull request should be submitted based on the commits made on top of the latest `master` branch in the upstream repository (`ohmybash/oh-my-bash` in this case). Your pull request is based on an old `master` branch, so the pull request contains unrelated commits, which have already been merged. You first need to fetch the upstream repository, and run  `git rebase -i upstream/master`, select only the relevant commits and remove the others in the opened editor, and then close the editor.

PR #705, opened 2025-08-07
Feature/jump completion

matlads:
This PR adds completion for the jump command, and updates the plugin documentation appropriately

### DIFF
--- a/completions/jump.completion.sh
+++ b/completions/jump.completion.sh
@@ -1,0 +1,10 @@
+#! bash oh-my-bash.module
+
+if _omb_util_command_exists jump; then
+	_jump()
+	{
+	    local JUMP_COMMANDS=$(jump 2>&1 | command grep -E '^ +' | awk '{print $1}')
+	    COMPREPLY=( $(compgen -W "$JUMP_COMMANDS" -- "${COMP_WORDS[1]}") )
+	}
+	complete -F _jump jump
+fi
--- a/plugins/jump/README.md
+++ b/plugins/jump/README.md
@@ -2,10 +2,25 @@
 
 The jump plugin enables jump on bash.
 
-To use it, install
-[jump](https://github.com/gsamokovarov/jump?tab=readme-ov-file#installation)
-and add jump to the plugins array of your bashrc file:
+Jump integrates with your shell and learns about your navigational habits by keeping track of the directories you visit. It gives you the most visited directory for the shortest search term you type.
 
-```bash
-plugins=(... jump)
-```
+## Installation
+
+1. [Install jump](https://github.com/gsamokovarov/jump?tab=readme-ov-file#installation)
+
+2. Enable the plugin by adding it to the plugins array of your bashrc file:
+
+    ```bash
+    plugins=(... jump)
+    ```
+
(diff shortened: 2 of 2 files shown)


completions/jump.completion.sh
```diff
@@ -0,0 +1,10 @@
+#! bash oh-my-bash.module
+
+if _omb_util_command_exists jump; then
+	_jump()
+	{
+	    local JUMP_COMMANDS=$(jump 2>&1 | command grep -E '^ +' | awk '{print $1}')
+	    COMPREPLY=( $(compgen -W "$JUMP_COMMANDS" -- "${COMP_WORDS[1]}") )
```
akinomyoga, line 7:
```suggestion
	    COMPREPLY=( $(compgen -W '$JUMP_COMMANDS' -- "${COMP_WORDS[1]}") )
```

The option argument of `compgen -W` is subject to `eval`, so there is a problem with shell double expansions. `$...` should be singly quoted to avoid an extra expansion.

completions/jump.completion.sh
```diff
@@ -0,0 +1,10 @@
+#! bash oh-my-bash.module
+
+if _omb_util_command_exists jump; then
+	_jump()
```
akinomyoga, line 4: Please follow the [naming convention](https://github.com/ohmybash/oh-my-bash/blob/master/CONTRIBUTING.md#naming-convention-of-functions-and-global-variables). I've fixed this in the other PR without explanation, so could you learn from the changes in the other PR?

plugins/jump/README.md
```diff
@@ -2,10 +2,25 @@
...
+3. Enable the completions by adding it to the completions array of bashrc file:
+
+    ```bash
+    completions=(... jump)
+    ```
+
+## Usage
+
+See the [jump usage documentation](https://github.com/gsamokovarov/jump?tab=readme-ov-file#usage)
+for information on how to use jump
```
akinomyoga, line 26:
```suggestion
for information on how to use jump.
```

plugins/jump/README.md
```diff
@@ -2,10 +2,25 @@
...
+    ```bash
+    plugins=(... jump)
+    ```
+
+3. Enable the completions by adding it to the completions array of bashrc file:
+
+    ```bash
+    completions=(... jump)
+    ```
+
```
akinomyoga, line 22: This explanation appears to imply that both the `jump` plugin and the `jump` completion are mandatory, which is not the case. Could you elaborate on the wording so that each of the plugin and completion can be enabled at the user's discretion?

akinomyoga (changes requested):

akinomyoga (changes requested):
I now noticed that the indentation is made with TAB, but I use two spaces in this repository. Could you update the indentation?

completions/jump.completion.sh
```diff
@@ -0,0 +1,10 @@
+#! bash oh-my-bash.module
+
+if _omb_util_command_exists jump; then
+  function _omb_plugin_jump_completion
+  {
```
akinomyoga, line 5:
```suggestion
  function _omb_plugin_jump_completion {
```

akinomyoga (approved):

PR #736, opened 2026-02-20
feat(kube-ps1): add kube-ps1 plugin

### LINKED ISSUE #719: Feature Request: integrate kube-ps1 as OMB plugin
## Description

I recently installed oh-my-bash and it's working great, but I'm unable to get [kube-ps1](https://github.com/jonmosco/kube-ps1) integration working properly. kube-ps1 is a very popular and useful plugin for Kubernetes users that displays the current Kubernetes context and namespace in the bash prompt.

kube-ps1 in action

## Current Situation

- kube-ps1 is not included as an oh-my-bash
- Manual installation requires cloning the repository and custom configuration
- The integration with oh-my-bash's theme system is not straightforward
- Other popular shell frameworks like oh-my-zsh already include kube-ps1 as a plugin.

## Proposed Solution

Add kube-ps1 as an plugin in oh-my-bash with:
- Proper plugin initialization
- Theme integration support
- Standardized configuration options
- Documentation

## Why this is valuable

- Kubernetes is widely used in development and DevOps
- kube-ps1 has over 2.8k stars on GitHub indicating high demand
- Would provide a consistent experience for Kubernetes users
- Aligns with oh-my-bash's goal of providing useful plugins

## Technical considerations

The plugin should:
- Follow existing plugin structure
- Integrate with the prompt system
- Respect existing oh-my-bash configuration
- Provide toggle enable/disable functionality

rafacouto:
Closes #719 

## Description

Adds a new `kube-ps1` plugin that displays the current Kubernetes context and
namespace in the shell prompt, with `kubeon` / `kubeoff` commands to toggle it
interactively.

## Changes

- `plugins/kube-ps1/kube-ps1.plugin.sh` — main plugin
- `plugins/kube-ps1/README.md` — documentation and configuration reference
- `plugins/kube-ps1/LICENSE` — Apache 2.0 license from the upstream project (required by its license terms)
- `plugins/README.md` — added `kube-ps1` entry to the plugin table

## Implementation details

- Internal functions and variables follow the `_omb_plugin_kube_ps1_*` naming convention.
- Uses `_omb_util_add_prompt_command()` and `_omb_util_command_exists()` from the OMB API.
- The `kube_ps1` segment is injected into `PS1` via a lazy-registered post-theme hook, so it works correctly regardless of which theme is active — no theme modifications required.
- The kubeconfig files are watched for changes; context/namespace are only re-fetched when a file is actually modified, keeping prompt overhead minimal.
- `kubeon -g` / `kubeoff -g` persist the state across sessions via `~/.kube/kube-ps1/disabled`.

## Usage

```bash
# ~/.bashrc
plugins=(... kube-ps1)
```

### DIFF
--- a/lib/utils.sh
+++ b/lib/utils.sh
@@ -347,8 +347,11 @@ _omb_util_unload_hook+=('PS1=$_omb_util_original_PS1')
 
 _omb_util_prompt_command=()
 function _omb_util_prompt_command_hook {
-  local status=$? lastarg=$_ hook
-  for hook in "${_omb_util_prompt_command[@]}"; do
+  local status=$? lastarg=$_ hook i
+  # Use an index loop so hooks registered mid-cycle (e.g. by earlier hooks) are
+  # picked up in the same prompt-render cycle rather than deferred to the next.
+  for ((i = 0; i < ${#_omb_util_prompt_command[@]}; i++)); do
+    hook=${_omb_util_prompt_command[i]}
     _omb_util_setexit "$status" "$lastarg"
     eval -- "$hook"
   done
--- a/plugins/kube-ps1/kube-ps1.plugin.sh
+++ b/plugins/kube-ps1/kube-ps1.plugin.sh
@@ -168,8 +168,14 @@ _omb_plugin_kube_ps1_symbol() {
 # ---------------------------------------------------------------------------
 _omb_plugin_kube_ps1_split_config() {
   local IFS="$1"
-  # word-split on IFS
+  # Disable globbing so unquoted $2 only word-splits on IFS and does not
+  # expand wildcards that may appear in KUBECONFIG paths.
+  local _glob_state=$-
+  set -f
+  # shellcheck disable=SC2086  # intentional word-splitting on IFS
   printf '%s\n' $2
+  # Restore globbing only if it was enabled before this call.
+  [[ $_glob_state == *f* ]] || set +f
 }
 
 _omb_plugin_kube_ps1_file_newer_than() {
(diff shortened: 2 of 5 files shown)


rafacouto:
Thanks @brbecker for the review.
1. fixed: e86438e
2. fixed: 8395eeb
3. fixed: 1f3315f
4. good idea

plugins/kube-ps1/kube-ps1.plugin.sh
```diff
@@ -0,0 +1,444 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# ---------------------------------------------------------------------------
+#
+# This file is a derivative work based on kube-ps1
+# <https://github.com/jonmosco/kube-ps1>.
```
akinomyoga, line 23: Is this based on which commit of jonmosco/kube-ps1? Could you include the link in the form of e.g. https://github.com/jonmosco/kube-ps1/blob/9b41c091d5dd4a99e58cf58b5d98a4847937b1bb/kube-ps1.sh?

plugins/kube-ps1/kube-ps1.plugin.sh
```diff
@@ -0,0 +1,444 @@
...
+#
+# ---------------------------------------------------------------------------
+#
+# Usage: add $(kube_ps1) to your PS1 or theme, and use
+#   kubeon  / kubeoff  to toggle the prompt info.
+#
+# ---------------------------------------------------------------------------
+
+# Debug
+[[ -n ${DEBUG:-} ]] && set -x
```
akinomyoga, line 33: Could you remove this? This is undocumented. Also, the variable `DEBUG` may be used by a user for a different purpose, in which case this would cause results surprising to the user.

rafacouto:
I've tested this plugin a lot in my environment and it works without problems... The bugs announced by Qodo have been fixed. What would be necessary to incorporate it into OMB?

akinomyoga (commented):
I realized that my comments have been left "pending" in the GitHub interface.

lib/utils.sh
```diff
@@ -347,8 +347,11 @@ _omb_util_unload_hook+=('PS1=$_omb_util_original_PS1')
 
 _omb_util_prompt_command=()
 function _omb_util_prompt_command_hook {
-  local status=$? lastarg=$_ hook
-  for hook in "${_omb_util_prompt_command[@]}"; do
+  local status=$? lastarg=$_ hook i
+  # Use an index loop so hooks registered mid-cycle (e.g. by earlier hooks) are
+  # picked up in the same prompt-render cycle rather than deferred to the next.
+  for ((i = 0; i < ${#_omb_util_prompt_command[@]}; i++)); do
```
akinomyoga, line 353: I doubt if we should do this. I think we can't say the original behavior is wrong, but it's just a design. Then, this PR tries to modify an existing utility that others have relied on, just to make the behavior useful for a specific plugin. This doesn't seem safe.

PR #750, opened 2026-04-01
Uses upfind to resolve SCM

### LINKED ISSUE #618: Prompt extremely slow suddenly
My Oh My Bash prompt is suddenly taking several hundred ms to display when hitting enter on the terminal. I'm not sure how to debug this. Uninstalling OMB fixes the problem, so it's related in some way.

EvansJahja:
Instead of checking current directory and invoking external tools, we check SCM by checking the existence of directories (.git, .hg, .svn) in current and ancestor directories (until root)

Resolves #618

### DIFF
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -121,7 +121,7 @@ function _omb_prompt_git_status_enabled {
 function _upfind() {
   local target current
   target="$1"
-  current="$(realpath --canonicalize-existing "$PWD")"
+  current="$(realpath "$PWD")"
 
   while [[ "$current" != "/" ]]; do
     if [[ -e "${current}/${target}" ]]; then
--- a/lib/omb-prompt-base.sh
+++ b/lib/omb-prompt-base.sh
@@ -117,13 +117,29 @@ function _omb_prompt_git_status_enabled {
     awk '$2== "1" {hide_status = 1;} END { print hide_status; }') != "1" ]]
 }
 
+# https://github.com/sharkdp/fd/issues/1537#issuecomment-2071361869
+function _upfind() {
+  local target current
+  target="$1"
+  current="$(realpath --canonicalize-existing "$PWD")"
+
+  while [[ "$current" != "/" ]]; do
+    if [[ -e "${current}/${target}" ]]; then
+      return
+    else
+      current="$(dirname "$current")"
+    fi
+  done
+  return 1
+}
+
+
+
 function scm {
   if [[ "$SCM_CHECK" = false ]]; then SCM=$SCM_NONE
-  elif [[ -f .git/HEAD ]]; then SCM=$SCM_GIT
-  elif _omb_util_binary_exists git && [[ -n "$(_omb_prompt_git rev-parse --is-inside-work-tree 2> /dev/null)" ]]; then SCM=$SCM_GIT
-  elif [[ -d .hg ]]; then SCM=$SCM_HG
-  elif _omb_util_binary_exists hg && [[ -n "$(command hg root 2> /dev/null)" ]]; then SCM=$SCM_HG
-  elif [[ -d .svn ]]; then SCM=$SCM_SVN
+    elif _upfind .git; then SCM=$SCM_GIT
+    elif _upfind .hg; then SCM=$SCM_HG
+    elif _upfind .svn; then SCM=$SCM_SVN
   else SCM=$SCM_NONE
   fi
 }


lib/omb-prompt-base.sh
```diff
@@ -117,13 +117,29 @@ function _omb_prompt_git_status_enabled {
...
 function scm {
   if [[ "$SCM_CHECK" = false ]]; then SCM=$SCM_NONE
-  elif [[ -f .git/HEAD ]]; then SCM=$SCM_GIT
-  elif _omb_util_binary_exists git && [[ -n "$(_omb_prompt_git rev-parse --is-inside-work-tree 2> /dev/null)" ]]; then SCM=$SCM_GIT
-  elif [[ -d .hg ]]; then SCM=$SCM_HG
-  elif _omb_util_binary_exists hg && [[ -n "$(command hg root 2> /dev/null)" ]]; then SCM=$SCM_HG
-  elif [[ -d .svn ]]; then SCM=$SCM_SVN
+    elif _upfind .git; then SCM=$SCM_GIT
+    elif _upfind .hg; then SCM=$SCM_HG
+    elif _upfind .svn; then SCM=$SCM_SVN
```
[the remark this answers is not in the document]
EvansJahja: on my potato machine, realpath and dirname takes about 0.005s. a single `hg root` invocation is at least 0.9 s. That said I'm open to better ways

lib/omb-prompt-base.sh
```diff
@@ -117,13 +117,29 @@ function _omb_prompt_git_status_enabled {
...
+  current="$(realpath --canonicalize-existing "$PWD")"
+
+  while [[ "$current" != "/" ]]; do
+    if [[ -e "${current}/${target}" ]]; then
+      return
+    else
+      current="$(dirname "$current")"
+    fi
+  done
+  return 1
```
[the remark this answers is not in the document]
EvansJahja: `--canonicalize-existing` is indeed GNU only, I'm not sure what it does. Should I remove it?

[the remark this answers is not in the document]
EvansJahja: I'm gonna settle with just dropping the flag for now
e1d0b47

lib/omb-prompt-base.sh
```diff
@@ -117,13 +117,29 @@ function _omb_prompt_git_status_enabled {
...
+
+
 function scm {
   if [[ "$SCM_CHECK" = false ]]; then SCM=$SCM_NONE
-  elif [[ -f .git/HEAD ]]; then SCM=$SCM_GIT
-  elif _omb_util_binary_exists git && [[ -n "$(_omb_prompt_git rev-parse --is-inside-work-tree 2> /dev/null)" ]]; then SCM=$SCM_GIT
-  elif [[ -d .hg ]]; then SCM=$SCM_HG
-  elif _omb_util_binary_exists hg && [[ -n "$(command hg root 2> /dev/null)" ]]; then SCM=$SCM_HG
-  elif [[ -d .svn ]]; then SCM=$SCM_SVN
+    elif _upfind .git; then SCM=$SCM_GIT
```
akinomyoga, line 140: `.git` can be a regular file, containing the path to the actual directory.

EvansJahja: This works correctly if there's a .git file/dir

akinomyoga: Ah, OK. I didn't see the detailed implementation of `_upfind`. Sorry for the noise.

akinomyoga: though the description of this PR said "by checking the existence of directories (.git, .hg, .svn) in current and ancestor directories (until root)".

EvansJahja: No worries. That's because I didn't even know that `.git` **file** is a thing

lib/omb-prompt-base.sh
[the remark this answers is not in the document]
akinomyoga: @EvansJahja Do you have a plan to work on this? I agree with the AI review that it's a problem to add extra forking cost. The fork & exec cost largely depends on the host system. It can be 0.05s or something in specific systems such as MSYS2. At least, `dirname` can be implemented without forking.

lib/omb-prompt-base.sh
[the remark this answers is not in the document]
akinomyoga: Oh My Bash already has a portable implementation for this. Use [`_omb_util_readlink`](https://github.com/ohmybash/oh-my-bash/blob/7d26139293dea898a98cf4adacc19af4b0955145/lib/readlink.sh#L107). This emulates GNU coreutils `readlink -f`, which is equivalent to `realpath`.

lib/omb-prompt-base.sh
```diff
@@ -117,13 +117,29 @@ function _omb_prompt_git_status_enabled {
     awk '$2== "1" {hide_status = 1;} END { print hide_status; }') != "1" ]]
 }
 
+# https://github.com/sharkdp/fd/issues/1537#issuecomment-2071361869
+function _upfind() {
```
akinomyoga, line 121: The internal function (which is not intended to be directly used by the interactive user) should be prefixed by `_omb_prompt`.

lib/omb-prompt-base.sh
akinomyoga, line 140: The indentation is broken. It seems there are 3 spaces in those lines.

lib/omb-prompt-base.sh
```diff
@@ -117,13 +117,29 @@ function _omb_prompt_git_status_enabled {
...
+  local target current
+  target="$1"
+  current="$(realpath "$PWD")"
+
+  while [[ "$current" != "/" ]]; do
```
akinomyoga, line 126: In Cygwin/MSYS2, this falls into an infinite loop because the root directory can also be `C:`, `\`, `\\`, etc. The loop should be terminated once `/` or `\` is not found in the middle in `current`.

PR #764, opened 2026-05-28
feat(lib/omb-prompt-base): Add support for OMB_PROMPT_FULL_PATH to pr…

native-api:
…int full workdir in prompt

Subj.

Use the result in all themes.

Printing the full path in prompt is imperative for me since I mostly work with source code trees and system configuration trees both of which are too complex to figure out where I am just by the basename.

And due to the fact that themes overwrite `PS1` in `PROMPT_COMMAND`, editing `PS1` in `~/.bashrc` has no effect.


lib/omb-prompt-base.sh
```diff
@@ -11,6 +11,10 @@ CLOCK_THEME_PROMPT_SUFFIX=''
 
 THEME_PROMPT_HOST='\H'
 
+[[ $OMB_PROMPT_FULL_PATH == true ]] \
+  && THEME_PROMPT_WORKDIR='\w' \
+  || THEME_PROMPT_WORKDIR='\W'
```
native-api, line 16: [CONTRIBUTING.md](https://github.com/ohmybash/oh-my-bash/blob/ae44713948daaf99b2e2c6fe7e436f6a83bf07da/CONTRIBUTING.md#improving-pr) says that internal variables should be named "`_omb_${namespace}_${varname}`". But I see that in reality, only internal functions are named so. So naming my new variable like that would be inconsistent with the codebase.

akinomyoga: Other variables in this file are not internal variables, but intended to be set by the user for customization. Otherwise it's strange that the variables always have specific constant values. In addition, while CONTRIBUTING.md says

> Initially, we haven't cared about the naming convention very much, but we now try to improve it. In particular, the new codes should follow this naming convention.

the code in `omb-prompt-base.sh` is not "the new codes" but one of the oldest codes in Oh-My-Bash.

PR #771, opened 2026-07-30
Add Do Not Track plugin

RobLoach:
This introduces a new plugin that adds a bunch of environment variables that will disable telemetry across a bunch of applications. Add the `donottrack` plugin, and sending usage statistics is now (hopefully) disabled.

I added the environment variables for a bunch of applications I use, along with a few others I found on the web. If you have others you'd like to add, please be my guest.

Inspired by https://donottrack.sh

### DIFF
--- a/plugins/donottrack/donottrack.plugin.sh
+++ b/plugins/donottrack/donottrack.plugin.sh
@@ -6,9 +6,9 @@
 # Opt out of the telemetry and usage reporting of various CLI applications by
 # exporting the environment variables they check.
 
-# Export the variable "$1" with the value "$2" unless it is already set.
+# Export the variable "$1" with the value "$2", keeping any existing value.
 function _omb_plugin_donottrack_set {
-  [[ ${!1+set} ]] || export "$1=$2"
+  export "$1=${!1-$2}"
 }
 
 # Generic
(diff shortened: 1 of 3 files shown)


plugins/donottrack/donottrack.plugin.sh
```diff
@@ -0,0 +1,89 @@
...
+# @RobLoach 2026
+#
+# Do Not Track Plugin
+#
+# Opt out of the telemetry and usage reporting of various CLI applications by
+# exporting the environment variables they check.
+
+# Export the variable "$1" with the value "$2" unless it is already set.
+function _omb_plugin_donottrack_set {
+  [[ ${!1+set} ]] || export "$1=$2"
```
akinomyoga: I think we probably want to export the variable even if it has already been set, in case that an existing variable does not have the export attribute.

akinomyoga:
```suggestion
  if [[ ${!1+set} ]]; then
    export "$1"
  else
    export "$1=$2"
  fi
```

Or, simply,

```suggestion
  export "$1=${1-$2}"
```

RobLoach: Should it be `export "$1=${!1-$2}"` ? With the `!`?

My assumption with only exporting it if it didn't exist was so that you could opt-in to telemetry on individual entries if you really wanted to.

akinomyoga: > Should it be `export "$1=${!1-$2}"` ? With the `!`?

Yeah, you are right. It should have been `${!1-$2}`.

> you could opt-in to telemetry on individual entries if you really wanted to.

But when the variable has a value, doesn't that mean the user has set the value?

RobLoach: Good point, fixed :melting_face:

akinomyoga (approved):

akinomyoga:
Thanks.

PR #774, opened 2026-08-21
themes/duru: fix undefined ${gray} in venv prompt

koppor:
The `venv` function in the duru theme colors the `" in "` text with `${gray}`, but `gray` is never defined: the color variables are all `_omb_prompt_*` (`lib/omb-prompt-colors.sh`), and plain `gray` is not among the deprecated compatibility names either (`black`, `cyan`, `white`, `orange` and the `bold_*`/`underline_*`/`background_*` variants are). The variable expands to empty, so the text renders in whatever color was active before instead of gray.

Use `${_omb_prompt_gray}`, matching the `_omb_prompt_red` already used on the next segment.

Noticed while porting duru to a custom theme.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SABAE23V7T3PvYht4QmG9j


akinomyoga:
Thank you for the report. I've checked the history, but it seems to have been broken since the beginning when the setting was introduced in e65c390b.

akinomyoga (approved):

PR #778, opened 2026-09-11
Plugin proposal : python-m

thomas-saigre:
The purpose of this plugin is to allow auto completion when running python module with `python3 -m ...`. Cf the readme file for more details.


akinomyoga:
What is the difference from [completions-core/python.bash:87-90@scop/bash-completion](https://github.com/scop/bash-completion/blob/19695e36d75514672f10bf8141a1a91d4252c9ec/completions-core/python.bash#L87-L90)? We should probably include it in scop/bash-completion.

thomas-saigre:
> What is the difference from [completions-core/python.bash:87-90@scop/bash-completion](https://github.com/scop/bash-completion/blob/19695e36d75514672f10bf8141a1a91d4252c9ec/completions-core/python.bash#L87-L90)? We should probably include it in scop/bash-completion.

I never heard of bash-completion, but my proposition seems indeed to do what it does, maybe in a less consistent way, as I manually check for sub-packages, and do not use `pkgutil` (cf [source code](https://github.com/scop/bash-completion/blob/19695e36d75514672f10bf8141a1a91d4252c9ec/helpers-core/python))

akinomyoga:
I checked the behavior of `bash-completion`'s package name generation, but it seems extremely slow particularly when `.` is included in the completion target. At least, there is room to improve.

